PR #348, opened 2015-02-25
Added Pod Coverage to installation guide for Ubuntu 14.04

marcgweg:
Added the package libpod-coverage-perl to the installation guidelines for zonemaster-engine for Ubuntu 14.04.

### DIFF
--- a/docs/documentation/installation.md
+++ b/docs/documentation/installation.md
@@ -34,7 +34,7 @@ Another application is the web interface which is split in two parts. The user v
     liblist-moreutils-perl libio-socket-inet6-perl libmodule-find-perl \
     libmoose-perl libnet-ip-perl libfile-sharedir-perl libhash-merge-perl \
     libreadonly-perl libldns-dev libmodule-install-perl \
-	libmail-rfc822-address-perl libintl-xs-perl
+    libmail-rfc822-address-perl libintl-xs-perl libpod-coverage-perl 
 
 **Install CPAN dependencies**
 


cdybedahl:
Is that really necessary? The POD coverage test is only useful for people developing Zonemaster itself, and these installation instructions are aimed at people only using it.

marcgweg:
Dear Calle,

> On 25 Feb 2015, at 09:58, Calle Dybedahl <notifications@github.com> wrote:
> 
> Is that really necessary? The POD coverage test is only useful for people developing Zonemaster itself, and these installation instructions are aimed at people only using it.
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/dotse/zonemaster/pull/348#issuecomment-75926145>.
> 
I understand this. be aware though that the make test gives a warning on this, ‘cause it can’t find mentioned Perl module. From a user point of view, this could be confusing. But then again, who is the user and what might we expect from him? ;-)

Regards,
Marc

cdybedahl:
It does, yes, but exactly that warning is quite common in CPAN modules. I think the kind of user who installs this manually will already be familiar with it. If experience shows that I'm wrong we can always add this later.

Thanks for the contribution anyway :-)

marcgweg:
Dear Calle,

> Thanks for the contribution anyway :-)
No problem. 

I am working on the installation instructions for the backend. Is it true, you must install the engine before installing the backend? 
If this is the case, doesn’t it make sense to refer to the installation of the engine, and describe the differential of the packages to install instead of installing (once again) all packages?

Regards,
Marc

mtoma:
Hello Marc,

The backend needs to be able to run the zonemaster engine to perform the tests.
The first dependency of the backend is the Zonemaster module what will install all the needed dependencies to run the Zonemaster engine. There might probably be a slight overlap between the backend modules and the ones used by the zonemaster engine (like Moose). I choose to list all modules required by the backend to not depend on the modules the zonemaster engine uses. This would break the installation instructions if the Zonemaster engine stops using one of the modules in the future.

Michal

----- Original Message -----

PR #388, opened 2015-06-08
Added nameserver08 for 0X20 test

sandoche2k:
@pawal please verify and integrate.

### DIFF
--- a/docs/requirements/TestRequirements.md
+++ b/docs/requirements/TestRequirements.md
@@ -95,7 +95,7 @@ Tests to implement
 |R79|If DS at parent, child zone must be signed              |[DNSSEC](../specifications/tests/DNSSEC-TP/dnssec11.md)|
 |R80|Preservation of wname case in DNS answers               |Not yet specified|
 |R81|Test Upward referral         |[NAMESERVER](../specifications/tests/Nameserver-TP/nameserver07.md)|
-|R82|Test Case sensitivity  |[NAMESERVER](../specifications/tests/Nameserver-TP/nameserver08.md)|
+|R82|Test QNAME Case sensitivity  |[NAMESERVER](../specifications/tests/Nameserver-TP/nameserver08.md)|
 
 
 Future tests
--- a/docs/requirements/TestRequirements.md
+++ b/docs/requirements/TestRequirements.md
@@ -95,6 +95,8 @@ Tests to implement
 |R79|If DS at parent, child zone must be signed              |[DNSSEC](../specifications/tests/DNSSEC-TP/dnssec11.md)|
 |R80|Preservation of wname case in DNS answers               |Not yet specified|
 |R81|Test Upward referral         |[NAMESERVER](../specifications/tests/Nameserver-TP/nameserver07.md)|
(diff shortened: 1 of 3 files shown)


pawal:
The test case identifier is very very long. Could you make it shortern?

Perhaps, "qname case preservation in answers from authoritative name server"

sandoche2k:
I have done it. But i have not used the same wording as yours, since the
verficiation here is done only at the question section

pawal:
I have read through the details in the "ordered description" now, but I fail to see where you verify that the qname matches the _answer_ which is the whole point of the test. It is useless to compare the qmane with the qname?  I think you just forgot to mention the DNS reply with the ANSWER and QUESTION sections.

sandoche2k:
Yes, i have not been precise to mention the response packet.

I have updated wherein the comparison is done between the QNAME in the
question section in the query with the QNAME in the question section in the
response.

I did test with some well known authoritative servers, wherein the answer
section of the response does not preserve the case. And, after discussion
with Stephane, i restricted to question section of the response.

sandoche2k:
@vlevigneron test spec completed

vlevigneron:
Implementation is completed but:

1/ Specifications have changed and are incomplete now (no idea why).
2/ I guess you mean Method4 and Method5 instead of Method2 and Method3
3/ Can you provide list of authoritative servers with the "bad" behaviour. Of general way, when you know of relevant examples, please provide them. This facilitates the implementation of unit tests.

sandoche2k:
Thanks Vincent. Fixed the incomplete specification.

No, i did mean "Method 2" and "Method 3".  The idea is to get the name
servers and query them with the mixed case FQDN. Correct me if i am wrong.

The examples i found in the literature have been fixed. If i have new
examples, i will update.

Sandoche.

vlevigneron:
Calle and I answered several times about the use of Method2...5, Method2
and Method3 return names while Method4 and Method5 return name server
objects you can query. So in this case, if we want to query them, it's
method4 and method5.

le 15 juin, sandoche2k a ?crit :
> Thanks Vincent. Fixed the incomplete specification.
> 
> No, i did mean "Method 2" and "Method 3".  The idea is to get the name
> servers and query them with the mixed case FQDN. Correct me if i am wrong.
> 
> The examples i found in the literature have been fixed. If i have new
> examples, i will update.
> 
> Sandoche.
> 
> On 12 June 2015 at 08:28, vlevigneron <notifications@github.com> wrote:
> 
> > Implementation is completed but:
> >
> > 1/ Specifications have changed and are incomplete now (no idea why).
> > 2/ I guess you mean Method4 and Method5 instead of Method2 and Method3
> > 3/ Can you provide list of authoritative servers with the "bad" behaviour.
> > Of general way, when you know of relevant examples, please provide them.
> > This facilitates the implementation of unit tests.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/dotse/zonemaster/pull/388#issuecomment-111377600>.
> >
> 
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/dotse/zonemaster/pull/388#issuecomment-112309466

-- 
	Vincent Levigneron  A.F.N.I.C.  Vincent.Levigneron@nic.fr

PR #390, opened 2015-06-09
Added nameserver 09 - For testing case sensitivity

sandoche2k:
@pawal verify and integrate


pawal:
qname has been misspelled wname in the requirements

pawal:
Also, there is no need to hyphenate the word insensitive.

sandoche2k:
@pawal  Corrections updated.

PR #396, opened 2015-06-16
Specifications for namesever08.md was incomplete. Fixed it.

sandoche2k:


pawal:
The test case identifier is very long, is it possible to make it shorter?

PR #471, opened 2016-10-03
Draft of a system testing document

mattias-p:
This is just a draft. It should probably be better integrated with other documents. Making a PR per request from @sandoche2k.


mattias-p:
@matsduf, I agree that proper translation is an important thing to test, but configurations aren't used like that.

The purpose of configurations is to provide different contexts for the tests to run in. The reason locale is a configuration parameter is that things like databases, gettext and Zonemaster itself are able to make use of the system locale. Since Zonemaster is a multilingual application that takes language preferences from the user, I believe the most important property of Zonemaster with regard to the system locale is the ability to properly ignore it.

If we test all supported languages in both "C" and "fr_FR.UTF-8" system locales, we can be confident that translations use the user provided locale and ignore the system locale.

So I think supported languages have very little bearing on the locale configuration parameter. It is meaningful to have at least two different locale parameter values so are able to spot deviations. It's also a good idea to test both using locale settings that matches a supported language and settings that don't.

I'll update Configuration section with a couple of sentences about purpose.

sandoche2k:
@matsduf if you could merge the pull request, i can comment on the draft committed by @mattias-p

matsduf:
@sandoche2k, can you leave your comments with the pull request? @mattias-p is about to do further updates.

docs/internal-documentation/maintenance/SystemTesting.md
```diff
@@ -0,0 +1,200 @@
...
+  * The supported versions are the LTS releases (excluding 12.04) as specified
+    at: https://wiki.ubuntu.com/Releases
+  * The supported patch level is the one specified at:
+    https://wiki.ubuntu.com/Releases
+
+
+### Locale
+
+The "C" locale is supported along with the locales supported for language
+translation.
```
matsduf: It is important that explicitly state whate locales that Zonemaster supports. I hade problem installing Zonemaster Backend with locale sv_SE.utf8. Some unit test failed then. It worked with locale C. I had problem running the backend with locale C.

Locale is important for Zonemaster. We need to decide and define what locales that are supported at installation. That could be a small list. We need to decide and define what locales that are supported when running backend and GUI. That could also be a small list. And we need to decide and define what locales that are supported when running CLI. That should be a larger list at least including locales used for French, Swedish and English. I guess there are multiple fr_??.utf8.

sandoche2k:
@matsduf if you want others to contribute on the document prepared by @mattias-p, then it should be merged to the repository.

mtoma:
The backend should now be working with all locales (tests were only done with utf8 locales).
The backend issue was caused by a backend bug that is now solved.

matsduf:
@sandoche2k, I suggest that we have the following steps:

1. Person A creates something and creates a pull request.
2. Other persons read and comment.
3. If there are comments that need updates, Person A updates the files and the pull request. We go back to step 2.
4. When we have reached (rough) consensus, we either merge or discard.

By keeping things as pull requests until we agree, we can easily distinguish between work-in-progress and updates. If we merge now the distinction is blurred. By keeping it as pull request, it is clear who is in charge of the files. 

Pull requests are as public as merged data.

Have I missed something?

sandoche2k:
@matsduf @mattias-p we have an excellent wiki (https://github.com/dotse/zonemaster/wiki/Pull-Request-Crib-Sheet) by Calle which explains what we do in the case of multiple authors are working on the same document. Git is a versioning software, and you do have a history and you can revert back when needed.  It is much better than sending back and forth mails. 
As @mattias-p is the owner of the document, he will send a signal to the product owner @matsduf when he feels that contributions made by others can be merged. 
All others update the document by creating pull requests. The pull request is merged only in the case of the document being mature enough to be merged in the "master" branch.
Please look at the Wiki in case of doubts.

mattias-p:
@sandoche2k I accept that email is not the best tool for this job.

I hadn't found the wiki. Thanks for pointing it out!

With regard to the crib sheet I consider myself to be in the middle of item 6. Since I would like to have feedback, and on your request, I created this PR. According to the crib sheet, and common sense, the work should not be merged until it is completed and reviewed.

If you or anyone feels that the review process is insufficient at this stage in development, please make your own pull requests to my feature branch.

mattias-p:
In 95e6dbd9abb2c9ff2e62183ad53b115baa9de551 I added a note that what we should declare what configuration parameters we support in the README. Of course, before we can declare them we must select them. Initially I thought the selection process should be formal, but I'm no longer convinced about that. What do you guys say? Instead I think the most important thing is that we're diligent in selecting and declaring the supported configuration parameters an appropriate time in the release cycle. Do we have a document where this step can be included?

Regarding 366e104b29679780bb2e1eb2ce67309f61197c7f, does anyone have ideas on how we should select what system locales to use during testing?

docs/internal-documentation/maintenance/SystemTesting.md
```diff
@@ -0,0 +1,211 @@
...
+### System locale
+
+Zonemaster promises to behave correctly when the system locale is set to one of
+these locales.
+
+... Do we support all possible locales in this sense? Since we cannot reasonably
+perform testing for all possible locales, supported system locales are grouped
+into a small set of "tier 1" system locales to be used in test configurations,
+and the remaining "tier 2" system locales.
+
```
matsduf: I think it is reasonably to support only en_US.utf8 and C for installation and running the backend and GUI. That keeps those parts down. And to support at least en_US.utf8, C, fr_FR.utf8 and sv_SE.utf8 for Zonemaster CLI. I would like to see it expanded to en_XX.utf8, C, C.utf8, fr_XX.utf8 and sv_XX.utf8. Changing locale when running Zonemaster CLI commands is managable.

docs/internal-documentation/maintenance/SystemTesting.md
```diff
@@ -0,0 +1,211 @@
...
+### Database engine version
+
+The following database engines are supported:
+
+* MySQL
+* PostgreSQL
+* SQLite
+
+... To be specified. It's slightly complicated.
+
```
matsduf: We have to keep down the number of versions that we support.

docs/internal-documentation/maintenance/SystemTesting.md
```diff
@@ -0,0 +1,211 @@
...
+For each configuration:
+
+1. Prepare a machine
+   1. Allocate a machine with the architecture specified by the configuration.
+   2. Install the OS as specified by the configuration on the machine.
+
+2. Install Zonemaster Engine
+   1. Follow the installation instruction at zonemaster-engine/docs/installation.md
+      to the letter.
+   2. Perform a sanity check.
```
matsduf: That document must be updated to specify an installation directly from Github, or point at some other document, since testing will/should be done before we have the package on Github.

docs/internal-documentation/maintenance/SystemTesting.md
mattias-p: Updating the release process document to run unit tests with different locales would catch future bugs of this type. I think it seems unlikely that locale related bugs show up in other places during installation. With this update and assumption we shouldn't have to test the installation with different locales.

mattias-p: The correct syntax for fr_??.utf8 is `fr.UTF-8`.
The runtime behaviour of CLI should be affected by the system locale. That should be reflected in test procedures rather than the test configurations.
The runtime behaviour of backend and GUI should not be affected by their respective system locales. There should be test configurations to cover these.

docs/internal-documentation/maintenance/SupportCriteria.md
```diff
@@ -0,0 +1,124 @@
...
+## General
+
+It is legitimate to avoid supporting items that are otherwise perfectly eligible
+for support, just for the sake of limiting the total number of configuration
+tuples.
+
+
+## Architectures
+
+Can we say something about how we chose to support only amd64/x86_64 and nothing
```
matsduf: Is this still work in progress, @mattias-p

docs/internal-documentation/maintenance/SupportCriteria.md
```diff
@@ -0,0 +1,124 @@
...
+else?
+
+
+## Operating systems
+
+Can we say something about how the current quartet of operating systems was
+chosen?
```
matsduf: Is this still work in progress, @mattias-p ?

docs/internal-documentation/maintenance/SupportCriteria.md
```diff
@@ -0,0 +1,124 @@
...
+  * LTS releases are listed here:
+    https://wiki.ubuntu.com/Releases
+  * The supported patch level is the one specified at:
+    https://wiki.ubuntu.com/Releases
+
+
+## Database engines
+
+Can we say something about how the current triple of database engines was
+chosen?
```
matsduf: Is this still work in progress, @mattias-p?

docs/internal-documentation/maintenance/SystemTesting.md
mattias-p: Can I refer to "the release branch"?

docs/internal-documentation/maintenance/SupportCriteria.md
mattias-p: Well, it's kind of stalled. I have no insights into how these were chosen in the first place, now why they are the best choice right now.

mattias-p:
Pushed three old commits I apparently had forgot to push.

docs/internal-documentation/maintenance/SupportCriteria.md
matsduf: I think we should only state that we only actively test Zonemaster under the selected OSs. I do not think that we have to motivate the choice.

docs/internal-documentation/maintenance/SupportCriteria.md
matsduf: I think that we should only state that Zonemaster backend only is integrated with the selected databases. I do not think that we have to motivate the choice.

docs/internal-documentation/maintenance/SystemTesting.md
```diff
@@ -44,58 +44,76 @@ The set of configurations must include at least:
...
-
+1. Build a preliminary distribution tarball for each component
+
+2. For each configuration:
+
+   1. Prepare a machine
+      1. Allocate a machine with the architecture specified by the configuration.
+      2. Follow the preparation document for the operating system specified by the configuration.
+         * [FreeBSD-Preparation.md](https://github.com/dotse/zonemaster/blob/master/docs/internal-documentation/distrib-testing/FreeBSD-Preparation.md)
+
```
matsduf: Why only FreeBSD?

mattias-p: Well, the others haven't been written yet. My idea was that a Debian-preparation document could be extracted from the Debian-System-Testing document of #473. The Ubuntu-preparation document wouldn't differ much from the Debian one - maybe they could even share the same instruction?

@sandoche2k what do you think of this?

sandoche2k:
Both debian and Ubuntu have same instructions

docs/internal-documentation/maintenance/SystemTesting.md
```diff
@@ -0,0 +1,182 @@
...
+* system locale (e.g. en_US.UTF-8)
+* Perl version (e.g. 5.20)
+
+The set of possible values for the following configuration parameters (except
+for system locale) are declared in the README.md of the top-level zonemaster
+repository.
+
+The set of possible values for the system locale configuration parameter is:
+* C
+* sv_SE.UTF-8
```
matsduf: It should be en_US.UTF-8, shouldn't it?

docs/internal-documentation/maintenance/SystemTesting.md
matsduf: At least add placeholder for the other OSs.

mattias-p:
@matsduf Before merging I'd like to review its integration with related documents (like installation instructions for individual components).

mattias-p:
@sandoche2k I guess it's going to be you and I testing the next release as well? How do you feel about doing it according to this document?

sandoche2k:
@matsduf i did do a review. You can merge it.

PR #473, opened 2016-10-04
Added documentation for Debian testing

sandoche2k:


mattias-p:
First of all, I agree that we should have an instruction for preparing test configurations for the different distributions we support, and I applaude this initiative. Secondly, I have two items that I think need to be resolved.

1. I think there should be no overlap between this instruction and the standard installation procedure. The actual standard installation procedure should be referred to instead. The main reason for this is that we really want good testing of the standard installation procedure.

2. The term "distribution" is used here whereas the term "operating system" is used in the SystemTesting document for the same thing. I think it's important that we use the same language across different documents to make it plain to the reader when an apple is an apple. Is any of the terms being used in other documents? I've used "operating system version" in SystemTesting.md - does the term "distribution version" make sense?

sandoche2k:
@mattias-p For your first point "why i have added a separate installation procedure? Normally when an user (external to the team) installs any of the components of Zonemaster, it should be from CPAN (sudo cpan -i Zonemaster). But, when we test, the latest fixes are not yet updated to CPAN. So we have to follow manual instructions of building the code. That is the reason this testing document is added to "internal-documentation". This will save us considerable amount of time as before every release we need to test multiple times.

For your second point distribution vs OS, there is a rather thin line. For e.g. Ubuntu is considered as a Linux distribution as well as OS. This is an internal document only for our eyes so we can discuss and determine the correct wordings accordingly.

docs/internal-documentation/distrib-testing/Debian-Testing-Instructions.md
```diff
@@ -0,0 +1,282 @@
...
+The documentation covers the following operating systems:
+
+ * [1] <a href="#Debian">Debian Wheezy (version 7)</a>
+
+
+The Engine Perl module name is 'Zonemaster'.
+
+## Prerequisite
+
+   No other Zonemaster component is required by the engine.
```
matsduf: Shouldn't net-ldns be considered to be a Zonemaster component, @sandoche2k?

matsduf:
@sandoche2k, can you address the issue so that we can merge?

docs/internal-documentation/distrib-testing/Debian-Testing-Instructions.md
sandoche2k: For me it is a library

Issue #450, opened 2016-02-10
Erroneous idea of what is a domain name


bortzmeyer:
[Reported by Jacques Latour, on an IETF mailing list.]

When testing the (delegated and fully functional) domain _sips._tcp.cam.ac.uk, the CLI reports:

```
  16.33 ERROR     Found illegal characters in the domain name (_sips._tcp.cam.ac.uk).
```

which is wrong: _ is perfectly legal for a _domain_ name. It seems Zonemaster assumes a domain name is always a host name, which is not true.

The Web interface says the same thing, plus it contains a typo: "Domain name name contains non-allowed character(s)" It is a blocking error, meaning no test is possible.

pawal:
I think it is clear that few registries would allow the registration of such domain name. However, they seem to have a proper use case further down the DNS hierarchy. I would have hoped to stay away from the dbound issues... Perhaps raise this issue on the list? (cam.ac.uk is within another administrative domain than ac.uk which is the registry.)

sandoche2k:
As per RFC 2181, it is a valid domain name and our pre-flight check should take this into account @vlevigneron

matsduf:
As of RFC 2181 any octet can be used for labels in a domain name. "[...] any binary string whatever can be used as the label of any resource record". If we accept to remove the restriction on underscore, then what is the argument against removing any other restrictions?

pawal:
If you have a delegation in the DNS hierarchy that you want to test, why would not Zonemaster let you test it?

matsduf:
The ticket only says "remove the restriction of underscore". Do you suggest that any restrictions should be removed?

pawal:
It depends on what you want Zonemaster to do, and the distinction between host names and domain names. I want all domains that are allowed by DNS to be testable by Zonemaster. So if you have an answer to the previous question, that would be helpful.

matsduf:
I think Zonemaster is most useful and less confusing as a tool checking host names. If you add some garbage, it will protest in an understandable way. 

It does not mean that it could not be useful for non-host names. Underscore is probably the most common deviation from host names. I like the suggestion of @mtoma that it should be something that you can set in the profile, only host name syntax or additional characters.

I think that we can wait a while with the ticket.

pawal:
Then I suggest we let the TRTF IETF draft be worked on, and then later follow those requirements.

matsduf:
[...]

PR #479, opened 2016-11-18
Update basic00.md

matsduf:
More explicit definition of a valid domain name. Added explicit support for IDN. Added support for underscore requested in issue #450.

### DIFF
--- a/docs/specifications/tests/Basic-TP/basic00.md
+++ b/docs/specifications/tests/Basic-TP/basic00.md
@@ -6,25 +6,45 @@
 ### Objective
 
 In order to begin testing the domain name from the input must be a valid
-domain name. The domain name must follow the rules defined in section 2.1
-of [RFC 1123](https://tools.ietf.org/html/rfc1123#section-2.1). The
-objective of this test is to see if it is possible to have the domain name
+domain name. The domain name must either be a 
+
+1. a valid IDN name (Internationalized Domain Name) ([RFC 5890](https://tools.ietf.org/html/rfc5890#page-13), 
+section 2.3.2.3), or
+2. a valid ASCII domain.
+
+The ASCII domain name is valid if follows the rules defined in section 2.1
+of [RFC 1123](https://tools.ietf.org/html/rfc1123#section-2.1), i.e. only
+consists of the ASCII characters "a-z", "A-Z", "0-9", "." and "-" with the
+extension of "_", standardized for SRV records 
+([RFC 2782](https://tools.ietf.org/html/rfc2782)).
+
+A valid IDN name with non-ASCII code points can always be converted 
+to a valid ASCII domain meeting the requirement above.
+
+The objective of this test is to see if it is possible to have the domain name
 mapped into a DNS packet in order to proceed with further testing.
+
 
 ### Inputs
 
-The label of the domain name to be tested.
+The domain name to be tested.
 
 ### Ordered description of steps to be taken to execute the test case
 
-1. Remove any trailing dot from the input domain name.
-2. If the total length of the input is more than 253 octets, this test
+1. If the domain name has more than one tailing dot, the test case
+fails.
+2. Remove any trailing dot from the input domain name.
+3. If the domain name contains non-ASCII characters, try to convert
+it to an ASCII domain with the IDN rules. If the conversion fails,
+then this test case fails.
+4. If the domain name does not meet the requirement of an ASCII domain
+above, then this test case fails.
+5. If the total length of the input is more than 253 octets, this test
    case fails.
-3. Separate the input by the dividing it into labels separated by the "."
+6. Separate the input by the dividing it into labels separated by the "."
    character.
-4. If any of the labels from step 3 is more than 63 octets, this test
-   case fails.
-5. If any of the labels from step 3 is zero octets in length, this test
+7. If any of the labels from step 3 is more than 63 octets, this test case fails.
+8. If any of the labels from step 3 is zero octets in length, this test
    case fails.
 
 ### Outcome(s)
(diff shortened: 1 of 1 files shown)


docs/specifications/tests/Basic-TP/basic00.md
```diff
@@ -6,25 +6,45 @@
...
 ### Inputs
 
-The label of the domain name to be tested.
+The domain name to be tested.
 
 ### Ordered description of steps to be taken to execute the test case
 
-1. Remove any trailing dot from the input domain name.
-2. If the total length of the input is more than 253 octets, this test
+1. If the domain name has more than one tailing dot, the test case
```
pawal: Could be replaced by something that says that any label in the domain name must have a length that is longer than zero octets. Such as the added step 8, irregardless of whether the domain name has been converted from an ULABEL or not.

matsduf: You are right that step 1 partly covers what step 8 covers. Maybe it is better to change step 1 to "If the input domain name contains two or more consecutive dots, then this test case fails" and remove step 8. Then that simple error will be caught early.

sandoche2k:
@matsduf @vlevigneron @mtoma 

I had a fruitful discussion with @pawal regarding https://github.com/dotse/zonemaster/blob/master/docs/specifications/tests/Basic-TP/basic00.md. We have two issues with it:
1. IDN domains
2. case of underscore : "_sips._tcp.cam.ac.uk"

The case of underscore should be rectified in Syntax

For IDN's, @pawal proposes not to change the current https://github.com/dotse/zonemaster/blob/master/docs/specifications/tests/Basic-TP/basic00.md

Rather, have a separate function such as here : https://godoc.org/github.com/miekg/dns/idn

Hence, from the CLI if we want to test "zonemaster-cli --convert-to-punycode räksmörgås.example"

If the user just tests as "zonemaster-cli räksmörgås.example" without the convert function, he has an advertisement.

But, from the web interface,  we make it implicit. When the web interface encounters non ascii domain, it calls the punycode conversion function in the engine. 

Advantages of this approach: 
The engine core functionality is to test DNS delegation configuration. All other functionalities are outside. 

@bortzmeyer for your view on this approach.

bortzmeyer:
@sandoche2k Seems perfectly fine.

matsduf:
@bortzmeyer and @sandoche2k!

I have a vision of the Zonemaster software using explicit test requirements, Test Cases, to define how the software behave when domains are tested. For me, the Zonemaster software consists of the Engine, the CLI, the Backend and the GUI. My vision is that the Test Cases covers all parts of Zonemaster.

We have Test Case BASIC00 that restricts the allowable characters in the domain name to be tested and also put some lenght restrictions on the domain name. BASIC00 is only partly implemented in Engine (length restrictions), and that is inherited in CLI and Backend. GUI has its own implementation, which seems to follow BASIC00 with the exception of IDN.

Even though BASIC00 forbids IDN names as U-labels to be tested, CLI, Backend and GUI have implemented support for U-labels.

What is your suggestion for BASIC00? What characters should Zonemaster support? How do we test that? We could allow a larger set of characters than the traditional host name, but then we should know that it works. Where are the real (not theoretical) use cases for domain names beyond the set that I suggested?

Remember that BASIC00 is for Zonemaster, not just for the Engine.

pawal:
This is not how we wrote the specifications from the start. The specifications only cover the behaviour of the engine, and it has been continuing to do so so far.

Using the engine there are some applications, and applications in the core project is the CLI and the GUI. There are applications outside of the core project such as several different programs I wrote that is only using the engine - as it was designed, using the exposed Perl APIs. It would be sad to see a direction where the separation of the testing library and the core applications are mixed up, thus making it harder to write software using the test library.

There can still be applications that limits the scope of what should be tested, such as some TLDs that only allow the testing of their own TLD, or a completely different set of restrictions, in the same way that you can write your own custom policy. Limiting those things in the engine is putting restrictions on all applications, and that would be very unfortunate.

That said, I would advise to separate specifications in the core project from the engine and the applications. The applications should have their own specifications, not only because they are different in scope, but also to not limit others to write applications for other uses.

matsduf:
@pawal, BASIC00 is there and has not been fully implemented in the Engine. Most people that go to the Zonemaster repository and find the specifications will expect that they are applied for all parts of Zonemaster, included CLI, Backend and GUI.

I still want to see real use case where characters beyond what I suggested are allowed. And the character set code be expanded, but it must be tested and correctly documented.

pawal:
It is very simple, to test delegations such as these (or OTHER yet unknown variants):

```
$~>dig ns räksmörgås.tset.se @vic20.blipp.com

; <<>> DiG 9.10.3-P4-Debian <<>> ns räksmörgås.tset.se @vic20.blipp.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 62815
;; flags: qr rd; QUERY: 1, ANSWER: 0, AUTHORITY: 2, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;r\195\164ksm\195\182rg\195\165s.tset.se. IN NS

;; AUTHORITY SECTION:
r\195\164ksm\195\182rg\195\165s.tset.se. 86400 IN NS ns3.nic.se.
r\195\164ksm\195\182rg\195\165s.tset.se. 86400 IN NS ns.nic.se.

;; Query time: 4 msec
;; SERVER: 2001:16d8:ff00:2a9::2#53(2001:16d8:ff00:2a9::2)
;; WHEN: Fri Nov 25 19:11:24 CET 2016
;; MSG SIZE  rcvd: 89
```

pawal:
This is fine if it is possible to disable basic00 in any test that I want to run.

sandoche2k:
@matsduf As there seems to be a consensus, i think you can merge the updates for basic00. @vlevigneron is waiting for spec updates to start coding.

pawal:
There is no way to disable basic00, or did you fix this also? Otherwise, please revert this to previous behavior.

PR #487, opened 2016-12-09
Updates to FreeBSD preparation

mattias-p:
Adds:
1. Updating of the ports database
2. Installation of curl (a prerequisite for the backend sanity check instruction)

### DIFF
--- a/docs/internal-documentation/distrib-testing/FreeBSD-Preparation.md
+++ b/docs/internal-documentation/distrib-testing/FreeBSD-Preparation.md
@@ -1,8 +1,19 @@
 # FreeBSD system preparation for Zonemaster
 
-1. Make a clean installation of FreeBSD.
+1. Use latest patch level and ports database
 
-2. Create a new user.
+   ```sh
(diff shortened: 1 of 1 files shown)


docs/internal-documentation/distrib-testing/FreeBSD-Preparation.md
```diff
@@ -18,10 +18,11 @@
    visudo
```
matsduf: Either sudo has to be installed, or use good old su.

mattias-p:
I believe you're already root at this point. It's a clean installation.

docs/internal-documentation/distrib-testing/FreeBSD-Preparation.md
mattias-p: Good point. I'm opting to install `sudo`. This way it's easier to have common instructions across operating systems for system testing.

matsduf:
All steps in this document assumes that you are root. I guess it is OK that it is implicitly so for now.

PR #488, opened 2016-12-09
Specify how to install tarballs in system testing

mattias-p:
I chose the `cpanm` approach because it works. Its only drawback, that I know of, is that it adds another prerequisite to system testing.

Alternatives:

Install using `make` like this:
```sh
tar xzvf ${MODULE}-${MODULE_VERSION}.tar.gz
cd ${MODULE}-${MODULE_VERSION}
perl Makefile.PL
make
sudo make install
```
The problem with the `make` approach is that it doesn't install the dependencies listed in Makefile.PL. I dug up [this link](http://www.dagolden.com/index.php/1528/five-ways-to-install-modules-prereqs-by-hand/) and tried `make installdeps` (on FreeBSD 10.3) but that target doesn't seem to be generated by Makefile.PL.

Install using `cpan` like this:
```sh
tar xzvf ${MODULE}-${MODULE_VERSION}.tar.gz
cd ${MODULE}-${MODULE_VERSION}
sudo cpan .
```
The problem with the `cpan` approach is that it insists on installing "recommended" dependencies and blows up for Backend when MySQL isn't installed.

### DIFF
--- a/docs/internal-documentation/maintenance/SystemTesting.md
+++ b/docs/internal-documentation/maintenance/SystemTesting.md
@@ -59,12 +59,22 @@ The set of configurations must include at least:
    2. Install Zonemaster Engine
       1. Follow the [dependencies](https://github.com/dotse/zonemaster-engine/blob/master/docs/installation.md#dependencies) section of the installation guide to the letter.
       2. Install the preliminary distribution tarball for zonemaster-engine.
+
+         ```sh
+         sudo cpanm Zonemaster-${ENGINE_VERSION}.tar.gz
+         ```
+
       3. Follow the [post-installation sanity check](https://github.com/dotse/zonemaster-engine/blob/master/docs/installation.md#post-installation-sanity-check) section of the installation guide to the letter.
 
    3. Install Zonemaster Backend
       1. Follow the [dependencies](https://github.com/dotse/zonemaster-backend/blob/master/docs/installation.md#dependencies)
          section of the installation guide to the letter.
       2. Install the preliminary distribution tarball for zonemaster-backend.
+
+         ```sh
+         sudo cpanm Zonemaster-WebBackend-${BACKEND_VERSION}.tar.gz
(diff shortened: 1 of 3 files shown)


docs/internal-documentation/maintenance/SystemTesting.md
```diff
@@ -59,12 +59,18 @@ The set of configurations must include at least:
    2. Install Zonemaster Engine
       1. Follow the [dependencies](https://github.com/dotse/zonemaster-engine/blob/master/docs/installation.md#dependencies) section of the installation guide to the letter.
       2. Install the preliminary distribution tarball for zonemaster-engine.
+         ```sh
+         sudo cpanm Zonemaster-${ENGINE_VERSION}.tar.gz
+         ```
```
matsduf: It would be better if the sudo-command was presented on its own line, if possible.

sandoche2k:
Did you verify that when you use `sudo cpanm ...tar.gz` file it installs the tarball?

My tests shows that on using cpan -i it was installing from CPAN. Just try to rename the file and check whether you can install.

sandoche2k:
sudo cpanm does not work in Ubuntu/Debian unless you have $PATH added.

mattias-p:
@sandoche2k Unless you have $PATH added to what?

Renaming the file doesn't seem to be a problem for `cpanm`.
```sh
$ cp Zonemaster-v1.0.16.tar.gz test.tgz
$ sudo cpanm test.tgz
--> Working on test.tgz
Fetching file:///usr/home/mattiasp/test.tgz ... OK
Configuring Zonemaster-v1.0.16 ... OK
Building and testing Zonemaster-v1.0.16 ... ^C
$
```

mattias-p:
@matsduf I think the instruction to install `cpanm` should go into the OS-preparation instructions. I'll include the changes for that in this PR but not until #487 is closed.

PR #501, opened 2017-04-27
Update delegation01.md

matsduf:
Clarified how name servers should be counted.


mattias-p:
I think this looks fine.

PR #519, opened 2017-10-27
Update README.md

matsduf:
Updated the table with supported db engines per OS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,17 +61,16 @@ Operating System | MySQL | PostgreSQL
 ---------------- | ------| -----------
 CentOS 7         | 5.6   |   (9.3)
 Debian 7.11      | 5.5   |   (9.3)
-Debian 8.9       | 5.5   |   9.4
+Debian 8.9       | 5.5   |   (9.4)
 FreeBSD 10.3     | (5.6) |   9.5       
 FreeBSD 11.1     | (5.6) |   9.5      
 Ubuntu 14.04     | 5.5   |   9.3
-Ubuntu 16.04     | 5.7   |   9.5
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,15 +57,15 @@ before we can include support for Debian 9._
...
-CentOS 7         |   5.5    |       |   9.2
-Debian 7.11      |          | 5.5   |   9.1
-Debian 8.9       |          | 5.5   |   9.4
-FreeBSD 10.3     |          | 5.6   |   9.5       
-FreeBSD 11.1     |          | 5.6   |   9.5      
-Ubuntu 14.04     |          | 5.5   |   9.3
-Ubuntu 16.04     |          | 5.7   |   9.5
+Operating System | MySQL | PostgreSQL
+---------------- | ------| -----------
+CentOS 7         | 5.6   |   -
```
mattias-p: Should we say something about why we don't support PostgreSQL on CentOS?

README.md
```diff
@@ -59,14 +59,20 @@ before we can include support for Debian 9._
...
+FreeBSD 11.1     | (5.6) |   9.5      
 Ubuntu 14.04     | 5.5   |   9.3
 Ubuntu 16.04     | 5.7   |   9.5
 
+_PostgreSQL has not been tested on CentOS and Debian 7.11, but we have no reason
+to beleive that the version listed for the OS would not work._
+
+_MySQL has not been tested on FreeBSD, but we have not reason to beleive that
+that the version listed for FreeBSD would not work._
+
```
mattias-p: "beleive" -> "believe", twice
"the version listed for the OS" -> "it", twice
"not reason" -> "no reason"
"that that" -> "that"

PR #530, opened 2017-11-17
Updated delegation 02

sandoche2k:


matsduf:
@sandoche2k, I suggest that we structure this test case differently. I think we should have distinct test cases that cover one thing instead of encapsulating several things. The test case is sorted under DELEGATION and the title is "Name servers must have distinct IP addresses".

I suggest:

* DELEGATION02 (this test case) should only test if there is an overlap of IP addresses between the nameservers (NS) in the delegation and it should disregard the nameserers (NS) in the zone. 
* Testing the numbers of nameservers in the delegation is already covered in DELEGATION01.
* The nameservers in the zone is not part of the delegation. Create a new test case, ZONExx, that tests if there is an overlap of IP addresses in the nameservers (NS) of the zone.
* Create a new test case, ZONExx, that tests if the number of NS in the zone is less than two.

sandoche2k:
The origin to the problem is in this test : https://zonemaster.net/test/d3c11a3298cdb7d1
In delegation: 
Parent list two NS : ns1.rayceem.com; ns2.rayceem.com 
Child list two NS : ns1.rayceem.com; ns2.rayceem.com 

Delegation01 passes, since there are two NS. 

The issue is that both the NS point to the same IP and the "NOTICE" result from zonemaster is : 

"IP 185.186.78.110 refers to multiple nameservers (ns1.rayceem.com; ns2.rayceem.com)." 

When we check from dnscheck, the result for delegation is : 

"Only one IPv4 name server was found for the zone. You should always have at least two IPv4 name servers for a zone to be able to handle transient connectivity problems."

The objective of delegation02 is "If the domain have several different name server names used, they can all be using the same IP address. This may be due to a configuration error, or
a workaround to a certain policy restriction. This test case checks that the name servers used does not resolve to reuse the same IP addresses."

So, my suggestion is that updating from "Notice" to "Error" and providing an appropriate clear message as in dnscheck.

sandoche2k:
Close the PR and create an issue to change the message of delegation02 from "Notice" to "Error"

Issue #498, opened 2017-02-21
Create documentation of config/policy/profile


matsduf:
There is a need of documentation of format and place of config, policy and profile. Engine uses the config and policy, but can be selected by CLI. Backend can select policy (but called "profile" in API). GUI can also expose a selection of policies (called "profiles").

A work has been started to unify these files and to make config selectable in Backend API.

Since config and policy are shared across Engine, CLI and Backend, documentation is needed here, with references from Engine, CLI and Backend.

matsduf:
We have decided to merge config and policy into a profile.

mattias-p:
There is a WIP at https://github.com/dotse/zonemaster/tree/profiles

sandoche2k:
@matsduf closed?

sandoche2k:
@mattias-p close?

PR #541, opened 2017-12-15
Profiles doc

mattias-p:
Partial fix for #498.
Updates to Engine and Backend are made in separate PRs.

### DIFF
--- a/docs/design/Profiles.md
+++ b/docs/design/Profiles.md
@@ -5,13 +5,16 @@
 Profiles is a configuration format and feature of Zonemaster Engine.
 Each of Zonemaster CLI, Backend and GUI allow users and administrators to configure Zonemaster Engine with different profiles.
 
+This document provides a high-level view of how profiles are applied across the different Zonemaster components.
+For details on how profiles integrate with a given component, refer to the respective component documentations.
+Links to the relevant sections are provided below.
+
 
 ## Background
 
 The Zonemaster requirements and specification documents speak of user-configurable profiles and policies for Zonemaster Engine and CLI.
 
 In release 2017.4 all of the configurable features "config", "filter", "policy" and "profile" from the requirements and specifications were integrated into a single data structure, also called "profile".
-This document provides a high-level view of the new "profile" feature.
 
 
 ## Profiles in Zonemaster Engine
--- a/docs/design/Profiles.md
+++ b/docs/design/Profiles.md
@@ -1,0 +1,52 @@
+# Profiles
+
+## Overview
+
+Profiles is a configuration format and feature of Zonemaster Engine.
+Each of Zonemaster CLI, Backend and GUI allow users and administrators to configure Zonemaster Engine with different profiles.
+
+
+## Background
+
+The Zonemaster requirements and specification documents speak of user-configurable profiles and policies for Zonemaster Engine and CLI.
+The Profiles feature integrates both of those into a single data structure.
+
+
+## Profiles in Zonemaster Engine
+
+Zonemaster Engine has an effective profile that guides how it performs its tests and analyzes their results.
+It comes with a sensible default profile while allowing applications to override the default behvior.
+This is described in greater detail in the [Zonemaster::Engine::Config] documentation.
+
+
+## Profiles in Zonemaster Backend
+
+Zonemaster Backend allows its administrators to configure available profiles for its users to choose from.
+Configuration is described in greater detail in the [Zonemaster Backend configuration] documentation.
+The [Zonemaster Backend RPC-API] documentation lists methods that accept profile name arguments.
+
+
+## Profiles in Zonemaster CLI
+
+Zonemaster CLI allows users to specify the path to a profile to be used when starting a test.
+If no profile is specified, the Zonemaster Engine default profile is used.
+
+
+## Profiles in Zonemaster GUI
+
+Zonemaster GUI allows the user select profile from a set of choices.
+The available profile choices are configured in a JSON file that maps profile name strings to description strings.
+
+
+-------
+[Zonemaster Backend RPC-API]: https://github.com/dotse/zonemaster-backend/blob/master/docs/API.md
+[Zonemaster Backend configuration]: https://github.com/dotse/zonemaster-backend/blob/master/docs/Configuration.md#profiles-section
+[Zonemaster::Engine::Config]: https://metacpan.org/pod/Zonemaster::Engine::Config
+
+
+Copyright (c) 2017, IIS (The Internet Foundation in Sweden)\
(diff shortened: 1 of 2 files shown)


docs/design/Profiles.md
```diff
@@ -0,0 +1,52 @@
+# Profiles
+
+## Overview
+
+Profiles is a configuration format and feature of Zonemaster Engine.
+Each of Zonemaster CLI, Backend and GUI allow users and administrators to configure Zonemaster Engine with different profiles.
```
matsduf: I think we should have a comment here that what was config and policy in previous versions of Engine and CLI have been integrated into profile. And that what was profile in previous version of Backend has now been expanded to be a full configuration of Zonemaster Engine.

mattias-p: I can add that to the Background section as a historical note. Also mentioning that the profiles feature was added in 2017.4. Are you okay with that?

matsduf: Yes, that sounds good.

mattias-p: I only rephrased the background section a bit. I opted not to add details about the relation between profiles and the various components. I'll add things to the component sections if there's anything more that needs to be highlighted.

mattias-p:
@sandoche2k Do you have any comments on this?

docs/design/Profiles.md
```diff
@@ -0,0 +1,53 @@
...
+
+
+## Background
+
+The Zonemaster requirements and specification documents speak of user-configurable profiles and policies for Zonemaster Engine and CLI.
+
+In release 2017.4 the profile and policy concept from the requirements and specifications were integrated into single data structure, also called a profile.
+
```
sandoche2k: Suggest to modify the above text as follows:
In release 2017.4 all earlier configurable features such as config,policy, filters and profile concept from the requirements and specifications were integrated into single data structure, also called a profile.

docs/design/Profiles.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Profiles in Zonemaster Engine
+
+Zonemaster Engine has an effective profile that guides how it performs its tests and analyzes their results.
+It comes with a sensible default profile while allowing applications to override the default behvior.
```
sandoche2k: behvior => behavior

sandoche2k:
@mattias-p i have submitted my comments on the content. From the contents already updated, i am not sure how you are going to organise the files. My point of view is that : 
In the Zonemaster repository, we provide an overview of the different configurable features for Zonemaster and link to the respective respositories in the backend, CLI and the GUI on how to override the default features for the user? Also, we need to have documentation on how the overriding profile file is loaded. 
This is all in terms with the number of discusssions that we had.

mattias-p:
@sandoche2k Thanks for reviewing! I improved the documentation in response to your comments.

I agree with your view on how the profiles documentation should be organized, and I updated the overview section to make this clear.

I've provided updates to the Engine and Backend documentations in dotse/zonemaster-engine#361 and dotse/zonemaster-backend#332.

I have not written detailed documentations for CLI and GUI. Do you have any ideas on whether those are required and how they should be organized?

sandoche2k:
@matsduf, @mattias-p I feel that we should simplify the documentation. Someone who is using the profile is supposed to be someone who is knowledgeable in DNS as well as system admin.

We should have a simple one file, wherein we give one or two examples on how one can override the default configurations.

matsduf:
@sandoche2k, I do not understand what you mean by "a simple one file". The profile is quite complex and we need a clear specification of it so that we know if we have implemented it correctly. And we also must give the users a chance to understand how it works so that they can create overrides.

sandoche2k:
I understand. But from an user perspective we should make it easy for him
to implement

matsduf:
@sandoche2k, have you looked at the PR under Engine, dotse/zonemaster-engine#361? There you can find the details needed for implementation.

sandoche2k:
@matsduf  i approve

sandoche2k:
@mattias-p  i think you can merge the PR. This will make it easier to review further.

mattias-p:
@sandoche2k @matsduf I think the profiles feature should be developed on feature branches in the respective repos. Then once everything is completed we can merge all of them into the develop branches at the same time. The upside is that completion of the profiles feature doesn't have to be a blocker for the next release. In my opinion this is a big win. The downside is that the develop branches needs to be merged into the feature branches regularly. In my opinion this is a small price - I can do it if no one else volunteers.

sandoche2k:
@mattias-p I agree with your proposal. But, i think we should focus to have the profile documentation @ implementation completed during the current release.

mattias-p:
@sandoche2k I agree that it would be very nice to have it done for the next release. But time flies and this feature needs changes in lots of places. It also requires some effort around backwards compatibility so we don't break Travis for CLI and Backend when updating the Engine API. Finally I've been working a lot with this lately but in the near future I need to focus on another project. I'm not saying it can't be done for the next release - I just think it makes sense to have the escape hatch.

sandoche2k:
Thanks for the input. We will try and if it is not possible, then we will
move to the next release.

mattias-p:
I've created "profiles" branches for zonemaster, zonemaster-engine, zonemaster-backend and zonemaster-cli, and updated the target of all the "profiles-doc" PRs to point to those.

docs/design/Profiles.md
```diff
@@ -0,0 +1,57 @@
...
+This document provides a high-level view of how profiles are applied across the different Zonemaster components.
+For details on how profiles integrate with a given component, refer to the respective component documentations.
+Links to the relevant sections are provided below.
+
+
+## Background
+
+The Zonemaster requirements and specification documents speak of user-configurable profiles and policies for Zonemaster Engine and CLI.
+
+In release 2017.4 all of the configurable features "config", "filter", "policy" and "profile" from the requirements and specifications were integrated into a single data structure, also called "profile".
```
matsduf: An update is needed here. May some placeholder instead of a version?

matsduf: @mattias-p, I second your proposal to have the profile change merged into a feature branch. It makes most sense.

mattias-p: Maybe "Previously"? The proper place for documenting changes between releases is the Changes file anyway.

matsduf: I suggest that >>In release ... called "profile".<< is removed and that we make sure there is an issue describing the the change and we link to that from CHANGES.

matsduf:
@sandoche2k, have you approved? If so, please press the approve button.

sandoche2k:
@mattias-p I think you do not have to wait until all the reviewers approve it. You can go ahead and merge.

PR #546, opened 2018-01-05
Update the procedure for verifying metadata

mattias-p:
META.yml contains all the metadata this step is intended to verify. Makefile.PL contains only part of it, but instead lots of other things. In other words, META.yml is a better place to check metadata.

This change also clarifies what to look for in the metadata.


mattias-p:
@matsduf It has no effect on testing. It just clarifies how to check the CPAN metadata. The current instruction is rather vague.

PR #547, opened 2018-01-05
Update README on issue reporting plus adding information on CLI locale problem

matsduf:
Clearified how to report issues and adding reference to missing CLI repository. Adding "notable bugs and issues" and in that adding information about problem with `--locale` option in CLI.

PR #548, opened 2018-01-08
Update README on supported OSs and DBs

matsduf:
Removed Debian 8 as supported OS. Added support for both PostgreSQL and MySQL on all supported OSs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,32 +45,27 @@ architecture listed below.
 
 * CentOS 7
 * Debian 7
-* Debian 8
+* (Debian 8)
 * (Debian 9)
 * FreeBSD 10.3
 * FreeBSD 11.1
 * Ubuntu 14.04
 * Ubuntu 16.04
 
-_Currently, Debian 9 is not supported due to a known bug in Zonemaster::LDNS. That has to be resolved 
-before we can include support for Debian 9._
+_Currently, Debian 8 and Debian 9 are not supported due to a known bug in Zonemaster::LDNS, see [Zonemaster-LDNS issue 10] and [Zonemaster-LDNS issue 30]. 
+That has to be resolved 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -168,3 +162,7 @@ For contacting the Zonemaster project, please send a mail to
 [zonemaster-gui]: https://github.com/dotse/zonemaster-gui
 [ldns]: https://www.nlnetlabs.nl/projects/ldns/
 [CPAN]: http://search.cpan.org/search?query=Zonemaster&mode=dist
+
+[Zonemaster-LDNS issue 10]: https://github.com/dotse/zonemaster-ldns/issues/10
+[Zonemaster-LDNS issue 30]: https://github.com/dotse/zonemaster-ldns/issues/30. 
+
```
sandoche2k: Why do we need to add the LDNS issue here, keeping in mind that Debian 8 is removed from the supporting OS

matsduf: I have created a reference that I use on line 55.

pawal:
I don't have either MySQL or PostgreSQL installed on my Zonemaster installations, as I don't use the backend. Hopefully you will still support Zonemaster Engine and CLI on Debian 8.

It should be noted that only the backend requires any of the databases.

mattias-p:
This PR has the same title as #547. Please give PRs more descriptive titles.

matsduf:
@mattias-p! I have updated  this PR #548 and also PR #547 with better titles.

matsduf:
@pawal! This issue on Debian 8 and Debian 9 is related to issues on Zonemaster-LDNS, see https://github.com/dotse/zonemaster-ldns/issues/10 and https://github.com/dotse/zonemaster-ldns/issues/30. Without Zonemaster-LDNS you cannot install Zonemaster-Engine.

PR #560, opened 2018-01-29
Several corrections (README.md)

JohanIIS:
@matsduf Minor adjustments to get started. Will take a lengthier look at this document in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,14 +44,12 @@ architecture listed below.
 
 * CentOS 7
 * Debian 7
+* Debian 8
+* Debian 9
 * FreeBSD 10.3
 * FreeBSD 11.1
 * Ubuntu 14.04
 * Ubuntu 16.04
-
-_Currently, Debian 8 and Debian 9 are not supported due to a known bug in Zonemaster::LDNS, see [Zonemaster-LDNS issue 10] and [Zonemaster-LDNS issue 30]. 
-These have to be resolved 
-before we can include support for Debian 8 and Debian 9._
 
 ### Supported database engine versions
 
@@ -78,8 +76,7 @@ FreeBSD 11.1     | 5.24
 Ubuntu 14.04     | 5.18
 Ubuntu 16.04     | 5.22
 
-Zonemaster has been tested with the combination of OS and Perl version listed in the table
-above. Perl versions earlier than 5.14.2 are not supported.
+Zonemaster requieres Perl version 5.14.2 or higher. Zonemaster has been tested with the default version of Perl in the OSs as listed in the table above.
 
 ## Localization
 
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@
 ## Background
 
 DNSCheck from IIS and Zonecheck from AFNIC are two different software
-packages that do DNS validation of the quality of a DNS
-delegation. AFNIC and IIS decided to develop a new DNS validation tool from the
-scratch under the name "Zonemaster". 
+packages that validates the quality of a DNS
+delegation. AFNIC and IIS came together to develop a new DNS validation tool from
+scratch under the name Zonemaster. 
 
 The Zonemaster implementation intends to be a major
 rewrite of the existing DNS validation tools developed by AFNIC (Zonecheck) and
@@ -16,25 +16,24 @@ IIS (DNSCheck), and implement the best parts of both.
 
 The components developed as part of the Zonemaster project will help different
 types of [users](USING.md) to check domain servers for configuration errors and
-generate a report that will enable to fix the errors.
+generate a report that will assist in fixing the errors.
 
 The ambition of the Zonemaster project is to develop and maintain an open source
-DNS validation tool, offering better performance than the existing tools and
-provide extensive documentation which could be re-used by similar projects in
+DNS validation tool, offering improved performance over existing tools and
+providing extensive documentation which could be re-used by similar projects in
 the future.
 
 ## Documentation
 
-The repository you are looking at is the main project repository. In this
+This is the main project repository. In this
 repository, documentation regarding the [design](docs/design),
 [requirements](docs/requirements) and [specifications](docs/specifications)
-for the zonemaster implementation are available.
-
-Also, in this repository, you can find a brief [user guide](USING.md).
+for the Zonemaster implementation are available. 
+We also have a brief [user guide](USING.md).
 
 ## Prerequisites
 
-Zonemaster comes with documentation for and has been tested on the operative systems and processor
+Zonemaster comes with documentation for and has been tested on the operating systems and processor
 architecture listed below.
 
 ### Supported processor architectures
@@ -45,16 +44,14 @@ architecture listed below.
 
 * CentOS 7
 * Debian 7
-* (Debian 8)
-* (Debian 9)
 * FreeBSD 10.3
 * FreeBSD 11.1
 * Ubuntu 14.04
 * Ubuntu 16.04
 
 _Currently, Debian 8 and Debian 9 are not supported due to a known bug in Zonemaster::LDNS, see [Zonemaster-LDNS issue 10] and [Zonemaster-LDNS issue 30]. 
-That has to be resolved 
-before we can include support for Debian 8 and Debian, respectively._
+These have to be resolved 
+before we can include support for Debian 8 and Debian 9._
 
 ### Supported database engine versions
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,16 +44,14 @@ architecture listed below.
 
 * CentOS 7
 * Debian 7
-* (Debian 8)
-* (Debian 9)
```
matsduf: They are only temporarily not supported and should be listed here.

README.md
```diff
@@ -45,16 +44,14 @@ architecture listed below.
...
 * FreeBSD 10.3
 * FreeBSD 11.1
 * Ubuntu 14.04
 * Ubuntu 16.04
 
 _Currently, Debian 8 and Debian 9 are not supported due to a known bug in Zonemaster::LDNS, see [Zonemaster-LDNS issue 10] and [Zonemaster-LDNS issue 30]. 
-That has to be resolved 
-before we can include support for Debian 8 and Debian, respectively._
+These have to be resolved
```
matsduf: "That" refer to the know bug (one) rather than the OSs (two). On the other hand, the bug seems to be resolved so this section should probably be removed when we update.

README.md
```diff
@@ -83,7 +79,7 @@ Ubuntu 14.04     | 5.18
 Ubuntu 16.04     | 5.22
 
 Zonemaster has been tested with the combination of OS and Perl version listed in the table
-above. Other versions of Perl might work. Perl versions earlier than 5.14.2 will _NOT_ work.
+above. Perl versions earlier than 5.14.2 are not supported.
```
matsduf: This changes the meaning. Perl versions earlier than 5.14.2 will not work. Perl version 5.22 has been tested to work with Ubuntu 16.04, but not with Ubuntu 14.04, but it might work with Ubuntu 14.04.

mattias-p: @matsduf What do your think of this wording? "Perl versions earlier than 5.14.2 are explicitly not supported." That clearly implies that there's a gray area that may or may not work.

matsduf: @mattias-p, I think that wording is harder to interpret than the current text. What is the problem with the current text?

README.md
JohanIIS: I just felt that OS support should really be binary; an OS version either is or isn't supported in the current public release. Bugs and/or planned support should be listed elsewhere.

README.md
JohanIIS: A theme that will be recurring in all of my edits is that I want to see specificity. Phrasings like "might work" are imprecise and don't convey anything of substance to the user. In this case, and if a user wants to give a different combination of OS/Perl a shot, there isn't really anything we can do to help them if we haven't tested and confirmed it ourselves. They can choose to try it or not, and a line from us about how they can do exactly that doesn't really provide any actual assistance.

README.md
matsduf: It is not planned support, it is temporarily not supported. Those are the OSs that we support, and by support principles we should support Debian 8 and Debian 9, but we could not. It is cleaner to list it here.

README.md
matsduf: We signal that it is worth trying other versions of Perl, and especially other OS/Perl version combinations, but nothing lower than Perl 5.14.2.

matsduf: @JohanIIS and @mattias-p, I suggest:

Zonemaster requieres Perl version 5.14.2 or higher. Zonemaster has been tested with the default version of Perl in the OSs as listed in the table above.

README.md
mattias-p: Should we really keep two issues open for a single bug? Wouldn't it be better to close one of them as a duplicate of the other?

README.md
mattias-p: The dropped support for Debian 8 and 9 is permanent for this particular release, and the documentation is part of the release. In this light I think we should list only the OS versions we actually do support for this particular release. If we feel squirmy about removing important OS versions from the list, we could formulate an excuse below the list.

We do have a process for updating the this list for the next release, so we should not worry about forgetting to add them back. The issues in question are tagged with "High priority", so we should not worry about forgetting to fix them for the next release either.

matsduf: This PR is for the develop branch, which will be merged to the master branch at next release. It is too early to say what OSs that we will successfully suppport then. The best guess is that we will successfully support Debian 8 and Debian 9. There is no reason to do any updates now. That should be done closer to release.

From [Support criteria](https://github.com/dotse/zonemaster/blob/master/docs/internal-documentation/maintenance/SupportCriteria.md) you can conclude that we should support Debian 8 and Debian 9, and then I think we should list them in the README. It could be listed differently than in parenthesis, but not just removed.

mattias-p: Fair enough as far as I'm concerned with regard to the situation at hand.

For future releases, do you agree with the reasoning in my previous comment? In that case we could update the [release process](https://github.com/dotse/zonemaster/blob/master/docs/internal-documentation/maintenance/ReleaseProcess.md#3-update-prerequisites) document with a guideline for temporarily dropped support.

matsduf: @mattias-p, I think it was correct to list Debian 8 and Debian 9 with a note that they could not be supported (even though they were on the list of OSs that we had planned to support). That could be done in different ways, but not just removing the "Debian 8" and "Debian 9" as if they were not meant to be supported.

matsduf:
@sandoche2k, please review this PR.

README.md
```diff
@@ -4,9 +4,9 @@
 ## Background
 
 DNSCheck from IIS and Zonecheck from AFNIC are two different software
-packages that do DNS validation of the quality of a DNS
-delegation. AFNIC and IIS decided to develop a new DNS validation tool from the
-scratch under the name "Zonemaster". 
+packages that validates the quality of a DNS
+delegation. AFNIC and IIS came together to develop a new DNS validation tool from
+scratch under the name Zonemaster.
```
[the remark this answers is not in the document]
matsduf: Yes and no. They partly say the same thing, partly contradict each other ("from scratch" vs. "rewrite of the existing") and partly support each other.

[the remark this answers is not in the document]
mattias-p: @sandoche2k Good catch!

@matsduf I agree.

Suggestion:

DNSCheck from IIS and Zonecheck from AFNIC are two different tools for validateing the quality of a DNS delegation. AFNIC and IIS came together to develop a new DNS validation tool from scratch. The result is Zonemaster which reimplements the best parts of both Zonecheck and DNSCheck.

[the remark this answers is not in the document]
JohanIIS: @mattias-p Your suggestion is a clear improvement, but I think the very first thing mentioned at the top of the page should be Zonemaster and not DNSCheck/Zonecheck. 

How would you (or @matsduf or @sandoche2k) describe Zonemaster in one (or two) sentence(s)?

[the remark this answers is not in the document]
matsduf: This is my suggestion:

>DNSCheck from IIS and Zonecheck from AFNIC are two tools that validates the quality of a DNS delegation. Both tools are old and needed a replacement. AFNIC and IIS came together to develop a new DNS validation tool under the name Zonemaster. Zonemaster is written from scratch, but the implementation has inherited DNS requirements and experience from Zonecheck and DNSCheck.

sandoche2k:
@JohanIIS I am Ok with @matsduf suggestion. Can you check and merge ?

matsduf:
@JohanIIS, please push an updated version.

matsduf:
Proposal of new start of the README.md file:

## Introduction

Zonemaster is a software packages that validates the quality of a DNS delegation. The ambition of the Zonemaster project is to develop and maintain an open source DNS validation tool, offering improved performance over existing tools and providing extensive documentation which could be re-used by similar projects in the future.

Zonemaster consists of several modules or components. The components will help different types of [users](USING.md) to check domain servers for configuration errors and generate a report that will assist in fixing the errors.

## Background

DNSCheck from IIS and Zonecheck from AFNIC are two old software packages that validates the quality of a DNS delegation. AFNIC and IIS came together to develop a new DNS validation tool from scratch under the name **Zonemaster**. Zonemaster intends to be a major rewrite of Zonecheck DNSCheck, and implement the best parts of both. 

## Documentation

(...)

matsduf:
@JohanIIS, @mattias-p and @sandoche2k! Please review my proposal.

JohanIIS:
Added new commit with Mats' introduction suggestion.

JohanIIS:
@matsduf Everyone seems to be in agreement, please merge.

PR #561, opened 2018-02-08
Create StyleGuide.md

JohanIIS:
Proposal for early revision of style guide. Will be much expanded as more phrases are nailed down. Only one point so far.

### DIFF
--- a/docs/internal-documentation/maintenance/StyleGuide.md
+++ b/docs/internal-documentation/maintenance/StyleGuide.md
@@ -1,0 +1,9 @@
+# Style Guide
+
+## Scope
+
+The following style guide applies only to document text and certain file names specified below.
+
+## Guidelines
+
+* Zonemaster is always written as **Zonemaster**, never shortened (e.g. "ZM") or uncapitalized (e.g. "zonemaster"). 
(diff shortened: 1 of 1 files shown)


docs/internal-documentation/maintenance/StyleGuide.md
```diff
@@ -0,0 +1,9 @@
+# Style Guide
+
+## Scope
+
+The following style guide applies only to document text and certain file names specified below.
```
matsduf: Does "document text" include POD documentation withing Perl modules and scripts? I think it should.

JohanIIS: I agree. It would probably be good for this document to include a full list of all types of files that should or shouldn't adhere to style rules, but I don't yet have insight into what they are.  More suggestions (or edits to the document itself) are needed.

matsduf: I suggest that "only" is removed not to give an impression that this should be interpret restrictive.

JohanIIS: It was originally intended to be restrictive but removing the word still works. Change committed. The idea was that certain files will be exempt from style rules for various reasons, but that will become clear as more files/types are added to the guide.

sandoche2k:
@JohanIIS please merge

PR #563, opened 2018-02-09
Update README.md

JohanIIS:
Fixed image link and adjusted text for syntax.


JohanIIS:
I agree that it might seem pointless to have a folder and README for one single file, and I think I prefer the idea of moving it up one level instead of renaming it.

mattias-p:
If the file is moved up one level, should it also be renamed "Requirements.md" to preserve the directory structure?

matsduf:
@sandoche2k, what do you say about my proposal?

sandoche2k:
@JohanIIS Can you merge the PR?

PR #565, opened 2018-02-14
Tweak release process

mattias-p:
Since we're using semantic versioning the version number position to bump is determined by the nature of the changes. In this light it only makes sense that the change log is updated before the version number.

Also, the current instructions for remaking the preliminary distribution tarballs doesn't really make sense. Some previous reorganization of the document probably failed to update this.


mattias-p:
@matsduf Updated.

@sandoche2k While I generally approve of your suggestion I'd rather not expand this particular PR to include new things. If you make PR for the ToDo I'll review it.

PR #566, opened 2018-02-22
Preparation for version v2017.4.2

matsduf:
Preparation for version v2017.4.2
* Update Changes.
* Update README.md.

### DIFF
--- a/Changes
+++ b/Changes
@@ -5,7 +5,7 @@ v2017.4.2 2018-02-23 Security release
 Changes:
 
 - Security release of zonemaster-gui
-- Update of validation in zonemaster-backend
+- Update of input validation of domain name in zonemaster-backend
 
 This version of Zonemaster also consists of the following components. For each
 component, see its Changes file.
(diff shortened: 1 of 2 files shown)


Changes
```diff
@@ -1,5 +1,22 @@
 Release history of Zonemaster
 
+v2017.4.2 2018-02-23 Security release
+
+Changes:
+
+- Security release of zonemaster-gui
+- Update of validation in zonemaster-backend
```
sandoche2k: Update of domain input validation in zonemaster-backend

Changes
```diff
@@ -1,5 +1,22 @@
 Release history of Zonemaster
 
+v2017.4.2 2018-02-23 Security release
+
```
sandoche2k: Shouldn't it be v2018.1?

As per https://github.com/dotse/zonemaster/blob/b461694b841121b2a0bf9a3b847fdb8d72f10857/docs/design/Versions%20and%20Releases.md
"Version Number Syntax for Zonemaster product"

matsduf: This is an update outside the plan. We have only patching of two components. We could make this v2018.1, but then we have to change the plan too.

Both solutions are OK for me.

sandoche2k:
@matsduf i approve for 2017.4.2

PR #570, opened 2018-03-01
Update test messages

### LINKED ISSUE #569: TestMessages.md not updated
The file TestMessages.md lists all messages that can be emitted "from the test cases", but it has not been updated. It should be considered to be the authoritative source, not the code.

* Consider moving the information into the test case specifications.
* Consider adding the default level (debug, info etc) to the messages as the authoritative source.

https://github.com/dotse/zonemaster/blob/master/docs/specifications/tests/TestMessages.md

mattias-p:
This PR updates both the test messages table itself and the utility that generates it. Both the table and the utility are updated in steps to make the changes easier to follow. Refer to commit messages for details.

Fixes #569.


utils/maptestmessages.pl
```diff
@@ -1,28 +1,36 @@
...
-    while ( my ($key, $list) = each %$ref ) {
+    for my $key (sort keys %$ref) {
+        my $list = $ref->{$key};
         for my $tag (map { uc( $module ) . ':' . $_ } sort { $a cmp $b } @$list) {
-	    # printf "%46s -> %s::%s\n", $tag, $module, $key;
-	    my $testmodule = sprintf("%s::%s", $module, $key);
-	    my $testlevel = "$module-TP";
-	    my $testcase = "$key.md";
-	    my $testlink = "$testlevel/$testcase";
-	    printf "| %-30s | [%s](%s) |\n", $tag, $testmodule, $testlink;
```
pawal: Why would you want to remove the link to the test case specification?

mattias-p: The link target is just relocated to the page footer. That saves some horizontal space in text mode.

pawal: But is text mode realistically the proper use case for this file? The whole point is that you can search for the log message, and find the exact specification of why you see the message. This adds another indirection of doing this.

mattias-p: I didn't think of that use case. I figured people would enter the log message into Google and end up on Github: https://google.com/search?q=zonemaster+BASIC%3ADOMAIN_NAME_TOO_LONG

sandoche2k:
@mattias-p merge?

mattias-p:
Reverted to using direct links to cater to @pawal's use case. It's still possible to avoid line wrapping with `less -S`.

PR #577, opened 2018-03-05
added reference documents for zonecheck and updated the links

sandoche2k:


sandoche2k:
IMHO It is better to have a reference

PR #583, opened 2018-03-08
Fix level headings

mattias-p:
Make the heading 1 of each level.md reflect the name of the test level. This improves navigability.


docs/specifications/tests/Nameserver-TP/level.md
```diff
@@ -1,4 +1,4 @@
-# List of Name Server tests
+# List of Nameserver tests
 
 These are tests of the properties of a name server.
```
matsduf: Should we write "name server" or "nameserver"? Both are used in DNS RFCs, "name server" is more "standard English", but "nameserver" makes more a special term. 

I think we should be consistent, and note that the first line after the headline has "name server". Both should be the same.

mattias-p: Another item for the style guide I suppose. Dropping that change for now.

Issue #518, opened 2017-10-24
Broken links


mattias-p:
There are some broken links that I'm not sure what to do about.

docs/requirements/TestRequirements.md:
 * Links to: http://www.zonecheck.fr/features.shtml - Is there a backup of this document?
 * Links to: http://www.afnic.fr/en/products-and-services/services/zonecheck/tests-policy-2.html - Is there a backup of this document?
 * Links to: <https://github.com/dotse/zonemaster/blob/develop/docs/specifications/tests/Delegation-TP/??> - R19, Does anybody know?
 * Links to: https://github.com/dotse/zonemaster/blob/develop/docs/specifications/tests/DNSSEC-TP/dnssec12.md - R84, Is there supposed to be a test case DNSSEC12?

docs/specifications/tests/TestMessages.md:
 * Links to: https://github.com/dotse/zonemaster/blob/develop/docs/specifications/tests/Meta-TP/meta01.md - META:443_HAS_TLSA and META:443_NO_TLSA, Is there supposed to be a Meta test plan?

docs/specifications/tests/MasterTestPlan.md:
 * Links to: https://github.com/dotse/zonemaster/blob/develop/docs/specifications/tests/Resolver-TP/level.md - Is there supposed to be a Resolver test plan?

WISHLIST.md:
 * Links to: http://www.dnssec-tools.org/testzone/ - Could not resolve host: www.dnssec-tools.org

sandoche2k:
@mattias-p For ZC, i have added the documents. Details here : https://github.com/dotse/zonemaster/pull/577

sandoche2k:
> Links to: https://github.com/dotse/zonemaster/blob/develop/docs/specifications/tests/Delegation-TP/?? > - R19, Does anybody know?

I think it should be linked to 
https://github.com/dotse/zonemaster/blob/master/docs/specifications/tests/Delegation-TP/delegation01.md

sandoche2k:
> Links to: https://github.com/dotse/zonemaster/blob/develop/docs/specifications/tests/DNSSEC-
> TP/dnssec12.md - R84, Is there supposed to be a test case DNSSEC12?

There should be a new test case as indicated in the future test section

sandoche2k:
> Links to: https://github.com/dotse/zonemaster/blob/develop/docs/specifications/tests/Meta-
> TP/meta01.md  - META:443_HAS_TLSA and META:443_NO_TLSA, Is there supposed to be a Meta > test plan?

[...]

PR #584, opened 2018-03-08
More links

mattias-p:
More link fixing. Using standard autolinks simplifies automatic link checking.

### DIFF
--- a/docs/internal-documentation/maintenance/SupportCriteria.md
+++ b/docs/internal-documentation/maintenance/SupportCriteria.md
@@ -48,28 +48,28 @@ Operating system specific guidelines:
 
 * CentOS:
   * Base Distributions are listed here:
-    https://en.wikipedia.org/wiki/CentOS#End-of-support_schedule
+    <https://en.wikipedia.org/wiki/CentOS#End-of-support_schedule>
   * The current minor releases are listed at:
-    https://wiki.centos.org/Download
+    <https://wiki.centos.org/Download>
 
 * Debian:
   * Current versions of "stable" and "oldstable" are listed here:
-    https://wiki.debian.org/DebianReleases#Current_Releases.2FRepositories
+    <https://wiki.debian.org/DebianReleases#Current_Releases.2FRepositories>
   * The current point releases are listed here:
-    https://wiki.debian.org/DebianReleases/PointReleases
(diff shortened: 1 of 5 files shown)


mattias-p:
The purpose is to increase the range of automatic link checking to include autolinks. Standard Commonmark parser libraries are able to extract standard autolinks but not [GFM autolinks](https://github.github.com/gfm/#autolinks-extension-). Those libraries are used by link checking tools.

I occasionally use tools to look for broken links. That's how I found the broken links in #518. Unfortunately the tool I used didn't know about the GFM autolink extension, so I there were lots of links that didn't get checked that time around. Among them the two links to <http://www.zonecheck.fr/features.shtml> that are fixed in this PR.

To mitigate the problem of overlooked links for the future I searched for GFM autolinks in zonemaster and added angle brackets to all the links I could find.

I agree that an item about this should be added to the style guide.

matsduf:
I want to see a more complete proposal of how links should be written Zonemaster documentation before this PR is merged. When I test, putting "<...>" around a link does not add anything. It does not change the parsing.

mattias-p:
This is not about presentation on Github. Github renders Commonmark autolinks and GFM autolinks identically.
This is about using tools to go over all the links in an entire repository to verify that they all still work properly. Since GFM autolinks is an extension to Commonmark, they are unlikely to be recognized by Commonmark parsing libraries, and this is why I missed lots of links when I believed I'd checked them all.

matsduf:
According to <http://spec.commonmark.org/0.28/#autolinks> the following three should result in autolinks but only the first does in Github:

1. <http://foo.bar.baz>
2. <a+b+c:d>
3. <localhost:5001/foo>

Am I doing something wrong?

mattias-p:
Looks like you've discovered two Github bugs.

They're autolinks according to GFM as well:
* <https://github.github.com/gfm/#example-582>
* <https://github.github.com/gfm/#example-585>

And for reference: <https://github.github.com/gfm/#what-is-github-flavored-markdown->

sandoche2k:
Except for the confusion of the links <..>, i am ok with the content

sandoche2k:
@mattias-p then pls. merge

sandoche2k:
@mattias-p please close

mattias-p:
@sandoche2k This is already merged.

PR #589, opened 2018-03-10
Renamed level.md to README.md

matsduf:
Renamed level.md to README.md to make it easier to step through since Github automatically show the README.md file in a directory. Plus updates as a consequence of the renaming.


utils/maprequirement.pl
```diff
@@ -228,7 +229,7 @@ sub outputResult {
 				$otcDesc = $testInfo->{'desc'};
 				$oLevel  = $testInfo->{'level'};
 				my $linkID = "[$otcID]($otcLink)";
-				printf "|%3s|%23s|[%10s](%s/level.md)|%-10s|%17s|\n",
+				printf "|%3s|%23s|[%10s](%s/$levelfile)|%-10s|%17s|\n",
```
mattias-p: Please either inline level.md or using another %s. Interpolating stuff into format strings adds unnecessary complexity.

docs/specifications/tests/README.md
```diff
@@ -17,10 +17,10 @@ project. All the details are in the [Master Test Plan](MasterTestPlan.md).
 
 1. Each test level has been separated into a separate directory below
    this directory.
-2. Under each test level directory there is a level.md document
+2. Under each test level directory there is a level document (README.md)
    describing the test level. See for example the level document from
-   [Syntax-TP](Syntax-TP/level.md).
-3. The level.md document describes the mapping from Test Requirement
+   [Syntax-TP](Syntax-TP/README.md).
+3. The level document (README.md) describes the mapping from Test Requirement
```
mattias-p: Pointing out the filename without the directory is of little help to people who don't already know where to look. Also, it was just defined in the previous item so people can be expected to know it without further explanation.

docs/specifications/tests/README.md
```diff
@@ -33,7 +33,7 @@ project. All the details are in the [Master Test Plan](MasterTestPlan.md).
 ## Document Hierarchy
 
 1. Each Test Level described in [Master Test Plan](MasterTestPlan.md)
-   should be linked directly to the correct level.md document.
+   should be linked directly to the correct level document (README.md).
```
mattias-p: I don't think it makes sense to point out the filename without the directory here either. Maybe the definition of the term could be highlighted or something?

docs/specifications/tests/README.md
matsduf: Here no specific directory can be given. I'll try an update.

pawal:
If you change to README.md, don't forget to change the relevant code in https://github.com/dotse/zonemaster/blob/master/utils/maprequirement.pl

docs/specifications/tests/README.md
matsduf: I'll try, but please suggest better wording if mine is not good enough. Please note that this PR is **ONLY** about changing the file names. Other updates better fit in a new PR.

matsduf:
@pawal, if you look at the updated files you will find that `maprequirement.pl` is included.

PR #595, opened 2018-03-13
Updated MasterTestPlan.md and added subordinate files

matsduf:
* Copied Algorithms-used-by-dnscheck.md from DNSCheck Wiki to Zonemaster repository
* Copied Tests.md from DNSCheck Wiki to Zonemaster repository (new name is DNSCheck-Tests.md)
* Corrected link in MasterTestPlan.md
* Updated links in MasterTestPlan.md from pointing at files in DNSCheck Wiki to the local files 
* Minior editorial changes of MasterTestPlan.md
* Corrected markdown encoding of headlines
* Updated links to right format
* Corrected second level i lists


matsduf:
@sandoche2k, can you review this PR?

mattias-p:
@sandoche2k I think there is a good argument why the documents should not be internal. The master test plan links to the documents, and the master test plan is not internal, so the documents shouldn't be either.

sandoche2k:
Maybe in an another directory. But not in the requirements directory. We
could provide a link as it is now from the Master Test Plan.

mattias-p:
They're mentioned in the master test plan under system requirements and as external references. What do you think of having them under docs/requirements but clarifying their status in the README?

matsduf:
@sandoche2k, the reason that I put the documents in "docs/requirement" instead of "docs/internal-documentation/requirements" are two fold:
* The documents are linked from external documents.
* The primary linking is from documents in the "requirements" directory.

I will make a new suggestion.

matsduf:
@sandoche2k, please review my updates.

docs/requirements/TestRequirements.md
```diff
@@ -130,13 +130,13 @@ Requirements on writing test specifications
 These are some requirements for writing specifications for "Zonemaster":
 
  1. Follow the framework of the IEEE 829-2008.
- 2. The documents must be in [Markdown Syntax]
-    (http://daringfireball.net/projects/markdown/syntax).
+ 2. The documents must be in 
+    [Markdown Syntax](http://daringfireball.net/projects/markdown/syntax).
```
sandoche2k: Why  [Markdown Syntax] in newline ?

matsduf: There must not be any white space or line break between "]" and "(". There are three options that give the same result on Github:

```
 2. The documents must be in 
     [Markdown Syntax](http://daringfireball.net/projects/markdown/syntax).

 2. The documents must be in [Markdown Syntax
     ](http://daringfireball.net/projects/markdown/syntax)

 2. The documents must be in [Markdown Syntax](
     http://daringfireball.net/projects/markdown/syntax)
```

mattias-p: There's also:

```
The documents must be in [Markdown Syntax].

[Markdown Syntax]: http://daringfireball.net/projects/markdown/syntax
```

https://spec.commonmark.org/0.28/#reference-link

Issue #513, opened 2017-10-02
Markdown rendering


mattias-p:
[Github has updated its markdown rendering engine](https://github.com/blog/2333-a-formal-spec-for-github-flavored-markdown), breaking some of documents. We need to update our markup to make it compliant with the new spec.

Specifically I've noted problems with headings that don't render properly anymore. [The new spec](https://github.github.com/gfm/#atx-headings) doesn't recognize NBSP as a separator between the '#' prefix and the heading text. We need to replace our NBSP characters with SPACE characters.

mattias-p:
Merged into develop.

mattias-p:
I made a review and found a bunch of other rendering problems, presumably (but not necessarily) introduced by the update to the markdown renderer.

Broken general layout:
* docs/internal-documentation/distrib-testing/Debian-Testing-Instructions.md

Broken tables:
* docs/requirements/TestRequirements.md#tests-to-implement
* docs/internal-documentation/functional-tests/Configuration-TP/configuration04.md#appendix

Broken image links:
* docs/design/README.md
* docs/internal-documentation/requirements/README.md
* docs/requirements/README.md

sandoche2k:
I think we can remove this directory docs/internal-documentation/distrib-testing. As, we are using the general installation instructions for preparing the release, this can be removed. I will make a PR, if you agree.

matsduf:
The installation document does not cover the general set-up of a server before installation, e.g. added locals. I think the documents could be useful, but they need updates.

sandoche2k:
Agreed. But "Debian-Testing-Instructions.md" should be removed.

matsduf:
@sandoche2k, do you want to keep the directory with "Debian-Preparation.md" and "FreeBSD-Preparation.md" or do you want to move it somewhere else?

I think it is fine to remove "Debian-Testing-Instructions.md".

mattias-p:
@matsduf This PR is about getting rid of broken markup. If we also want to rearrange the repository I suggest a separate issue is created for that.

PR #603, opened 2018-03-28
Fix issue #513

sandoche2k:


matsduf:
@sandoche2k, can you please give a short description of this PR. It seems to cover more than issue #513. Does it resolve everything in #513?

mattias-p:
AFAICT it solves all but one remaining item from #513.
[docs/internal-documentation/requirements/README.md](https://github.com/dotse/zonemaster/blob/develop/docs/internal-documentation/requirements/README.md) still has a broken image link.

sandoche2k:
The link for the image seems to be the same as others. But, it is broken ?

mattias-p:
Yes it is broken. Does the image show up for you? It shouldn't because the `requirements` directory doesn't have a `docs` subdirectory. The best fix would probably be to replace the current heading with plain text one simply saying "Internal Documentation".

matsduf:
@sandoche2k, it does not look like this PR solves everything what is stated in #513, and it does other things (removing file instead of fixing markdown), but I approve what it does.

I think all PRs should have a short description of what it does, and if it refers to an issue, there should be a link to that issue in the description.

PR #609, opened 2018-04-11
Update delegation01

matsduf:
Updated DELEGATION01:
* Added test for number of NS with IPv4.
* Added test for number of NS with IPv6.
* Added the messages to the Test Case definition.
* Added default levels of messages.
* Messages ENOUGH_NS_GLUE, NOT_ENOUGH_NS_GLUE, ENOUGH_NS and NOT_ENOUGH_NS renamed to ENOUGH_NS_PARENT, NOT_ENOUGH_NS_PARENT, ENOUGH_NS_CHILD and NOT_ENOUGH_NS_CHILD, respectively.
* Added logic to count the numer of name servers with IPv4 resolution and IPv6 resolution, respectively.
* The following messages are new: NO_IPV4_NS_PARENT, NOT_ENOUGH_IPV4_NS_PARENT, ENOUGH_IPV4_NS_PARENT, 
NO_IPV6_NS_PARENT, NOT_ENOUGH_IPV6_NS_PARENT, ENOUGH_IPV6_NS_PARENT, 
NO_IPV4_NS_CHILD, NOT_ENOUGH_IPV4_NS_CHILD, ENOUGH_IPV4_NS_CHILD, 
NO_IPV6_NS_CHILD, NOT_ENOUGH_IPV6_NS_CHILD and ENOUGH_IPV6_NS_CHILD.

### DIFF
--- a/docs/specifications/tests/Delegation-TP/delegation01.md
+++ b/docs/specifications/tests/Delegation-TP/delegation01.md
@@ -20,45 +20,33 @@ The domain name to be tested ("child zone").
 
 ## Ordered description of steps to be taken to execute the test case
 
-1. Obtain the complete set of the names of the name servers for the *child 
-   zone* (distinct NS records) from parent using [Method2].
-2. If the number of name servers from parent is at least two, emit the
-   *ENOUGH_NS_PARENT*, else emit *NOT_ENOUGH_NS_PARENT*.
-3. Obtain the IP addresses for the set of the name servers from parent 
-   using [Method4].
-4. If there are no name servers from parent resolving to an IPv4 
-   address then emit *NO_IPV4_NS_PARENT*.
-5. If the number of name servers from parent resolving to an IPv4 
-   address is one then emit *NOT_ENOUGH_IPV4_NS_PARENT*.
-6. If the number of name servers from parent resolving to an IPv4 
-   address is at least two then emit *ENOUGH_IPV4_NS_PARENT*.
-7. If there are no name servers from parent resolving to an IPv6 
-   address then emit *NO_IPV6_NS_PARENT*.
-8. If the number of name servers from parent resolving to an IPv6 
-   address is one then emit *NOT_ENOUGH_IPV6_NS_PARENT*.
-9. If the number of name servers from parent resolving to an IPv6 
-   address is at least two then emit *ENOUGH_IPV6_NS_PARENT*.
-10. Obtain the complete set of the names of the name servers for the 
-    *child zone* from the *child zone* using [Method3].
-11. If the number of name servers from child is at least two emit the
-   *ENOUGH_NS_CHILD*, else emit *NOT_ENOUGH_NS_CHILD*.
-12. Obtain the IP addresses for the set of the name servers from 
-    *child zone* using [Method5] and, for 
-    out-of-bailiwick NS, using [Method4].
-13. If there are no name servers from child resolving to an IPv4 
-   address then emit *NO_IPV4_NS_CHILD*.
-14. If the number of name servers from child resolving to an IPv4 
-   address is one then emit *NOT_ENOUGH_IPV4_NS_CHILD*.
-15. If the number of name servers from child resolving to an IPv4 
-   address is two then emit *ENOUGH_IPV4_NS_CHILD*.
-16. If there are no name servers from child resolving to an IPv6 
-   address then emit *NO_IPV6_NS_CHILD*.
-17. If the number of name servers from child resolving to an IPv6 
-   address is one then emit *NOT_ENOUGH_IPV6_NS_CHILD*.
-18. If the number of name servers from child resolving to an IPv6 
-   address is two then emit *ENOUGH_IPV6_NS_CHILD*.
+ 1. Using [Method2], obtain the set of "in-delegation name server names" for the *child zone* (distinct NS records).
+ 2. Count the *in-delegation name server names*:
+    1. If zero or one, emit *NOT_ENOUGH_NS_PARENT*.
+    2. If two or more, emit the *ENOUGH_NS_PARENT*.
+ 3. Using [Method4], obtain the set of "in-delegation glue addresses" for the *child zone*.
+ 4. Count the IPv4 addresses among the *in-delegation glue addresses*:
+    1. If zero, emit *NO_IPV4_NS_PARENT*.
+    2. If one, emit *NOT_ENOUGH_IPV4_NS_PARENT*.
+    3. If two or more, emit *ENOUGH_IPV4_NS_PARENT*.
+ 5. Count the IPv6 addresses among the *in-delegation glue addresses*:
+    1. If zero, emit *NO_IPV6_NS_PARENT*.
+    2. If one, emit *NOT_ENOUGH_IPV6_NS_PARENT*.
+    3. If two or more, emit *ENOUGH_IPV6_NS_PARENT*.
+ 6. Using [Method3], obtain the set of "in-zone name server names".
+ 7. Count the *in-delegation name server addresses*:
+    1. If zero or one, emit *NOT_ENOUGH_NS_CHILD*.
+    2. If two or more, emit *ENOUGH_NS_CHILD*.
+ 8. Using [Method4] and [Method5], obtain the set of "in-zone name server addresses".
+ 9. Count the IPv4 addresses among the *in-zone name server addresses*:
+    1. If zero, emit *NO_IPV4_NS_CHILD*.
+    2. If one, emit *NOT_ENOUGH_IPV4_NS_CHILD*.
+    3. If two or more, emit *ENOUGH_IPV4_NS_CHILD*.
+10. Count the IPv6 addresses among the *in-zone name server addresses*:
+    1. If zero, emit *NO_IPV6_NS_CHILD*.
+    2. If one, emit *NOT_ENOUGH_IPV6_NS_CHILD*.
+    3. If two or more, emit *ENOUGH_IPV6_NS_CHILD*.
 
- 
 ## Outcome(s)
 
 The outcome of this Test Case is "fail" if there is at least one message
--- a/docs/specifications/tests/Delegation-TP/delegation01.md
+++ b/docs/specifications/tests/Delegation-TP/delegation01.md
@@ -77,22 +77,22 @@ Else the outcome of this Test Case is "pass".
 
 Message                       | Default severity level (if message is emitted)
 :-----------------------------|:-----------------------------------
-NOT_ENOUGH_NS_DEL             | ERROR
-ENOUGH_NS_DEL                 | INFO
-NO_IPV4_NS_DEL                | ERROR
-NOT_ENOUGH_IPV4_NS_DEL        | ERROR
-ENOUGH_IPV4_NS_DEL            | INFO
-NO_IPV6_NS_DEL                | NOTICE
-NOT_ENOUGH_IPV6_NS_DEL        | NOTICE
-ENOUGH_IPV6_NS_DEL            | INFO
-NOT_ENOUGH_NS_CHILD           | ERROR
-ENOUGH_NS_CHILD               | INFO
-NO_IPV4_NS_CHILD              | ERROR
-NOT_ENOUGH_IPV4_NS_CHILD      | ERROR
 ENOUGH_IPV4_NS_CHILD          | INFO
-NO_IPV6_NS_CHILD              | NOTICE
-NOT_ENOUGH_IPV6_NS_CHILD      | NOTICE
+ENOUGH_IPV4_NS_DEL            | INFO
 ENOUGH_IPV6_NS_CHILD          | INFO
+ENOUGH_IPV6_NS_DEL            | INFO
+ENOUGH_NS_CHILD               | INFO
+ENOUGH_NS_DEL                 | INFO
+NO_IPV4_NS_CHILD              | ERROR
+NO_IPV6_NS_CHILD              | NOTICE
+NO_IPV4_NS_DEL                | ERROR
+NO_IPV6_NS_DEL                | NOTICE
+NOT_ENOUGH_IPV4_NS_CHILD      | ERROR
+NOT_ENOUGH_IPV6_NS_CHILD      | ERROR
+NOT_ENOUGH_IPV4_NS_DEL        | ERROR
+NOT_ENOUGH_IPV6_NS_DEL        | ERROR
+NOT_ENOUGH_NS_CHILD           | ERROR
+NOT_ENOUGH_NS_DEL             | ERROR
 
 
 ## Special procedural requirements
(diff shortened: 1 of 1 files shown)


docs/specifications/tests/Delegation-TP/delegation01.md
```diff
@@ -1,48 +1,98 @@
...
-If the total amount of name servers used for the domain are two or more,
-then the test case succeeds.
+The domain name to be tested.
 
-### Special procedural requirements
+## Ordered description of steps to be taken to execute the test case
+
+1. Obtain the complete set of name servers (distinct NS records) from 
+   parent using [Method2](../Methods.md).
+2. If the number of name servers from parent is at least two emit the
```
mattias-p: Would you consider adding a comma after the condition? That way the reader gets a visual hint about the structure of the sentence.

In some places there are comma signs before "else" too. I find those helpful as well.

docs/specifications/tests/Delegation-TP/delegation01.md
```diff
@@ -1,48 +1,98 @@
...
+   message *ENOUGH_NS_PARENT* else emit  
+   *NOT_ENOUGH_NS_PARENT*.
+3. Obtain the IP addresses for the set of the name servers from parent 
+   using [Method4](../Methods.md).
+4. If the number of name servers from parent resolving to an IPv4
```
mattias-p: These "if, else if, else" sentences are rather long and complex, and not so easy to read. If we end up having lots of those in the test cases, maybe we should find a more readable way to express them.

docs/specifications/tests/Delegation-TP/delegation01.md
```diff
@@ -1,48 +1,98 @@
...
+
+1. Obtain the complete set of name servers (distinct NS records) from 
+   parent using [Method2](../Methods.md).
+2. If the number of name servers from parent is at least two emit the
+   message *ENOUGH_NS_PARENT* else emit  
+   *NOT_ENOUGH_NS_PARENT*.
+3. Obtain the IP addresses for the set of the name servers from parent 
+   using [Method4](../Methods.md).
+4. If the number of name servers from parent resolving to an IPv4 
+   address is none then emit  *NO_IPV4_NS_PARENT*, else
```
mattias-p: "The number [...] is none" really pokes my eye. It ought to be "zero".

mattias-p:
I forgot to mention. The Steps never refer to the Inputs. Did you leave that out on purpose?

docs/specifications/tests/Delegation-TP/delegation01.md
```diff
@@ -1,48 +1,98 @@
...
 
-If the total amount of name servers used for the domain are two or more,
-then the test case succeeds.
+The domain name to be tested.
 
-### Special procedural requirements
+## Ordered description of steps to be taken to execute the test case
+
+1. Obtain the complete set of name servers (distinct NS records) from 
+   parent using [Method2](../Methods.md).
```
mattias-p: "Obtain the domain names of the name servers from parent using Method2."?

sandoche2k:
@matsduf i almost agree with the test case. But i have an issue in approving. The reasons are 
1. We need to get the methods finalised
2. Recent test cases, that you have written indicates what the messages should be. If we follow that, then we need to revamp all test specs.

docs/specifications/tests/Delegation-TP/delegation01.md
matsduf: I think referring to the names of the name servers as "domain names" would be confusing. Would the following be better?

```
Obtain the complete set names of the name servers for the child zone (...)
 ```

Comment (mattias-p) [reply docs/specifications/tests/Delegation-TP/delegation01.md]: Sure. Don't forget the "of" in "complete set of names".

File: docs/specifications/tests/Delegation-TP/delegation01.md [hunk 5]
```diff
@@ -1,48 +1,98 @@
...
+   if it is one emit *NOT_ENOUGH_IPV6_NS_CHILD* else emit
+   *ENOUGH_IPV6_NS_CHILD*.
+
+ 
+## Outcome(s)
+
+The outcome of the test case depends on the highest level of the messages 
+generated.
+
+Message                       | Default level (if message is emitted)
```
mattias-p: How about "Default severity"?

mattias-p: Oh, and it's always the default level, right? Whether or not the message is emitted.

docs/specifications/tests/Delegation-TP/delegation01.md
```diff
@@ -1,48 +1,98 @@
...
+9. If the number of name servers from child resolving to an IPv6 
+   address is none then emit  *NO_IPV6_NS_CHILD*, else 
+   if it is one emit *NOT_ENOUGH_IPV6_NS_CHILD* else emit
+   *ENOUGH_IPV6_NS_CHILD*.
+
+ 
+## Outcome(s)
+
+The outcome of the test case depends on the highest level of the messages 
+generated.
```
mattias-p: The messages aren't generated. They're emitted.

matsduf:
@sandoche2k,

1. Yes, we need to have the methods finalized, but this update does not depend on the finalized methods. I think it would be hard to try to rewrite the all the test cases in one PR. That would be too much to review. I think we have to accept that we would have to come back to some Test Cases again re-adjust.

2. Yes, the messages have been added to make the test cases specification the authoritative source of the entire test case, of which I consider the messages to be part of. Yes, we would have to update all Test Case specifications, which I think we should anyway.

docs/specifications/tests/Delegation-TP/delegation01.md
```diff
@@ -1,48 +1,119 @@
...
+## Outcome(s)
+
+The outcome of this Test Case is "fail" if there is at least one message
+with the severity level *ERROR* or *CRITICAL*.
+
+The outcome of this Test Case is "warning" if there is at least one message
+with the severity level *WARNING*, but no message with severity level
+*ERROR* or *CRITICAL*.
+
+In other cases the outcome of this Test Case is "pass".
```
mattias-p: Would it be possible to avoid causal use of the word "case" here?

mattias-p:
Did you see that?! I added a commit into your repo! I didn't intend to. I intended to create a new branch in my own repo. Shame on me for using the Github web editor...

Well, now that it's there, can you look at it and say what you think? It felt that the Steps were getting awfully long and needed some structure.

matsduf:
@mattias-p, can you please review again. I have handled your comments and your commit.

sandoche2k:
@matsduf merg?

matsduf:
@sandoche2k, I raised the question with @mattias-p if we really should merge this into the develop branch now or maybe after release or into a feature branch.

sandoche2k:
@matsduf @mattias-p Even though the implementation is not done, we could merge into develop branch and to the release. The reason, i feel is that currently there is already confusion with different branches.

mattias-p:
I don't think we should make releases that introduce new features in the documentation but doesn't implement them. That doesn't look good and it's very confusing when you want to go back and see when a feature was first introduced.

sandoche2k:
@mattias-p I have two arguments why the specifications updated should be merged into develop:

1. We mentioned that only for profile, we will have a separate branch. Even with this single branch that is other than develop we are having confusions. This is a real situation as of now.
2. Currently i do not have a timeline, when the all the specs will be updated and implemented. The specs updated in a separate branch will be forgotten after a while, depending on people coming in and going out of the project.

My two cents is that update the specs as soon as it is approved into develop branch. During the release merge into the master branch and create an issue in the engine mentioning that the specs updated should be implemented.

mattias-p:
@sandoche2k I do appreciate the confusion. Since the team is new to feature branches, it's a good idea to only have a few - possibly as few as one. As we gain experience with this way of working, I'm convinced we'll appreciate it more and more until it feels completely natural. That said I concede that we aren't there yet.

You mention the creation of an issue not to forget the implementation of the merged documentation. For a half-done feature branch (that doesn't already have an associated ticket) you'd do the same thing - create an issue with a link to the feature branch. The two problems are actually quite analogous. The differ only in where the unfinished work is located.

It's true that we don't have a timeline for these changes, but in my view that's an argument for not having unimplemented documentation into the develop branch. When managing multiple feature branches, any confusion is kept within the project team. If we make a release with unimplemented documentation, the confusion spreads to our users.

In the end there are pros and cons.

matsduf:
@JohanIIS, added your suggested updates. Instead of removing the terms "in-bailiwick" and "out-of-bailiwick" I added a reference. I have done so in other specifications of test cases. I feel it is better to use the same definition as elsewhere.

docs/specifications/tests/Delegation-TP/delegation01.md
```diff
@@ -1,49 +1,130 @@
...
-### Special procedural requirements
+ 1. Using [Method2], obtain the complete set of names of the name servers 
+    from parent for the *Child Zone* (distinct NS records).
+
+ 2. Count the name server names:
+    1. If zero or one, emit *NOT_ENOUGH_NS_PARENT*.
+    2. If two or more, emit *ENOUGH_NS_PARENT*.
+
+ 3. Obtain the IP addresses for the set of name servers from the parent for the 
+    *Child Zone*.
```
mattias-p: How's this to be done? Is Method3 the best way?

docs/specifications/tests/Delegation-TP/delegation01.md
```diff
@@ -1,49 +1,130 @@
...
+
+ 4. Count the number of name server names that resolve into at least one IPv4 
+    address:
+    1. If zero, emit *NO_IPV4_NS_PARENT*.
```
mattias-p: Do we really need separate messages for the zero case? Can't we just reuse the NOT_ENOUGH messages? That seems sufficient to me.

matsduf: Is it because the number of messages will be high? I feel that it is easier to write good messages in plain language when separating the two. Secondly, one could decide that it is OK not to have name servers on IPv6, but if you have there should be at least two.

docs/specifications/tests/Delegation-TP/delegation01.md
matsduf: Will update, but please note that all updated test cases probably have the same text.

docs/specifications/tests/Delegation-TP/delegation01.md
matsduf: Yes, will update.

docs/specifications/tests/Delegation-TP/delegation01.md
```diff
@@ -1,49 +1,130 @@
...
+
+ 6. Using [Method3], obtain the complete set of names of the name servers
+    from the *Child Zone* for the *Child Zone*. 
+
+ 7. Count the name server names:
+    1. If zero or one, emit *NOT_ENOUGH_NS_CHILD*.
+    2. If two or more, emit *ENOUGH_NS_CHILD*.
+
+ 8. Using [Method5], obtain the IP addresses for the [in-bailiwick] name servers 
+    from the *Child Zone* for the *Child Zone*.
```
mattias-p: I understand Method5 to comprise all of the name servers as reported by the child zone itself. Here Method5 is used to obtain only the in-bailiwick name servers. Is there an additional filtering step hiding here?

docs/specifications/tests/Delegation-TP/delegation01.md
```diff
@@ -1,49 +1,130 @@
...
+
+ 9. Using [Method4], obtain the IP addresses for the [out-of-bailiwick] name 
+    servers for the *Child Zone*.
```
mattias-p: I understand don't understand Method4 to concern itself with what name servers the child zone itself reports.

docs/specifications/tests/Delegation-TP/delegation01.md
matsduf: The test case checks the the name servers both in the delegation (parent) and in the NS RR set in the zone (child). Method5 can only fetch in-bailiwick name server IP addresses which means that it could be stated that step 8 has redundant instructions. Maybe the following text would be better:


> 8. Using [Method5], obtain the IP addresses for the name servers from the *Child Zone* for the *Child Zone* (only the IP addresses for the [in-bailiwick] name servers will be available).

docs/specifications/tests/Delegation-TP/delegation01.md
matsduf: In step 8 we fetched the IP addresses for the in-bailiwick name servers, but we also need the the IP addresses for the out-of-bailiwick name servers, which method5 cannot provide, but method4 can (at least updated method4).

docs/specifications/tests/Delegation-TP/delegation01.md
mattias-p: I think that's a better wording.

docs/specifications/tests/Delegation-TP/delegation01.md
```diff
@@ -1,49 +1,118 @@
...
+
+ 6. Using [Method3], obtain the complete set of names of the name servers
+    from the *Child Zone* for the *Child Zone*. 
+
+ 7. Count the name server names:
+    1. If zero or one, emit *NOT_ENOUGH_NS_CHILD*.
+    2. If two or more, emit *ENOUGH_NS_CHILD*.
+
+ 8. Using [Method5], obtain the IP addresses for the name servers from 
+    the *Child Zone* for the *Child Zone*.
```
mattias-p: I interpret this step to obtains a set of IP-addresses. That's insufficient information to make the determination in the next step. Maybe we need a new method that returns a set of (name, address) pairs?

matsduf: Maybe we should have just two methods, one method that returns the name server names and its IP addresses from the delegation point of view and one method that returns the name server names and its IP addresses from the child zone point of view. The test cases could choose to call the methods in one of three ways:
* Set of IP addresses.
* Set of name servers (names).
* Set of name servers with IP adresses.

mattias-p: I agree that there's a symmetry here, but I don't think the right way to express that symmetry is to cram three basic methods into one super-method and having two of those super-methods. The simple solution is just having six methods and expressing the symmetry in their naming.

The structure of the return types is also interesting. We seem to agree that the first two methods return sets of IP addresses and sets of names, respectively. For the third kind you seem to think of the return type as a set of (name, set of IP addresses)-pairs where all names are unique. A mathematically equivalent (and much simpler) type is a set of (name, IP address)-pairs without the uniqueness requirement. This also happens to match Engine's internal representation. A `Zonemaster::Engine::Nameserver` represents a (name, IP address)-pair together with a bunch of additional properties. I suggest that we use the simpler type in the specifications as well.

Lastly, we need names for things. The two points of view need to be named, as well as the three basic methods. We also need some stylistic convention for the names - like use of abbreviations, casing and word separators. I don't really have a suggestion for the styling, but here's my two cents for the naming and return types:
* Delegation name server addresses: `Set[IP address]`
* Delegation name server names: `Set[name]`
* Delegatoin name servers: `Set[(name, IP address)]`
* In-zone name server addresses: `Set[IP address]`
* In-zone name server names: `Set[name]`
* In-zone name servers: `Set[(name, IP address)]`

PR #610, opened 2018-04-11
Update delegation02

matsduf:
Updated delegation02:
* Updated to new format.
* Updated what IP addresses that are fetched.
* Added explicit messages to the test case specification.
* Added default level of the messages.
* Specification mixed parent and child, which are kept separated.
* Message  SAME_IP_ADDRESS is split into two messages, DEL_NS_SAME_IP  and CHILD_NS_SAME_IP, respectively.
* Message DISTINCT_IP_ADDRESS is split into two messages, DEL_DISTINCT_NS_IP and CHILD_DISTINCT_NS_IP, respectively.

### DIFF
--- a/docs/specifications/tests/Delegation-TP/delegation02.md
+++ b/docs/specifications/tests/Delegation-TP/delegation02.md
@@ -24,7 +24,8 @@ The domain name to be tested.
    [Method2] and the IP addresses for each name using [Method4].
 
 2. If the same IP address is found for two or more name server names, 
-   emit *PARENT_NS_SAME_IP_ADDRESS* for each repeated address.
+   emit *DEL_NS_SAME_IP* for each repeated address, else emit
+   *DEL_DISTINCT_NS_IP*.
 
 3. Obtain the complete set of name server names from the child using 
    [Method3] and the IP addresses for each name using [Method5] for
@@ -32,7 +33,8 @@ The domain name to be tested.
    names only. 
 
 4. If the same IP address is found for two or more name server names, 
-   emit *CHILD_NS_SAME_IP_ADDRESS* for each repeated address.
+   emit *CHILD_NS_SAME_IP* for each repeated address, else emit
+   *CHILD_DISTINCT_NS_IP*.
 
 ## Outcome(s)
 
@@ -45,11 +47,12 @@ with the severity level *WARNING*, but no message with severity level
 
 In other cases the outcome of this Test Case is "pass".
 
-Message                       | Default severity level (if message is emitted)
-:-----------------------------|:-----------------------------------
-PARENT_NS_SAME_IP_ADDRESS     | ERROR
-CHILD_NS_SAME_IP_ADDRESS      | ERROR
-
+Message               | Default severity level (if message is emitted)
+:---------------------|:-----------------------------------
+DEL_NS_SAME_IP        | ERROR
+CHILD_NS_SAME_IP      | ERROR
+DEL_DISTINCT_NS_IP    | INFO
+CHILD_DISTINCT_NS_IP  | INFO
 
 ## Special procedural requirements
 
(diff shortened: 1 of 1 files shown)


docs/specifications/tests/Delegation-TP/delegation02.md
```diff
@@ -1,48 +1,76 @@
...
-1. Obtain the IP addresss of the name servers from the parent using [Method
-4](../Methods.md#method-4-obtain-glue-address-records-from-parent) and the child using
-   [Method 5](../Methods.md#method-5-obtain-the-name-server-address-records-from-child).
-2. If any of the IP addresses resolved in step 1 is not unique, then this
-   test case fails.
+1. Obtain the complete set of name servers (distinct NS records) from 
+   parent using [Method2].
+2. Obtain the IP addresses for the set of the name servers from parent 
+   using [Method4]. (This should include all IP address
+   for both in-bailiwick and out-of-bailiwick name servers.)
```
mattias-p: This is meant as an explanation of the contents of the return value of Method4, right? Or is it an instruction to do something more? If it's an instruction it should be made into its own step. If it's an explanation I it's better to just remove it (you can always click the link if you're interested in the details).

docs/specifications/tests/Delegation-TP/delegation02.md
```diff
@@ -1,48 +1,76 @@
...
+3. If the same IP address is found for two or more name servers then 
+   emit *PARENT_NS_SAME_IP_ADDRESS* and the IP address.
+4. Repeat step 3 for all IP addresses obtained in step 2.
```
mattias-p: AFAICT this step could simply be removed.

docs/specifications/tests/Delegation-TP/delegation02.md
```diff
@@ -1,48 +1,76 @@
...
+5. Obtain the complete set of name servers from the child using 
+   [Method3].
```
mattias-p: What are these names used for? Either some subsequent step should refer to this result, or this step should be removed.

docs/specifications/tests/Delegation-TP/delegation02.md
matsduf: This is a compensation for the unclearity of Method4 today.

docs/specifications/tests/Delegation-TP/delegation02.md
matsduf: Unless it is said you could stop with the first instance.

docs/specifications/tests/Delegation-TP/delegation02.md
matsduf: You need to have both name and address to see that  more than one name uses the same address. If you just get a set of IP addresses then you cannot see it.

matsduf:
@mattias-p, I have updated the text. Please review.

@sandoche2k, please review.

docs/specifications/tests/Delegation-TP/delegation02.md
```diff
@@ -1,49 +1,93 @@
...
+2. If the same IP address is found for two or more name server names, 
+   emit *[DEL_NS_SAME_IP]* for each repeated address, else emit
+   *[DEL_DISTINCT_NS_IP]*.
 
-If all the IP addresses used by the name servers for the domain are unique,
-then the test succeeds.
+3. Obtain the complete set of name server names from the child using 
+   [Method3] and the IP addresses for each name using [Method5] for
+   [in-bailiwick] names and using [Method4] for [out-of-bailiwick] 
+   names only.
```
mattias-p: I believe this needs to be clarified/corrected.

matsduf: Corrected.

PR #615, opened 2018-04-24
Update consistency01

matsduf:
Updated consistency01

* Updated to new format.
* Added explicit messages.
* Added explicit reference to serial number arithmetic.
* Added explicit reference to accepted difference.
* Removed copyright statement.
* Removed description/headline from identifier line.

Changes compared to current implementation (develop branch pre v2018.1):

**In practice no changes.**

The implementation has the accepted serial difference hardcoded to 0, which means that the code will only consider equal or not equal serial number. The requirement for the new message in the specification, *MULTIPLE_SOA_SERIALS_OK*, will never be met.

The way that serial values are compared, when not equal, is not correct in the implementation, but that is only relevant if the accepted serial difference is at least 1.

Also see #617.

### DIFF
--- a/docs/specifications/tests/Consistency-TP/consistency01.md
+++ b/docs/specifications/tests/Consistency-TP/consistency01.md
@@ -39,14 +39,14 @@ serial number consistency.
     server IP.
 
  3. If the name server does not respond with a DNS response, emit 
-    *[NO_RESPONSE]* and go to next name server.
+    *[NO_RESPONSE]* and go to next name server IP.
 
  4. If the name server responds with a DNS response but no a SOA record 
     is included, emit *[NO_RESPONSE_SOA_QUERY]* and got to next name 
-    server.
+    server IP.
 
  5. Retrieve the SOA SERIAL from the responses and go to next name server
-    until all name server IPs have been queried.
+    IP until all name server IPs have been queried.
 
  6. If at least one SOA SERIAL has been retrieved and all serial 
     numbers are identical, emit *[ONE_SOA_SERIAL]*.
--- a/docs/specifications/tests/Consistency-TP/consistency01.md
+++ b/docs/specifications/tests/Consistency-TP/consistency01.md
@@ -1,17 +1,17 @@
-## CONSISTENCY01: SOA serial number consistency
+# CONSISTENCY01: SOA serial number consistency
 
-### Test case identifier
+## Test case identifier
 
 **CONSISTENCY01:** SOA serial number consistency
 
-### Objective
+## Objective
 
 For the data served by the authoritative name servers for a designated zone
 to be consistent, all authoritative name servers must serve the same SOA
 record for the designated zone.   
 
-If the serial number (explained in 3.3.13. of [RFC 1035](https://tools.ietf.org/html/rfc1035)) 
-, which is part of the SOA record, is not consistent between authoritative servers, 
+If the serial number (explained in 3.3.13. of [RFC 1035]), 
+which is part of the SOA record, is not consistent between authoritative servers, 
 there is a possibility that the data served is inconsistent. The reasons for this 
 inconsistency may be different - such as misconfiguration, or as a result of slow 
 propagation to the secondary name servers.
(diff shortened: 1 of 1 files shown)


docs/specifications/tests/Consistency-TP/consistency01.md
```diff
@@ -1,52 +1,82 @@
...
-     3. If the difference is less than or equal to the minor value,
-        then the test is passed.
+ 1. Obtain the list of name server IPs for the *child zone* from [Method4] 
+    and [Method5].
+ 2. Create an SOA query for *child zone* apex and send it to all name 
+    server IPs.
+ 3. Retrieve the SOA RR from the responses from all name server IPs.
+ 4. If a name server does not respond, emit *NO_RESPONSE*.
+ 5. If a name server does not include a SOA record in the response,
+    emit *NO_RESPONSE_SOA_QUERY*.
```
mattias-p: In my view a better name for this would be NO_ANSWER_SOA_RR_SET, but maybe it's not worth changing the name.

docs/specifications/tests/Consistency-TP/consistency01.md
```diff
@@ -1,52 +1,82 @@
...
+
+In other cases the outcome of this Test Case is "pass".
+
+Message                       | Default severity level (if message is emitted)
+:-----------------------------|:-----------------------------------
+NO_RESPONSE                   | WARNING
+NO_RESPONSE_SOA_QUERY         | DEBUG
+ONE_SOA_SERIAL                | INFO
+MULTIPLE_SOA_SERIALS          | WARNING
+MULTIPLE_SOA_SERIALS_OK       | NOTICE
```
mattias-p: This needs an accompanying implementation.

matsduf: Yes, true.

docs/specifications/tests/Consistency-TP/consistency01.md
matsduf: It is not bad enough to be worth changing it. It make sense at least.

docs/specifications/tests/Consistency-TP/consistency01.md
mattias-p: The current implementation can neither emit MULTIPLE_SOA_SERIALS nor MULTIPLE_SOA_SERIALS_OK. So either they should be removed from the specification or the code should be updated to emit them when appropriate.

matsduf: The current specification will emit *MULTIPLE_SOA_SERIALS* when more then one serial value has been found.

*The current implementation is compatible with the updated specification.*

When the updated specification has been accepted we should look at issue #617. If the decision is to keep the algorithm we should create an issue that requests the update of the implementation in the following respects:

1. The accepted SOA serial difference must be configurable.
2. Sorting of serial values must follow the arithmetic of [RFC 1982].
3. The comparison of serial values to the accepted SOA serial must follow the arithmetric of [RFC 1982]

[RFC 1982]: https://tools.ietf.org/html/rfc1982

mattias-p: Right. MULTIPLE_SOA_SERIALS is emitted when specified. I take that back.

MULTIPLE_SOA_SERIALS_OK on the other hand is a new message that is never emitted.

matsduf: Correct. It is never emitted by the code and it is never emitted by the updated specification as long as accepted SOA sercial difference is zero.

What is the problem of approving?

mattias-p: We shouldn't be declaring messages that we're not emitting. If you remove the MULTIPLE_SOA_SERIALS_OK step I will approve. It can be added in a later PR together with a corresponding implementation.

matsduf: @mattias-p, without *MULTIPLE_SOA_SERIALS_OK* the logic in the test case is broken.

I would have understood your objection if there was no plan to make it possible to emit the message, but as you can read from my note above there is plan to make it possible.

mattias-p: I don't see how the spec is broken without the MULTIPLE_SOA_SERIALS_OK message.

I also don't see why this new message cannot be added together with the new code that is going to be written. To me, that would be the reasonable thing to do.

I don't think we want to make this release with promises about a new message that is never emitted. In general I think we want to avoid even introducing incorrectness in the develop branch when we can avoid it. In this particular case it's even easy to avoid it - just defer the merging of the specification of the new message until it's actually implemented.

mattias-p:
Could you add the `SOA_SERIAL` message to this specification?

matsduf:
SOA_SERIAL serial added. @mattias-p, please review again. Please approve if possible.

@sandoche2k, can you review?

docs/specifications/tests/Consistency-TP/consistency01.md
```diff
@@ -33,27 +33,37 @@ serial number consistency.
 
  1. Obtain the list of name server IPs for the *child zone* from [Method4] 
     and [Method5].
+
  2. Create an SOA query for *child zone* apex and send it to all name 
     server IPs.
+
  3. Retrieve the SOA RR from the responses from all name server IPs.
```
mattias-p: Wouid it be possible to more this step down until after we've checked which responses have SOA RR sets? I think the flow of the TC would be improved that way.
Also, why do we extract the entire SOA RR and not just the SOA/SERIAL?

docs/specifications/tests/Consistency-TP/consistency01.md
```diff
@@ -1,52 +1,97 @@
...
 
-All authoritative name servers must have consistent serial numbers. If the
-test does not find any inconsistency, then the test case passes.
+ 3. If the name server does not respond with a DNS response, emit 
+    *[NO_RESPONSE]* and go to next name server IP.
 
-### Special procedural requirements	
+ 4. If the name server responds with a DNS response but no a SOA record 
+    is included, emit *[NO_RESPONSE_SOA_QUERY]* and got to next name 
+    server IP.
```
mattias-p: "got to next" -> "go to next"

docs/specifications/tests/Consistency-TP/consistency01.md
```diff
@@ -1,52 +1,97 @@
...
+
+ 7. If at least two serial numbers are different:
+    1. Order the serial number values from smallest to largest following
+       the arithmetic for serial number.
+    2. If there is not a single, uniquely defined order of the serial 
+       numbers, emit *[SOA_SERIAL_VARIATION]* and *[MULTIPLE_SOA_SERIALS]*.
+    3. If the difference between the first and the last serial number
+       is larger than *Accepted Serial Difference*, using arithemtic
+       for serial number, emit *[SOA_SERIAL_VARIATION]* and 
+       *[MULTIPLE_SOA_SERIALS]*.
```
mattias-p: "arithemtic" -> "arithmetic"

Same goes for next step too.

docs/specifications/tests/Consistency-TP/consistency01.md
```diff
@@ -56,17 +101,35 @@ A manual inspection of the SOA serial may be needed to determine if the zone
 updates work properly or not, and if the serial values are within a
 reasonable range, then the test is OK.
 
-### Intercase dependencies
+When comparing SOA serial it must be done using the aritmetic defined in
+[RFC 1982].
```
mattias-p: "aritmetic" -> "arithmetic"

matsduf:
@mattias-p, can you approve it now? It is easier to keep track of which ones are done and which ones that are still open. I have also asked @johaniis to review the ones that are approved by at least one person.

@sandoche2k, can you please review it?

docs/specifications/tests/Consistency-TP/consistency01.md
```diff
@@ -1,52 +1,97 @@
...
-     2. If the difference is greater than the minor value, then the test
-        case fails.
-     3. If the difference is less than or equal to the minor value,
-        then the test is passed.
+ 1. Obtain the list of name server IPs for the *Child Zone* from [Method4] 
+    and [Method5].
 
-### Outcome(s)
+ 2. Create an SOA query for *Child Zone* apex and send it to each name 
+    server IP.
```
mattias-p: The use of "apex" here is meant to be educational, right? Have you considered separating this step into two clearly separate parts - on purely instructional and one purely educational? The current wording is helpful to the already knowledgeable people, but it's confusing to the very people who could use the help.

docs/specifications/tests/Consistency-TP/consistency01.md
```diff
@@ -1,52 +1,97 @@
...
+
+ 5. Retrieve the SOA SERIAL from the responses and go to next name server
+    IP until all name server IPs have been queried.
+
+ 6. If at least one SOA SERIAL has been retrieved and all serial 
+    numbers are identical, emit *[ONE_SOA_SERIAL]*.
+
+ 7. If at least two serial numbers are different:
+    1. Order the serial number values from smallest to largest following
+       the arithmetic for serial number.
```
mattias-p: The specified comparison works for two numbers, but the operator is non-transitive so this step is impossible to implement for arbitrary inputs. For example, given SERIAL_BITS=4, each of the SOA serials 0, 5, 10 is greater than one of the other two and lesser than the other. We should have a correct way to handle zones whose SOA serial have gone bonkers.

mattias-p: Have you considered adding an "if possible" here?

matsduf: fixed.

docs/specifications/tests/Consistency-TP/consistency01.md
matsduf: Removed "apex". Not needed here.

docs/specifications/tests/Consistency-TP/consistency01.md
```diff
@@ -1,52 +1,97 @@
...
+    and [Method5].
 
-### Outcome(s)
+ 2. Create an SOA query for *Child Zone* apex and send it to each name 
+    server IP.
 
-All authoritative name servers must have consistent serial numbers. If the
-test does not find any inconsistency, then the test case passes.
+ 3. If the name server does not respond with a DNS response, emit 
+    *[NO_RESPONSE]* and go to next name server IP.
```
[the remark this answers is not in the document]
matsduf: Fixed.

docs/specifications/tests/Consistency-TP/consistency01.md
```diff
@@ -1,52 +1,97 @@
...
 
-### Special procedural requirements	
+ 4. If the name server responds with a DNS response but no a SOA record
```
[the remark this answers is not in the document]
matsduf: Fixed.

PR #623, opened 2018-05-07
Testing preparation files for Debian and CentOS

sandoche2k:


docs/internal-documentation/distrib-testing/Centos-preparation.md
```diff
@@ -0,0 +1,223 @@
...
+   ```sh
+   yum update
+   yum groupinstall "Development Tools"
+   ```
+
+3. Install prerequisites
+
+   ```sh
+   yum install cpanminus
+   ```
```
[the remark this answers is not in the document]
sandoche2k: I think we discussed during last F2F that we will move to cpanm. Isn't it correct?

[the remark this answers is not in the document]
mattias-p: I don't recall what conclusion we reached regarding cpan/cpanm. We should have created issues on the subject.

Anyway here are my two cents: I'm all for moving to cpanm. However, more important is that we use the same tool when testing as is prescribed in the respective installation instructions. After all we do the testing to make sure people don't run into problems when installing according to the installation instruction.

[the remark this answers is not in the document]
matsduf: @mattias-p, see https://github.com/dotse/zonemaster/issues/524.

If we use cpanm we have to include installation instructions for it. It requires more modules to be installed compared to using cpan. I am not against using cpanm, but we should be consistent.

sandoche2k:
I am also for moving to "cpanm". I will test again Debian with cpanm.
Centos is already tested with cpanm.

mattias-p:
@sandoche2k Please use the tool that is prescribed by the installation instruction. If/when we're moving to `cpanm`, the installation instruction should be updated first.

sandoche2k:
Please specify which tool.

On Tue, May 8, 2018 at 9:59 AM, Mattias Päivärinta <notifications@github.com

mattias-p:
As of today, all installation instructions use `cpan` so we should use `cpan` when testing. If we like `cpanm` better, we should update the installation instructions to use `cpanm` before we start using `cpanm` when testing. That way we ensure that users who follow the installation instruction (for any given release) use the same tool that was used in the installation testing (for that release).

sandoche2k:
Once the testing instructions are updated, we could update the installation
instructions.

sandoche2k:
@mattias-p and @matsduf i have updated the PR

sandoche2k:
It facilitates to follow this document while preparing the release, rather than going to the release policy document

matsduf:
@sandoche2k, the ReleaseProcess document is not a policy document. It is a step-by-step document how to create the distribution packages. I think it is better to have that in one place instead of several.

sandoche2k:
@mattias-p and @matsduf i think we are reading into this document more than what we should do? 

First of all this is under the directory "internal documentation", meaning to be used by us when we are testing for the release.  It is easy for me if i follow one text and then copy and paste while testing rather than going to different links. 

Anyway i have updated the document as requested. My two cents is that we are losing the objective of this document in trying to be rigorous.

docs/internal-documentation/distrib-testing/Centos-preparation.md
```diff
@@ -0,0 +1,43 @@
...
+
+   ```sh
+   exit
+   ```
+
+
+## Preparing to install the repositories
+
+1. Follow the steps (1-7) of the [release
+process](https://github.com/dotse/zonemaster/blob/master/docs/internal-documentation/maintenance/ReleaseProcess.md) to prepare the tarballs in a separate machine
```
matsduf: I think we want to link to the ReleaseProcess document in the same branch as the document as one is reading:

```
1. Follow the steps (1-7) of the [release process] to prepare the tarballs in a separate machine

[release process]: ../maintenance/ReleaseProcess.md
```

docs/internal-documentation/distrib-testing/Debian-Preparation.md
```diff
@@ -40,3 +39,13 @@
    ```sh
    exit
    ```
+
+
+## Preparing to install the repositories
+
+1. Follow the steps (1-7) of the [release
+process](https://github.com/dotse/zonemaster/blob/master/docs/internal-documentation/maintenance/ReleaseProcess.md)
+to prepare the tarballs in a separate machine
```
matsduf: I think we want to link to the ReleaseProcess document in the same branch as the document as one is reading:

```
1. Follow the steps (1-7) of the [release process] to prepare the tarballs in a separate machine

[release process]: ../maintenance/ReleaseProcess.md
```

docs/internal-documentation/distrib-testing/Centos-preparation.md
matsduf: @mattias-p, I do not agree with you. The preparations are to create a machine to test on. The ReleaseProcesses is to create packages to install. I think the document is fine if the last link could be updated to point at the ReleaseProcess document in the same branch.

mattias-p: @matsduf Testing is an important part of the release process, not the other way around.

sandoche2k:
@matsduf @mattias-p Since this document will not be used by users, i think we can conclude on what we really need to put here after this release.

mattias-p:
Created #631 as a reminder to come back to this.

PR #633, opened 2018-06-13
Update syntax06

matsduf:
* Converted to new format.
* Updated reference to new RFCs (obsoleting previous).
* Made the steps fully explcit.
* Added a message NO_RESPONSE.
* Added a message RNAME_RFC822_VALID for the situation that the RNAME
  address is fully validated.
* Added check for mail domain in DNS and a message RNAME_MAIL_DOMAIN_INVALID

This update is not compatible with the current implementation.

### DIFF
--- a/docs/specifications/tests/Syntax-TP/syntax06.md
+++ b/docs/specifications/tests/Syntax-TP/syntax06.md
@@ -58,9 +58,10 @@ field should follow the rules of an e-mail address also defined in
        mail exchange (domain name in RDATA of the MX record) do:
        1. Create address queries (A and AAAA) and send it to a 
           resolving name server.
-       2. If the mail exchange is "localhost" then consider all A and
-          AAAA lookups as failing.
-       3. Ignore any CNAME replies (i.e. do not follow them).
+       2. If the mail exchange name is "localhost" then consider all 
+          A and AAAA lookups as failing.
+       3. If the mail exchange name points at a CNAME, then ignore
+          any A or AAAA records, i.e. consider them as not existing.
        4. Disregard any A record with 127.0.0.1 or AAAA with ::1.
        5. If all MX have been processed and neither A or AAAA record 
           was returned in the answer section for any mail exchange then 
--- a/docs/specifications/tests/Syntax-TP/syntax06.md
+++ b/docs/specifications/tests/Syntax-TP/syntax06.md
@@ -19,7 +19,7 @@ field should follow the rules of an e-mail address also defined in
 1. Obtain the set of name server IP addresses using [Method4] and [Method5]
    ("Name Server IP").
 
-2. Create a SOA query for the apex of the *Child Zone* with "rd" flag unset.
+2. Create a SOA query for the apex of the *Child Zone* with RD flag unset.
 
 3. For each name server in *Name Server IP* do:
    1. Send the SOA query over UDP to the name server.
@@ -43,10 +43,13 @@ field should follow the rules of an e-mail address also defined in
    9. Create an MX query for the domain part and send it to a resolving
        name server. 
    10. If the MX lookup does not return a DNS response with with RCODE 
-       "NOERROR" emit *[RNAME_MAIL_DOMAIN_INVALID]*.
+       "NOERROR", then:
+       1. Emit *[RNAME_MAIL_DOMAIN_INVALID]*.
+       2. Go to next server.
    11. If the MX lookup returned a CNAME or a chain of CNAMEs then
        use the final name instead of the domain part in the steps below.
-   12. If the MX lookup returned no MX record, then 
+   12. If the MX lookup returned a NO DATA response (no MX record), 
+       then 
        1. Create address queries (A and AAAA) for the domain part and
           send it to a resolving name server.
        2. Disregard any A record with 127.0.0.1 or AAAA with ::1.
@@ -65,10 +68,8 @@ field should follow the rules of an e-mail address also defined in
        5. If all MX have been processed and neither A or AAAA record 
           was returned for any mail exchange then emit 
           *[RNAME_MAIL_DOMAIN_INVALID]*.
-
-4. If at least one RNAME has been fully validated and no 
-   *[RNAME_RFC822_INVALID]* or *[RNAME_MAIL_DOMAIN_INVALID]*
-   has not been emitted, then emit *[RNAME_RFC822_VALID]*.
+   14. If no *[RNAME_MAIL_DOMAIN_INVALID]* has not been emitted, 
+       then emit *[RNAME_RFC822_VALID]* for that RNAME.
 
 
 ## Outcome(s)
--- a/docs/specifications/tests/Syntax-TP/syntax06.md
+++ b/docs/specifications/tests/Syntax-TP/syntax06.md
@@ -30,7 +30,8 @@ field should follow the rules of an e-mail address also defined in
       answer section, then:
       1. Output *[NO_RESPONSE_SOA_QUERY]*.
       2. Go to next server.
-   4. Extract the RNAME from the SOA record.
+   4. Extract the RNAME from the SOA record (from the first SOA record if
+      multiple).
    5. Convert the first "." without backslash quoting to an "@" in 
       the RNAME.
    6. Convert any backslash quoted "." to a single "." without quoting
@@ -41,36 +42,39 @@ field should follow the rules of an e-mail address also defined in
       2. Go to next server.
    8. Extract the domain part (to the right of "@") from the *Mail 
       address* ("Domain Part").
-   9. Create an MX query for the *Domain Part* and do a [DNS Lookup]
-      of that query. 
+   9. Create an MX query for the *Domain Part* and do a
+      [DNS Lookup][terminology] of that query. 
    10. If the lookup of MX does not return a DNS response with RCODE 
        "NOERROR", then:
        1. Output *[RNAME_MAIL_DOMAIN_INVALID]*.
        2. Go to next server.
-   11. If the MX lookup returned a CNAME or a chain of CNAMEs then
-       set *Domain Part* to be the the final name instead of the
-       domain from the email address.
+   11. For the MX lookup, CNAME or a chain if CNAMEs is followed, if
+       any. If MX record or records are found via CNAME, then
+       set *Domain Part* to be the MX owner name (instead of the
+       domain part of *Mail Address*).
    12. If the MX lookup returned a NO DATA response (no MX record), 
        then:
        1. Create address queries (A and AAAA) for the *Domain Part*
-          and do [DNS Lookups][DNS Lookup] of those queries. 
-       2. Disregard any A record with 127.0.0.1 or AAAA with ::1.
-       3. If no A or AAAA records with the same owner name as *Domain
+          and do [DNS Lookup][terminology] of those queries. 
+       2. Disregard all A and AAAA records outside the answer section.
+       3. Disregard any A record with 127.0.0.1 or AAAA with ::1.
+       4. If no A or AAAA records with the same owner name as *Domain
           Part* were found in the responses 
           then output *[RNAME_MAIL_DOMAIN_INVALID]*.
-   13. If the MX lookup returned one or more MX records, then for each 
-       domain name in RDATA of the MX record ("Mail Exchange") do:
-       1. Create address queries (A and AAAA) and do 
-          [DNS Lookups][DNS Lookup] of those queries. 
-       2. Disregard all A and AAAA records that do not have the same
+   13. If the MX lookup returned one or more MX records, then for each
+       MX record extract the domain name in RDATA ("Mail Exchange") 
+       and do:
+       1. Create address queries (A and AAAA) of *Mail Exchange* 
+          and do [DNS Lookup][terminology] of those queries. 
+       2. Disregard all A and AAAA records outside the answer section.
+       3. Disregard any A record with 127.0.0.1 or AAAA with ::1.
+       4. Disregard all A and AAAA records that do not have the same
           owner name as *Mail Exchange* (i.e. do not follow
           any CNAME).
-       3. Disregard all A and AAAA records outside the answer section.
-       4. Disregard any A record with 127.0.0.1 or AAAA with ::1.
-       5. If all MX have been processed and neither A or AAAA record 
-          was returned for any mail exchange, then output 
+   14. If the MX lookup returned one or more MX records and neither 
+          A nor AAAA record was returned for any mail exchange, then output 
           *[RNAME_MAIL_DOMAIN_INVALID]*.
-   14. If no *[RNAME_MAIL_DOMAIN_INVALID]* has been outputted, 
+   15. If no *[RNAME_MAIL_DOMAIN_INVALID]* has been outputted, 
        then output *[RNAME_RFC822_VALID]* for that RNAME.
 
 
@@ -123,6 +127,7 @@ respected.
 [RFC 1035]:                https://tools.ietf.org/html/rfc1035
 [RFC 1912]:                https://tools.ietf.org/html/rfc1912
 [RFC 5322, section 3.4.1]: https://tools.ietf.org/html/rfc5322#section-3.4
+[terminology]:             #terminology
 
 [NO_RESPONSE]:               #outcomes
 [NO_RESPONSE_SOA_QUERY]:     #outcomes
--- a/docs/specifications/tests/Syntax-TP/syntax06.md
+++ b/docs/specifications/tests/Syntax-TP/syntax06.md
@@ -51,9 +51,13 @@ field should follow the rules of an e-mail address also defined in
           send it to a resolving name server.
        2. If the domain part is "localhost" then consider all A and
           AAAA lookups as failing.
-       3. Disregard any A record with 127.0.0.1 or AAAA with ::1.
-       4. If neither A or AAAA record was returned for the domain part
-          in the answer section then emit *[RNAME_MAIL_DOMAIN_INVALID]*.
+       3. Disregard all A and AAAA records that do not have the same
+          owner name as the mail domain (i.e. do not follow
+          any CNAME).
+       4. Disregard any A record with 127.0.0.1 or AAAA with ::1.
+       5. Disregard any A or AAAA records outside the Answer section.
+       6. If no A or AAAA was returned then emit 
+          *[RNAME_MAIL_DOMAIN_INVALID]*.
    13. If the MX lookup returned one or more MX records, then for each
        mail exchange (domain name in RDATA of the MX record) do:
        1. Create address queries (A and AAAA) and send it to a 
@@ -64,9 +68,10 @@ field should follow the rules of an e-mail address also defined in
           owner name as the mail exchange name (i.e. do not follow
           any CNAME).
        4. Disregard any A record with 127.0.0.1 or AAAA with ::1.
-       5. If all MX have been processed and neither A or AAAA record 
-          was returned in the answer section for any mail exchange then 
-          emit *[RNAME_MAIL_DOMAIN_INVALID]*.
+       5. Disregard any A or AAAA records outside the Answer section.
+       6. If all MX have been processed and neither A or AAAA record 
+          was returned for any mail exchange then emit 
+          *[RNAME_MAIL_DOMAIN_INVALID]*.
 
 4. If at least one RNAME has been fully validated and no 
    *[RNAME_RFC822_INVALID]* or *[RNAME_MAIL_DOMAIN_INVALID]*
--- a/docs/specifications/tests/Syntax-TP/syntax06.md
+++ b/docs/specifications/tests/Syntax-TP/syntax06.md
@@ -55,7 +55,7 @@ field should follow the rules of an e-mail address also defined in
    12. If the MX lookup returned a NO DATA response (no MX record), 
        then:
        1. Create address queries (A and AAAA) for the *Domain Part*
-          and do [DNS Lookup][terminology] of those queries. 
+          and do [DNS Lookups][terminology] of those queries. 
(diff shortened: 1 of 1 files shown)


docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,114 @@
...
+2. Create a SOA query for the apex of the *Child Zone* with "rd" flag unset.
 
-### Outcome(s)
+3. For each name server in *Name Server IP* do:
+   1. Send the SOA query over UDP to the name server.
+   2. If the name server does not respond, emit *[NO_RESPONSE]* and go
+      to next server.
+   3. If the name server responds with DNS response but does not 
+      include a SOA record in the answer section, emit 
+      *[NO_RESPONSE_SOA_QUERY]* and go to next server.
```
mattias-p: There's no explicit requirement for an AA flag here. Is that correct?

Also. There's a lot going on here. Do you think you could split this big step into smaller ones?

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,114 @@
...
+      (see section 5.1, 5.3 and 8 in [RFC1035] for the use of backslash).
+   7. If the converted string (mail address) does not meet the 
+      specification in [RFC 5322, section 3.4.1], then emit 
+      *[RNAME_RFC822_INVALID]* and go to next server.
+   8. Extract the domain part (to the right of "@") from the mail 
+      address.
+   9. Create an MX query for the domain part and send it to a resolving
+      name server. If the lookup does not return a DNS response with
+      with either RCODE "NOERROR" or "NXDOMAIN" emit 
+      *[RNAME_MAIL_DOMAIN_INVALID]*.
```
mattias-p: I think the steps from here on feel somewhat convoluted. Do you think they could be expressed more clearly?

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,113 @@
...
+      name server. If the lookup does not return a DNS response with
+      with RCODE "NOERROR" emit *[RNAME_MAIL_DOMAIN_INVALID]*.
+   10. If the lookup in previous step returned MX record or records for 
+       the mail domain, then replace the mail domain with the mail 
+       exchange or exchanges from RDATA of the MX record or records in
+       next step.
+   11. Create address querie (A and AAAA) for the mail domain or mail
+       exchage in previous steps (multiple queries if there are multiple
+       mail exchanges). At least one query must successfully return
+       an IP address or else emit *[RNAME_MAIL_DOMAIN_INVALID]*.
```
mattias-p: Skulle det vara rimligt att skriva ut vad som krävs av svarspaketen m.a.p. rcode, flaggor och sektioner, eller är det överflödigt?

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,113 @@
...
+      *[RNAME_RFC822_INVALID]* and go to next server.
+   8. Extract the domain part (to the right of "@") from the mail 
+      address.
+   9. Create an MX query for the domain part and send it to a resolving
+      name server. If the lookup does not return a DNS response with
+      with RCODE "NOERROR" emit *[RNAME_MAIL_DOMAIN_INVALID]*.
+   10. If the lookup in previous step returned MX record or records for 
+       the mail domain, then replace the mail domain with the mail 
+       exchange or exchanges from RDATA of the MX record or records in
+       next step.
```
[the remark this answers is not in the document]
matsduf: All steps are needed to do complete test. I have updated to be more explicit (more, but smaller steps).

docs/specifications/tests/Syntax-TP/syntax06.md
matsduf: In this case we are sending the queries to a resolving name server and accepts what we get without checking flags.

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -39,16 +41,30 @@ field should follow the rules of an e-mail address also defined in
...
+       3. Disregard any A record with 127.0.0.1 or AAAA with ::1.
+       4. If neither A or AAAA record was returned for the domain part
+          in the answer section then emit *[RNAME_MAIL_DOMAIN_INVALID]*.
+   14. If the MX lookup returned one or more MX records, then for each
+       mail exchange (domain name in RDATA of the MX record) do:
+       1. Create address queries (A and AAAA) and send it to a 
+          resolving name server.
+       2. If the mail exchange is "localhost" then consider all A and
+          AAAA lookups as failing.
+       3. Ignore any CNAME replies (i.e. do not follow them).
```
mattias-p: Won't the resolving name server resolve the CNAMEs?

matsduf: Yes, but do not follow/accept those. If the MX is

```
sth.namn.se.  MX 0 mail.namn.se.
```
then we do an address lookup of mail.namn.se and we might get from the resolver
```
;; ANSWER SECTION:
mail.namn.se.  CNAME  post.namn.se.
post.namn.se.  A 192.0.2.10
```
Then we MUST ignore that A record, i.e. do not follow the CNAME.

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,131 @@
...
+       2. Disregard all A and AAAA records that do not have the same
+          owner name as the mail exchange name (i.e. do not follow
+          any CNAME).
+       3. Disregard any A record with 127.0.0.1 or AAAA with ::1.
+       4. Disregard any A or AAAA records outside the Answer section.
+       5. If all MX have been processed and neither A or AAAA record 
+          was returned for any mail exchange then emit 
+          *[RNAME_MAIL_DOMAIN_INVALID]*.
+   14. If no *[RNAME_MAIL_DOMAIN_INVALID]* has not been emitted, 
+       then emit *[RNAME_RFC822_VALID]* for that RNAME.
```
mattias-p: Too many negations.

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,131 @@
...
+       3. Disregard any A or AAAA records outside the Answer section.
+       4. If no A or AAAA was returned then emit 
+          *[RNAME_MAIL_DOMAIN_INVALID]*.
+   13. If the MX lookup returned one or more MX records, then for each
+       mail exchange (domain name in RDATA of the MX record) do:
+       1. Create address queries (A and AAAA) and send it to a 
+          resolving name server.
+       2. Disregard all A and AAAA records that do not have the same
+          owner name as the mail exchange name (i.e. do not follow
+          any CNAME).
```
mattias-p: There's no corresponding step for this in the no-MX branch. Is that intentional?

mattias-p:
I've got two suggestions that you may want to consider:
1. In the no-MX and MX branches, focus on what information to consider, instead of what information to disregard.
2. Combine the no-MX and MX branches into one by using an intermediate data set that's populated differently depending on the response of the MX query.

docs/specifications/tests/Syntax-TP/syntax06.md
matsduf: Yes.

When we ask for MX it is OK to follow CNAME and that is covered in "If the MX lookup returned a CNAME or a chain of CNAMEs then use the final name instead of the domain part in the steps below."

When we come to the non-MX we have already followed the CNAME and there cannot be any more CNAME to follow.

When we come to the MX then that MX (or those MX) are already based on the CNAME above (if any), and now we do a look-up of the mail exchange in the MX, and that must NOT follow CNAME.

Should I clarify?

docs/specifications/tests/Syntax-TP/syntax06.md
matsduf: Agree. Will fix.

docs/specifications/tests/Syntax-TP/syntax06.md
mattias-p: Ok, I see how you're thinking. I'm looking at this from a different angle. I'm looking at it from an algorithmic point of view as opposed to an educational point of view.

From an algorithmic point of view, these two cases behave differently. In the no-MX case you should consider A and AAAA records with any owner name, and in the MX case you should only consider A and AAAA records with a specific owner name. If the resolver responds with additional A and AAAA records with different owner names for whatever reason, the MX and no-MX cases are bound to behave differently.

I think it would be nice if both branches presented the same behavior. But I don't think it's critical.

matsduf:
@mattias-p, can you please re-review. I have updated after you change request.

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,129 @@
...
+   10. If the MX lookup does not return a DNS response with with RCODE 
+       "NOERROR", then:
+       1. Output *[RNAME_MAIL_DOMAIN_INVALID]*.
+       2. Go to next server.
+   11. If the MX lookup returned a CNAME or a chain of CNAMEs then
+       use the final name instead of the domain part in the steps below.
+   12. If the MX lookup returned a NO DATA response (no MX record), 
+       then 
+       1. Create address queries (A and AAAA) for the domain part (or 
+          updated by CNAMEs above) and send it to a resolving
```
mattias-p: Have you considered giving a name to "domain part (or updated by CNAMEs above)"?

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,129 @@
...
+      (see section 5.1, 5.3 and 8 in [RFC1035] for the use of backslash).
+   7. If the converted string (mail address) does not meet the 
+      specification in [RFC 5322, section 3.4.1], then 
+      1. Output *[RNAME_RFC822_INVALID]*.
+      2. Go to next server.
+   8. Extract the domain part (to the right of "@") from the mail 
+      address.
+   9. Create an MX query for the domain part and send it to a resolving
+       name server. 
+   10. If the MX lookup does not return a DNS response with with RCODE
```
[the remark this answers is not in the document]
matsduf: Fixed.

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,129 @@
...
+   9. Create an MX query for the domain part and send it to a resolving
+       name server. 
+   10. If the MX lookup does not return a DNS response with with RCODE 
+       "NOERROR", then:
+       1. Output *[RNAME_MAIL_DOMAIN_INVALID]*.
+       2. Go to next server.
+   11. If the MX lookup returned a CNAME or a chain of CNAMEs then
+       use the final name instead of the domain part in the steps below.
+   12. If the MX lookup returned a NO DATA response (no MX record), 
+       then
```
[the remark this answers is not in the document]
matsduf: Fixed.

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,129 @@
...
+   13. If the MX lookup returned one or more MX records, then for each
+       mail exchange (domain name in RDATA of the MX record) do:
+       1. Create address queries (A and AAAA) and send it to a 
+          resolving name server.
+       2. Disregard all A and AAAA records that do not have the same
+          owner name as the mail exchange name (i.e. do not follow
+          any CNAME).
+       3. Disregard any A record with 127.0.0.1 or AAAA with ::1.
+       4. If all MX have been processed and neither A or AAAA record 
+          was returned for any mail exchange then output
```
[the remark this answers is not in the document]
matsduf: Fixed.

docs/specifications/tests/Syntax-TP/syntax06.md
matsduf: Fixed.

docs/specifications/tests/Syntax-TP/syntax06.md
matsduf: Will not update.

sandoche2k:
@matsduf should we move this to 2019.1?

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,129 @@
...
+       "NOERROR", then:
+       1. Output *[RNAME_MAIL_DOMAIN_INVALID]*.
+       2. Go to next server.
+   11. If the MX lookup returned a CNAME or a chain of CNAMEs then
+       set *Domain Part* to be the the final name instead of the
+       domain from the email address.
+   12. If the MX lookup returned a NO DATA response (no MX record), 
+       then:
+       1. Create address queries (A and AAAA) for the *Domain Part*
+          and send it to a resolving name server.
```
mattias-p: send it -> send them?

There's another one like this below.

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,129 @@
...
+       2. Disregard any A record with 127.0.0.1 or AAAA with ::1.
+       3. If no A or AAAA records with the same owner name as *Domain
+          Part* were found in the responses 
+          then output *[RNAME_MAIL_DOMAIN_INVALID]*.
```
mattias-p: Should A records outside of the answer section in the response to the A query be considered (and analogously for AAAA records)?

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,129 @@
...
+   13. If the MX lookup returned one or more MX records, then for each
+       mail exchange (domain name in RDATA of the MX record) do:
```
mattias-p: for each mail exchange (domain name in RDATA of the MX record) do -> for each domain name in RDATA of the MX record ("Mail Exchange") do

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,129 @@
...
+       1. Create address queries (A and AAAA) and send it to a 
+          resolving name server.
+       2. Disregard all A and AAAA records that do not have the same
+          owner name as the mail exchange name (i.e. do not follow
+          any CNAME).
+       3. Disregard any A record with 127.0.0.1 or AAAA with ::1.
+       4. If all MX have been processed and neither A or AAAA record 
+          was returned for any mail exchange, then output 
+          *[RNAME_MAIL_DOMAIN_INVALID]*.
```
mattias-p: I really think it's improper to check if the foreach-loop has completed its work from within the foreach loop itself.
The problem could be solved using much simpler constructs. E.g. by introducing a single boolean variable. The variable could be initialized to false before the loop, set to true when relevant addresses are found inside the loop, and checked after the loop to determine whether RNAME_MAIL_DOMAIN_INVALID is to be emitted.

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,129 @@
...
+3. For each name server in *Name Server IP* do:
+   1. Send the SOA query over UDP to the name server.
+   2. If the name server does not respond with a DNS response, then:
+      1. Output *[NO_RESPONSE]*.
+      2. Go to next server.
+   3. If the DNS response does not include a SOA record in the 
+      answer section, then:
+      1. Output *[NO_RESPONSE_SOA_QUERY]*.
+      2. Go to next server.
+   4. Extract the RNAME from the SOA record.
```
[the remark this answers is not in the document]
matsduf: It is illegal to return more than one SOA record for a zone. My experience is that it is very unusual that it happens, so that is why I have chosen to ignore that case. We query for SOA in many specifications. Should check for multiple SOA records in all cases?

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,129 @@
...
+   5. Convert the first "." without backslash quoting to an "@" in 
+      the RNAME.
+   6. Convert any backslash quoted "." to a single "." without quoting
+      (see section 5.1, 5.3 and 8 in [RFC1035] for the use of backslash).
+   7. If the converted string (mail address) does not meet the
```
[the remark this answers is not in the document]
matsduf: Updated.

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,129 @@
...
+   5. Convert the first "." without backslash quoting to an "@" in 
+      the RNAME.
+   6. Convert any backslash quoted "." to a single "." without quoting
+      (see section 5.1, 5.3 and 8 in [RFC1035] for the use of backslash).
+   7. If the converted string (mail address) does not meet the 
+      specification in [RFC 5322, section 3.4.1], then 
+      1. Output *[RNAME_RFC822_INVALID]*.
+      2. Go to next server.
+   8. Extract the domain part (to the right of "@") from the mail 
+      address ("Domain Part").
```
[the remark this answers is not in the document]
matsduf: Updated.

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,129 @@
...
+   9. Create an MX query for the *Domain Part* and send it to a resolving
+       name server.
```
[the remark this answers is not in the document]
matsduf: Good catch. Updated.

docs/specifications/tests/Syntax-TP/syntax06.md
matsduf: Fixed.

docs/specifications/tests/Syntax-TP/syntax06.md
matsduf: No. Updated to clarify.

docs/specifications/tests/Syntax-TP/syntax06.md
matsduf: Accepted.

docs/specifications/tests/Syntax-TP/syntax06.md
matsduf: For a computer program I agree, but for a human program I think the current solution is simpler.

docs/specifications/tests/Syntax-TP/syntax06.md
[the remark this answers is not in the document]
mattias-p: I agree that validation of the number of SOA records should only be validated by a single test case. Maybe we could have a comment to use the first SOA record in case more than one are present?

sandoche2k:
@mattias-p do you agree with @matsduf final changes

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,134 @@
...
+       "NOERROR", then:
+       1. Output *[RNAME_MAIL_DOMAIN_INVALID]*.
+       2. Go to next server.
+   11. If the MX lookup returned a CNAME or a chain of CNAMEs then
+       set *Domain Part* to be the the final name instead of the
+       domain from the email address.
+   12. If the MX lookup returned a NO DATA response (no MX record), 
+       then:
+       1. Create address queries (A and AAAA) for the *Domain Part*
+          and do [DNS Lookups][DNS Lookup] of those queries.
```
mattias-p: This link doesn't work.

docs/specifications/tests/Syntax-TP/syntax06.md
mattias-p: It looks like you clarified this in step 3.13 but not here in step 3.12.

docs/specifications/tests/Syntax-TP/syntax06.md
mattias-p: I disagree with the notion of "human programs". All the programs we care about are maintained by humans.

docs/specifications/tests/Syntax-TP/syntax06.md
```diff
@@ -1,54 +1,134 @@
...
+      address* ("Domain Part").
+   9. Create an MX query for the *Domain Part* and do a [DNS Lookup]
+      of that query. 
+   10. If the lookup of MX does not return a DNS response with RCODE 
+       "NOERROR", then:
+       1. Output *[RNAME_MAIL_DOMAIN_INVALID]*.
+       2. Go to next server.
+   11. If the MX lookup returned a CNAME or a chain of CNAMEs then
+       set *Domain Part* to be the the final name instead of the
+       domain from the email address.
```
mattias-p: email address -> \*Mail address\*?

Maybe even put the "instead of"part within parentheses?

docs/specifications/tests/Syntax-TP/syntax06.md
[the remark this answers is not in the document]
matsduf: A new test case has to be created (#666). Here we will just take the first. This is valid for many test cases.

docs/specifications/tests/Syntax-TP/syntax06.md
matsduf: Adjusted.

docs/specifications/tests/Syntax-TP/syntax06.md
matsduf: Should be fixed now.

docs/specifications/tests/Syntax-TP/syntax06.md
matsduf: Updated and made in consistent in 3.12 and 3.13.

docs/specifications/tests/Syntax-TP/syntax06.md
matsduf: Found a solution without control variable. The text is written to be read, not to be executed.

PR #643, opened 2018-09-10
CDS consistency check

huguei:
This test case description have its companion implementation on zonemaster-engine project:
https://github.com/zonemaster/zonemaster-engine/pull/434

### DIFF
--- a/docs/specifications/tests/Consistency-TP/consistency06.md
+++ b/docs/specifications/tests/Consistency-TP/consistency06.md
@@ -25,7 +25,7 @@ The domain name to be tested.
 ### Ordered description of steps to be taken to execute the test case
 
  1. Obtain the list of name servers from [Method4] and [Method5].
- 2. Retrieve the CDS RR from all the name servers.
+ 2. Retrieve the CDS RR from all the name servers using TCP.
  3. The CDS RR RDATA value in its "string representation" must be the
     same for all servers from step 2.
 
@@ -43,6 +43,12 @@ on the ignored result.
 If there's no CDS record in the zone, we consider such absence as a normal
 answer, so it should be consistent between all the authoritative
 nameservers to pass the test.
+
+When bootstraping a new CDS key for setting up a DNSSEC delegation for
+the first time, there's a need to have extra security defenses against
+UDP injection attacks. Therefore, the transport for the query should be
+forced to TCP. If there's no TCP response from any nameserver, the test
+should fail.
 
 ### Intercase dependencies
 
--- a/docs/specifications/tests/Consistency-TP/consistency06.md
+++ b/docs/specifications/tests/Consistency-TP/consistency06.md
@@ -1,0 +1,60 @@
+## CONSISTENCY06: CDS record consistency
+
+### Test case identifier
+
+**CONSISTENCY06:** CDS record consistency
+
+### Objective
+
+To be able to participate on the automation of DNSSEC delegation trust
+maintenance (RFC7344), a zone must have the same CDS record content
+between all the authoritative servers.
+
+The objective of this test is to verify that the CDS record RDATA value
+is consistent between different authoritative name servers.
+
+For reference purposes : [RFC7344](https://tools.ietf.org/html/rfc7344) 
+explains the CDS record and consistency requirements, and section 5.1 of 
+[RFC 4034](https://tools.ietf.org/html/rfc4034) explains the format
+of the DS format RDATA, from which CDS is derived.
+
+### Inputs
+
+The domain name to be tested.
+
+### Ordered description of steps to be taken to execute the test case
+
+ 1. Obtain the list of name servers from [Method4] and [Method5].
+ 2. Retrieve the CDS RR from all the name servers.
+ 3. The CDS RR RDATA value in its "string representation" must be the
+    same for all servers from step 2.
+
+### Outcome(s)
+
+All authoritative name servers must have consistent CDS RR RDATA. If
+the test does not find any inconsistency, then the test case passes.
+
+### Special procedural requirements	
+
+If either IPv4 or IPv6 transport is disabled, ignore the evaluation of the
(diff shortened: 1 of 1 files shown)


mattias-p:
Thanks for the contribution. Much appreciated!

PRs should be made against the develop branch (it says so [here](https://github.com/zonemaster/zonemaster#participation)). Another test case named consistency06 has already been merged in the develop branch so this one should be renamed to consistency07.

huguei:
Thanks. I withdraw the last PR and create a new one over develop branch
https://github.com/zonemaster/zonemaster-engine/pull/437

sandoche2k:
@huguei i moved it to develop branch and renamed to consistency07. Please verify.

huguei:
Thanks, it looks good.

matsduf:
@huguei, I have some changes that I want to see in the proposed specification. Do you prefer if I push commits to your branch or do you prefer if I create a pull request against your branch?

matsduf:
@huguei, your PR only contains check of CDS records. What about CDNSKEY records?  From RFC 7344:

> If the Child publishes either the CDS or the CDNSKEY resource record,
> it SHOULD publish both.  If the Child knows which the Parent
> consumes, it MAY choose to only publish that record type (for
> example, some Children wish the Parent to publish a DS, but they wish
> to keep the DNSKEY "hidden" until needed).  If the Child publishes
> both, the two RRsets MUST match in content.

matsduf:
@huguei and @sandoche2k, I have updated the proposed Test Case. Firstly, I have moved it into the new format of specifications. Secondly, after reading the [RFC 7344] there were some things that needed updates, e.g. support of multiple CDS records. There are still some areas where I want to have more discussions. The RFC standardizes two record types, CDS and CDNSKEY, and also a relation between the two. This proposal to Test Case only covers CDS. Should CDNSKEY also be covered? The RFC also has requirements on the RRSIG for the CDS (and CDNSKEY). It must be there and the signer must be some specific DNSKEY. Should that also be verified? In this Test Case or in another?

Please review!

[RFC 7344]: https://tools.ietf.org/html/rfc7344

matsduf:
@huguei, please wait with the fix until the test case is completed.

For me it seems like it would be best to check for both CDS and CDNSKEY in the same Test Case since they are connected according to the RFC. The RFC says that both SHOULD be there, but not MUST. I do not think we should have a new module for that. Rather add another Test Case in Zone that checks for the RRSIG etc.

The RFC never talks about TCP. Why do you want to require TCP? Is there any other document calling for TCP?

huguei:
@matsduf ok, works for me. Let's add CDNSKEY to the same test and let the RRSIG ones in another test.

About TCP, it's not explicitely mentioned in the RFC, but the security considerations in point 9 "...If it is used, strict rules for inclusion of keys -- such as hold-down times, challenge data inclusion, or similar -- MUST be used along with some kind of challenge mechanism." are being deployed in the TLDs as a best practice of requiring TCP transport for querying CDS, because protects better against injection attacks. And as TCP is a MUST transport for DNS ([RFC7766](https://www.rfc-editor.org/rfc/rfc7766.txt)), then we should prefer TCP over UDP for this specific case. Should do add an "intercase dependency" in the test case specification, pointing to Connectivity/connectivity02 ?

huguei:
> About TCP, it's not explicitely mentioned in the RFC, but the security considerations in point 9 "...If it is used, strict rules for inclusion of keys -- such as hold-down times, challenge data inclusion, or similar -- MUST be used along with some kind of challenge mechanism." are being deployed in the TLDs as a best practice of requiring TCP transport for querying CDS, because protects better against injection attacks. And as TCP is a MUST transport for DNS ([RFC7766](https://www.rfc-editor.org/rfc/rfc7766.txt)), then we should prefer TCP over UDP for this specific case. Should do add an "intercase dependency" in the test case specification, pointing to Connectivity/connectivity02 ?

Edit: also see [RFC8078](https://www.rfc-editor.org/rfc/rfc8078.txt) section 3.2 for the extra checks for acceptance.

matsduf:
@huguei, I interpret you that the requirement is that they are available over TCP and that is covered by [Connectivity02] because it will return an error if DNS is not available over TCP. Do we have to test anything else? The use of TCP or UDP for retreiving CDS and CDNSKEY is a question of policy, isn't it?

Thank you for providing the reference to the other RFC on CDS and CDNSKEY.

Let's update the specification to include CDNSKEY too and then create the other Test Case to verify the requirement that the records are signed etc.

[Connectivity02]: https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/Connectivity-TP/connectivity02.md

huguei:
I withdraw this pull request and created a new one on https://github.com/zonemaster/zonemaster/pull/648
Thanks.

matsduf:
See https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/DNSSEC-TP/dnssec15.md, https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/DNSSEC-TP/dnssec16.md, https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/DNSSEC-TP/dnssec17.md and https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/DNSSEC-TP/dnssec18.md of which -15, -16 and -17 will be included in the v2021.1 release due at the end of May 2021.

matsduf:
@huguei, Friday 2021-05-21 we will speak about Zonemaster at a LACTLD meeting. Will you be there?

huguei:
> @huguei, Friday 2021-05-21 we will speak about Zonemaster at a LACTLD meeting. Will you be there?

Sure, we'll see you there! I suppose it will be more geared towards general or introductory topics. I don't think we can talk about CDS ;)

matsduf:
That is correct. We will just mention CDS/CDNSKEY, but we can take a discussion afterwards via email. You will get my address in our presentation later today.

PR #654, opened 2018-12-12
Test case for testing with unknown version and option together added

sandoche2k:
PR to test the combination of unknown version and option.

### DIFF
--- a/docs/specifications/tests/Nameserver-TP/nameserver14.md
+++ b/docs/specifications/tests/Nameserver-TP/nameserver14.md
@@ -30,13 +30,13 @@ This test case actually combines the test options in test cases
       	next server.
 	3. Else, if the DNS response has the RCODE "FORMERR" then output
       	*[NO_EDNS_SUPPORT]*.
-	4. Else, if the RCODE has "NOERROR", and has an "OPTION-CODE" and
-	"Version Number" greater than zero,  then output
-	[UNKNOWN_OPTION_CODE_VERSION]. 
-	5. Else, if the DNS response has the RCODE "NOERROR" with an OPT record
-	with version greater than zero, then output [UNSUPPORTED_EDNS_VER]
-	6. Else, if the DNS response has the RCODE "NOERROR" and "OPTION-CODE"
-	present in the response then output [UNKNOWN_OPTION_CODE].
+	4. Else, if the RCODE is "NOERROR", and the response has EDNS version
+	number greater than zero and contains an "OPTION-CODE",  then output
+	*[UNSUPPORTED_EDNS_VER]* and *[UNKNOWN_OPTION_CODE]*. 
+	5. Else, if the RCODE is "NOERROR", and the response has EDNS version
+	number greater than zero,  then output *[UNSUPPORTED_EDNS_VER]* .
+	6. Else, if the RCODE is "NOERROR", and contains an "OPTION-CODE",  then
+	output *[UNKNOWN_OPTION_CODE]*.
 	7. Else, if the DNS response meet the following four criteria,
       	then just go to the next name server (no error):
 		1. If the SOA is not obtained as response in the ANSWER section.
@@ -63,7 +63,6 @@ NO_EDNS_SUPPORT                   | WARNING
 NS_ERROR			  | WARNING     
 UNKNOWN_OPTION_CODE               | WARNING
 UNSUPPORTED_EDNS_VER      	  | WARNING
-UNKNOWN_OPTION_CODE_VERSION       | WARNING
 
 ### Special procedural requirements
 
@@ -84,4 +83,3 @@ https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-paramet
 [NS_ERROR]: #outcomes
 [UNKNOWN_OPTION_CODE]: #outcomes
 [UNSUPPORTED_EDNS_VER]: #outcomes
-[UNKNOWN_OPTION_CODE_VERSION]: #outcomes
--- a/docs/specifications/tests/Nameserver-TP/nameserver14.md
+++ b/docs/specifications/tests/Nameserver-TP/nameserver14.md
@@ -1,0 +1,76 @@
+# nameserver14: Test for unknown version with unknown OPTION-CODE
+
+## test case identifier
+
+**NAMESERVER14** 
+
+### Objective
+
+This test case actually combines the test options in nameserver 10 and
+nameserver 11. 
+
+### Inputs
+
+"Child Zone" - The domain name to be tested.
+
+### Ordered description of steps to be taken to execute the test case
+
+1. Create an SOA query for the *Child Zone* with an OPT record with 
+   EDNS version set to "1" and  with an OPT record with
+   EDNS OPTION-CODE set to anything other than it is already assigned as in the
+[IANA-DNSSYSTEM-PARAMETERS] and no other EDNS options or flags.
+
+2. Obtain the set of name server IP addresses using [Method4] and [Method5]
+   ("Name Server IP").
+
+3. For each name server in *Name Server IP* do:
+	1. Send the SOA query to the name server and collect the response.
+	2. If there is no DNS response, output *[NO_RESPONSE]* and go to
+      next server.
+	3. The DNS response has the RCODE "FORMERR" then output
+      *[NO_EDNS_SUPPORT]*.
+
+4. Else, if the DNS response meet the following four criteria,
+      then just go to the next name server (no error):
+	1. If the SOA is not obtained as response in the ANSWER section.
+	2. It has the RCODE "BADVERS".
+	3. The pseudo-section response has an OPT record with version set to 0.
+	4. The option is not present in the response
+
+5. Else output *[NS_ERROR]*.
+ 
+### Outcome(s)
+
+The outcome of this Test Case is "fail" if there is at least one message
+with the severity level *ERROR* or *CRITICAL*.
+
+The outcome of this Test Case is "warning" if there is at least one message
+with the severity level *WARNING*, but no message with severity level
+*ERROR* or *CRITICAL*.
+
+The outcome of this Test case is "pass" in all other cases.
+
+Message                           | Default severity level (when message is outputed)
+:---------------------------------|:--------------------------------------------------
+NO_RESPONSE                       | WARNING
+NO_EDNS_SUPPORT                   | NOTICE
+NS_ERROR			  | WARNING     
+
+### Special procedural requirements
+
+If either IPv4 or IPv6 transport is disabled, ignore the evaluation of the
+result of any test using this transport protocol and log a message reporting
+the ignored result.
+
+### Intercase dependencies
+
+None.
+
+[RFC 6891]: https://tools.ietf.org/html/rfc6891
+[IANA-DNSSYSTEM-PARAMETERS]:
+https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-11
+[Method4]: ../Methods.md#method-4-delegation-name-server-addresses
+[Method5]: ../Methods.md#method-5-in-zone-addresses-records-of-name-servers
+[NO_RESPONSE]: #outcomes
+[NO_EDNS_SUPPORT]: #outcomes
+[BAD_UNSUPPORTED_VER]: #outcomes
--- a/docs/specifications/tests/Nameserver-TP/nameserver14.md
+++ b/docs/specifications/tests/Nameserver-TP/nameserver14.md
@@ -74,4 +74,5 @@ https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-paramet
 [Method5]: ../Methods.md#method-5-in-zone-addresses-records-of-name-servers
 [NO_RESPONSE]: #outcomes
 [NO_EDNS_SUPPORT]: #outcomes
+[NS_ERROR]: #outcomes
 [EDNS_INCORRECT_OPTION]: #outcomes
--- a/docs/specifications/tests/Nameserver-TP/nameserver14.md
+++ b/docs/specifications/tests/Nameserver-TP/nameserver14.md
@@ -31,14 +31,16 @@ This test case actually combines the test options in test cases
 	3. Else, if the DNS response has the RCODE "FORMERR" then output
       	*[NO_EDNS_SUPPORT]*.
 	4. Else, if the RCODE has "NOERROR", and it has an  OPT record with
-	any version, then output [EDNS_INCORRECT_OPTION]. 
-	5. Else, if the DNS response meet the following four criteria,
+	any version, then output [UNKNOWN_OPTION_CODE]. 
+	5. Else, if the DNS response has the RCODE "NOERROR" then output
+	[UNSUPPORTED_EDNS_VER].
+	6. Else, if the DNS response meet the following four criteria,
       	then just go to the next name server (no error):
 		1. If the SOA is not obtained as response in the ANSWER section.
 		2. It has the RCODE "BADVERS".
 		3. The pseudo-section response has an OPT record with version set to 0.
 		4. The option is not present in the response
-	6. Else output *[NS_ERROR]*.
+	7. Else output *[NS_ERROR]*.
  
 ### Outcome(s)
 
@@ -56,7 +58,8 @@ Message                           | Default severity level (when message is outp
 NO_RESPONSE                       | WARNING
 NO_EDNS_SUPPORT                   | WARNING
 NS_ERROR			  | WARNING     
-EDNS_INCORRECT_OPTION             | WARNING
+UNKNOWN_OPTION_CODE               | WARNING
+UNSUPPORTED_EDNS_VER      	  | WARNING
 
 ### Special procedural requirements
 
@@ -75,4 +78,5 @@ https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-paramet
 [NO_RESPONSE]: #outcomes
 [NO_EDNS_SUPPORT]: #outcomes
 [NS_ERROR]: #outcomes
-[EDNS_INCORRECT_OPTION]: #outcomes
+[UNKNOWN_OPTION_CODE]: #outcomes
+[UNSUPPORTED_EDNS_VER]: #outcomes
--- a/docs/specifications/tests/Nameserver-TP/nameserver14.md
+++ b/docs/specifications/tests/Nameserver-TP/nameserver14.md
@@ -31,7 +31,7 @@ This test case actually combines the test options in test cases
 	3. Else, if the DNS response has the RCODE "FORMERR" then output
       	*[NO_EDNS_SUPPORT]*.
 	4. Else, if the RCODE has "NOERROR", and has an "OPTION-CODE" and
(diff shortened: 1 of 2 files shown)


docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,76 @@
+# nameserver14: Test for unknown version with unknown OPTION-CODE
```
matsduf: "nameserver14" --> "NAMESERVER14"

docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,76 @@
...
+
+## test case identifier
+
+**NAMESERVER14** 
+
+### Objective
+
+This test case actually combines the test options in nameserver 10 and
+nameserver 11.
```
matsduf: "in nameserver 10 and nameserver 11" --> "in test cases NAMESERVER10 and NAMESERVER11".

Shouldn't those be links?

docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,76 @@
...
+### Inputs
+
+"Child Zone" - The domain name to be tested.
+
+### Ordered description of steps to be taken to execute the test case
+
+1. Create an SOA query for the *Child Zone* with an OPT record with 
+   EDNS version set to "1" and  with an OPT record with
+   EDNS OPTION-CODE set to anything other than it is already assigned as in the
+[IANA-DNSSYSTEM-PARAMETERS] and no other EDNS options or flags.
```
matsduf: I have a problem with the wording since it sounds as if more than one OPT record is added. I suggest:
```
1. Create an SOA query for the *Child Zone* with an OPT record with 
   EDNS version set to "1" and  with EDNS OPTION-CODE set to an
   unassigned value (see [IANA-DNSSYSTEM-PARAMETERS]) and
   with no other EDNS options or flags.
```

docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,76 @@
...
+
+2. Obtain the set of name server IP addresses using [Method4] and [Method5]
+   ("Name Server IP").
+
+3. For each name server in *Name Server IP* do:
+	1. Send the SOA query to the name server and collect the response.
+	2. If there is no DNS response, output *[NO_RESPONSE]* and go to
+      next server.
+	3. The DNS response has the RCODE "FORMERR" then output
```
matsduf: "Else, if the DNS response ..."

docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,76 @@
...
+      *[NO_EDNS_SUPPORT]*.
+
+4. Else, if the DNS response meet the following four criteria,
```
matsduf: Indent to be on the same level as the previous.

docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,76 @@
...
+      then just go to the next name server (no error):
+	1. If the SOA is not obtained as response in the ANSWER section.
+	2. It has the RCODE "BADVERS".
+	3. The pseudo-section response has an OPT record with version set to 0.
+	4. The option is not present in the response
```
matsduf: Indent these four.

docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,76 @@
...
+The outcome of this Test Case is "warning" if there is at least one message
+with the severity level *WARNING*, but no message with severity level
+*ERROR* or *CRITICAL*.
+
+The outcome of this Test case is "pass" in all other cases.
+
+Message                           | Default severity level (when message is outputed)
+:---------------------------------|:--------------------------------------------------
+NO_RESPONSE                       | WARNING
+NO_EDNS_SUPPORT                   | NOTICE
```
matsduf: For all NAMESERVER test cases we must have the same level for the same message. We have to set it to WARNING.

docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,76 @@
...
+
+If either IPv4 or IPv6 transport is disabled, ignore the evaluation of the
+result of any test using this transport protocol and log a message reporting
+the ignored result.
+
+### Intercase dependencies
+
+None.
+
+[RFC 6891]: https://tools.ietf.org/html/rfc6891
```
matsduf: Not used.

docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,76 @@
...
+[IANA-DNSSYSTEM-PARAMETERS]:
+https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-11
+[Method4]: ../Methods.md#method-4-delegation-name-server-addresses
+[Method5]: ../Methods.md#method-5-in-zone-addresses-records-of-name-servers
+[NO_RESPONSE]: #outcomes
+[NO_EDNS_SUPPORT]: #outcomes
+[BAD_UNSUPPORTED_VER]: #outcomes
```
matsduf: Not used.

sandoche2k:
@matsduf Modified as per your review.

docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,77 @@
+# NAMESERVER14: Test for unknown version with unknown OPTION-CODE
+
+## Test case identifier
+
+**NAMESERVER14** 
+
+### Objective
+
+This test case actually combines the test options in test cases
+[NAMESERVER10](../nameserver10.md) and [NAMESERVER11](../nameserver11.md).
```
matsduf: What does this Test Case add? Isn't everything covered by the other two test cases?

docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,77 @@
...
+
+None.
+
+[IANA-DNSSYSTEM-PARAMETERS]:
+https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-11
+[Method4]: ../Methods.md#method-4-delegation-name-server-addresses
+[Method5]: ../Methods.md#method-5-in-zone-addresses-records-of-name-servers
+[NO_RESPONSE]: #outcomes
+[NO_EDNS_SUPPORT]: #outcomes
+[EDNS_INCORRECT_OPTION]: #outcomes
```
matsduf:
```
[NS_ERROR]: #outcomes
```

docs/specifications/tests/Nameserver-TP/nameserver14.md
sandoche2k: I did have that question. But in the EDNS compliance tests, they are doing a separate test by combining both the options (unknown option code and version). 

If @vlevigneron also thinks that this test is not needed, we can remove it.

matsduf: @sandoche2k, does the test case specification really catch that?

sandoche2k: @matsduf Updated. Pls. verify.

sandoche2k:
In Namserver 10, if there is an ANSWER, then " NO_EDNS_SUPPORT"

Good catch.. i have updated

docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,87 @@
...
+3. For each name server in *Name Server IP* do:
+
+	1. Send the SOA query to the name server and collect the response.
+	2. If there is no DNS response, output *[NO_RESPONSE]* and go to
+      	next server.
+	3. Else, if the DNS response has the RCODE "FORMERR" then output
+      	*[NO_EDNS_SUPPORT]*.
+	4. Else, if the RCODE has "NOERROR", and has an "OPTION-CODE" and
+	"Version Number" present in the response,  then output
+	[UNKNOWN_OPTION_CODE_VERSION].
```
matsduf: I suggest:

4. Else, if the RCODE has "NOERROR", and has an OPT record with "OPTION-CODE", then output
    [UNKNOWN_OPTION_CODE], and if the OPT version is greater than zero, then also output
    [UNSUPPORTED_EDNS_VER].

docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,87 @@
...
+	1. Send the SOA query to the name server and collect the response.
+	2. If there is no DNS response, output *[NO_RESPONSE]* and go to
+      	next server.
+	3. Else, if the DNS response has the RCODE "FORMERR" then output
+      	*[NO_EDNS_SUPPORT]*.
+	4. Else, if the RCODE has "NOERROR", and has an "OPTION-CODE" and
+	"Version Number" present in the response,  then output
+	[UNKNOWN_OPTION_CODE_VERSION]. 
+	5. Else, if the DNS response has the RCODE "NOERROR" and "Version
+	number" present in the response then output [UNSUPPORTED_EDNS_VER]
```
matsduf: I suggest:

5. Else, if the DNS response has the RCODE "NOERROR" with an OPT record with version greater
    than zero output [UNSUPPORTED_EDNS_VER].

docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,87 @@
...
+	6. Else, if the DNS response has the RCODE "NOERROR" and "OPTION-CODE"
+	present in the response then output [UNKNOWN_OPTION_CODE].
```
matsduf: I guess this is not needed. The rest of the errors go to NS_ERROR.

BTW, it is not possible to have an OPT record without ENDS version is set.

docs/specifications/tests/Nameserver-TP/nameserver14.md
sandoche2k: updated

docs/specifications/tests/Nameserver-TP/nameserver14.md
sandoche2k: @matsduf i am following this : 

EDNS - Unknown Version with Unknown Option
dig +norec +edns=100 +noednsneg +ednsopt=100 soa zone @server
expect: BADVERS
expect: OPT record with version set to 0
expect: not to see SOA
expect: that the option will not be present in response
See RFC6891

matsduf: The problem is not to define good behavior. The problem is to label/classify everything else.

sandoche2k: I do not understand your statement. EDNS compliance test follows all these tests : http://users.isc.org/~marka/summary.html

docs/specifications/tests/Nameserver-TP/nameserver14.md
matsduf: @sandoche2k, you did not like my proposal?

```
	4. Else, if the RCODE has "NOERROR", and has an "OPTION-CODE" and
            (...)
```
If the response has an OPTION-CODE set it means that an OPT record is present. For the clarity I think we should say that.
```
        (...)
	"Version Number" present in the response,  then output
        (...)
```
If there is an OPT record there is, by definition, the EDNS version is always set. That is nothing to test for.
```
        (...)
	[UNKNOWN_OPTION_CODE_VERSION]. 
        (...)
```
What would this combination message say? What not use the two different messages instead?

docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,87 @@
...
+      	next server.
+	3. Else, if the DNS response has the RCODE "FORMERR" then output
+      	*[NO_EDNS_SUPPORT]*.
+	4. Else, if the RCODE has "NOERROR", and has an "OPTION-CODE" and
+	"Version Number" present in the response,  then output
+	[UNKNOWN_OPTION_CODE_VERSION]. 
+	5. Else, if the DNS response has the RCODE "NOERROR" with an OPT record
+	with version greater than zero, then output [UNSUPPORTED_EDNS_VER]
+	6. Else, if the DNS response has the RCODE "NOERROR" and "OPTION-CODE"
+	present in the response then output [UNKNOWN_OPTION_CODE].
```
matsduf: We will never come to "6" since "4" will catch all cases that "6" would catch.

docs/specifications/tests/Nameserver-TP/nameserver14.md
matsduf: @sandoche2k, the ISC link specifies what to expect when everything is compliant, and then the following step is fine,
```
   7. Else, if the DNS response meet the following four criteria, then just go to the 
      next name server (no error):
      1. If the SOA is not obtained as response in the ANSWER section.
      2. It has the RCODE "BADVERS".
      3. The pseudo-section response has an OPT record with version set to 0.
      4. The option is not present in the response
```
We have identified two clear "error" cases:
```
   2. If there is no DNS response, output *[NO_RESPONSE]* and go to
      next server.
   3. Else, if the DNS response has the RCODE "FORMERR" then output
      *[NO_EDNS_SUPPORT]*.
```
The problem is what we should output in other cases. We have responses with NOERROR with and without OPT record. We have responses with OPT record with and without OPTION-CODE. And those responses with OPT record can have EDNS version 0 or other version. (An OPT record always has an EDNS version.)

docs/specifications/tests/Nameserver-TP/nameserver14.md
sandoche2k: 4 becomes true only if both option-code and version number are present. 
5 and 6 tests them individually.

matsduf: @sandoche2k, the test `"Version Number" present in the response` is the same thing as saying that OPT record is present. If there is an OPTION-CODE there is an OPT record. The following two condition statements are indentical:
1. The response has an "OPTION-CODE" and "Version Number" is present in the response.
2. The response has an "OPTION-CODE".

sandoche2k: Maybe you are right. But, i understand it differently from reading the RFC 6891: 

An OPT RR has a fixed part and a variable set of options expressed as   {attribute, value} pairs. 

The version number is part of the OPT record TTL field (which is the fixed part fo the OPT RR) as explained in 6.1.3 section of the RFC 6891

The OPTION-CODE is part of the variable part as explained in 6.1.2 section of the RFC 6891

matsduf: I think it is quite clear from the RFC. The EDNS version sits in the TTL field, which is always there, or else we have a broken DNS  packet. The EDNS version is always defined if the OPT record is there. What we can test is the version value.

From 6.1.3 of https://tools.ietf.org/html/rfc6891
```
The extended RCODE and flags, which OPT stores in the RR Time to Live
   (TTL) field, are structured as follows:

                  +0 (MSB)                            +1 (LSB)
       +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
    0: |         EXTENDED-RCODE        |            VERSION            |
       +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
    2: | DO|                           Z                               |
       +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+

(...)

   VERSION
      Indicates the implementation level of the setter.  Full
      conformance with this specification is indicated by version '0'.
(...)
```

vlevigneron: You are both right but I am not sure you both answer the same question...

Perhaps that with a different approach it could be less confusing:

1/ Please do not use `"Version Number" present in the response`, because it is always present if you have an OPT record (you should say `Version number is greater than 0` as used elesewhere in the text).
2/ Use a matrix for step 4,5,6

sandoche2k:
@vlevigneron good catch for (1). I have updated. 
@matsduf i am also saying the same. But do you agree that the version and the OPTION-CODE are separate values?

sandoche2k:
Thanks @matsduf. Is the recent update Ok with you?

docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,87 @@
...
+3. For each name server in *Name Server IP* do:
+
+	1. Send the SOA query to the name server and collect the response.
+	2. If there is no DNS response, output *[NO_RESPONSE]* and go to
+      	next server.
+	3. Else, if the DNS response has the RCODE "FORMERR" then output
+      	*[NO_EDNS_SUPPORT]*.
+	4. Else, if the RCODE has "NOERROR", and has an "OPTION-CODE" and
+	"Version Number" greater than zero,  then output
+	[UNKNOWN_OPTION_CODE_VERSION].
```
matsduf: I still think that this test should output two messages (UNSUPPORTED_EDNS_VER and UNKNOWN_OPTION_CODE) instead of a combination message.

I suggest the follwoing text:
```
	4. Else, if the RCODE is "NOERROR", and the response has EDNS version greater than 
            zero and contains an "OPTION-CODE",  then output *[UNSUPPORTED_EDNS_VER]* 
            and *[UNKNOWN_OPTION_CODE]*.
```

docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,87 @@
...
+	1. Send the SOA query to the name server and collect the response.
+	2. If there is no DNS response, output *[NO_RESPONSE]* and go to
+      	next server.
+	3. Else, if the DNS response has the RCODE "FORMERR" then output
+      	*[NO_EDNS_SUPPORT]*.
+	4. Else, if the RCODE has "NOERROR", and has an "OPTION-CODE" and
+	"Version Number" greater than zero,  then output
+	[UNKNOWN_OPTION_CODE_VERSION]. 
+	5. Else, if the DNS response has the RCODE "NOERROR" with an OPT record
+	with version greater than zero, then output [UNSUPPORTED_EDNS_VER]
```
matsduf: I suggest
```
	5. Else, if the RCODE is "NOERROR", and the response has EDNS version greater than 
            zero,  then output *[UNSUPPORTED_EDNS_VER]* .
```

docs/specifications/tests/Nameserver-TP/nameserver14.md
```diff
@@ -0,0 +1,87 @@
...
+	6. Else, if the DNS response has the RCODE "NOERROR" and "OPTION-CODE"
+	present in the response then output [UNKNOWN_OPTION_CODE].
```
matsduf: I suggest:
```
	4. Else, if the RCODE is "NOERROR", and contains an "OPTION-CODE",  then output 
            *[UNKNOWN_OPTION_CODE]*.
```

sandoche2k:
@matsduf updated

Issue #657, opened 2018-12-14
Report.md cannot be correctly generated


matsduf:
The document docs/specifications/tests/[Report.md] is created from the directory docs/specifications/[tests] with the following command:

```
make map
```

It cannot be correctly created when the Test Case specification is in the updated format.

Expected extract:

|Req|Requirement desription|Level     |Test Case ID|Test Description|
|:--|:---------------------|:---------|:-----------|:---------------|
|R08|The child domain has atleast one working name server|[  Basic-TP](Basic-TP/README.md)|[BASIC02](./Basic-TP/basic02.md)|  The domain must have at least one working name server|

Actual extract:

|Req|Requirement desription|Level     |Test Case ID|Test Description|
|:--|:---------------------|:---------|:-----------|:---------------|
|R08|The child domain has atleast one working name server|[  Basic-TP](Basic-TP/README.md)|[missing](./Basic-TP/basic02.md)|          missing|

No error message is generated.

[Report.md]: https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/Report.md
[tests]: https://github.com/zonemaster/zonemaster/tree/master/docs/specifications/tests

sandoche2k:
@matsduf close?

sandoche2k:
@matsduf any updates?

matsduf:
The document is removed by #881.

PR #659, opened 2018-12-19
Update maprequirement issue 657

matsduf:
The script maprequirement.pl is used to generat the file docs/specifications/tests/Report.md

That Report depends on a number of files:
* docs/requirements/TestRequirements.md
* The test plan readme files, e.g. docs/specifications/tests/Address-TP/README.md
* The test case specification files, e.g. docs/specifications/tests/Address-TP/address01.md

The script depends on those file having a specific formatting or else the Report will be incomplete. The update here, after doing some cleanup and adding more verification, is to adopt to the new format of the first line in the test case description.

This resolves issue #657.


sandoche2k:
@matsduf pls. merge

PR #664, opened 2019-01-07
Add style guide entry regarding case-insensitive regexes

mattias-p:
The issue with case-insensitive regexes came up here: https://github.com/zonemaster/zonemaster-backend/pull/455#commitcomment-31854097


sandoche2k:
@mattias-p merge for 2018.2?

Issue #763, opened 2019-05-13
Update DNSESEC05 on legal values


matsduf:
When [draft-ietf-dnsop-algorithm-update] has been accepted as RFC then [DNSSEC05] specification must be updated.

Also see https://indico.dns-oarc.net/event/31/contributions/688/attachments/650/1067/matthijs-mekking-oarc30-unsupported-algorithms.pdf

[draft-ietf-dnsop-algorithm-update]: https://datatracker.ietf.org/doc/draft-ietf-dnsop-algorithm-update/?include_text=1

[DNSSEC05]: https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/DNSSEC-TP/dnssec05.md

Solution in #777.

matsduf:
The created PR must not be merged before the RFC is ready.

pawal:
I don't see any reason at all to update DNSSEC05, the specification already refers to the IANA registry of DNSSEC Algorithm Numbers - and this registry is being affected by the acceptance of the mentioned draft.

matsduf:
@pawal, if you take a look at the specification you will see why it must be updated.

https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/DNSSEC-TP/dnssec05.md

sandoche2k:
Should we follow the RFC or the IANA registry? If there is a new RFC the IANA registry will be updated isn't it?

matsduf:
Resolved by #777.

PR #777, opened 2019-06-12
Updated DNSSEC05

matsduf:
* Solves issue #763. 
* Meet specification in new RFC 8624 with more strict requirements.
* Decreased wordiness.
* Increased default level for several messages.
* Message ALGORITHM_DELETE_DS has been removed (algorithm handled
  as ALGORITHM_NOT_ZONE_SIGN).
* Message ALGORITHM_INDIRECT_KEY has been removed (algorithm handled
  as ALGORITHM_RESERVED).
* Some editorial updates.

When this PR has been merged, an issue should be created in Zonemaster-Engine.

### DIFF
--- a/docs/specifications/tests/DNSSEC-TP/dnssec05.md
+++ b/docs/specifications/tests/DNSSEC-TP/dnssec05.md
@@ -5,20 +5,24 @@
 
 ## Objective
 
-A domain name (zone) should only use DNSKEY algorithms that are specified 
-by IANA to be used for a DNSKEY. A public domain name (zone) should not
-use private algorithms. The IANA registry of [DNSSEC Algorithm Numbers]
-specifies which algorithms to use.
+A domain name (zone) should only use DNSKEY algorithms that are specified
+by [RFC 8624], section 3.1 and the [IANA registry] of *DNSSEC Algorithm
+Numbers* to be used for DNSSEC signing. A public domain name (zone) should not use
+private algorithms.
 
-The DNSKEY record is defined in [RFC 4034, section 2].
+If [RFC 8624] and [IANA registry] disagree on the same algorithm, the
+RFC takes precedence until the registry has a been updated with a
+reference to the RFC.
 
 ## Inputs
 
 * The domain name to be tested ("Child Zone").
+* The status of all algorithms from [RFC 8624] and [IANA registry]
+  ("Algorithm Status").
 
 ## Ordered description of steps to be taken to execute the test case
 
-1. Create a DNSKEY query with DO flag set for the apex of the 
+1. Create a DNSKEY query with DO flag set for the apex of the
    *Child Zone*.
 
 2. Retrieve all name server IP addresses for the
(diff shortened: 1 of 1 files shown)


sandoche2k:
@sandoche2k @mattias-p to do the review.

docs/specifications/tests/DNSSEC-TP/dnssec05.md
```diff
@@ -5,20 +5,24 @@
...
 ## Objective
 
-A domain name (zone) should only use DNSKEY algorithms that are specified 
-by IANA to be used for a DNSKEY. A public domain name (zone) should not
-use private algorithms. The IANA registry of [DNSSEC Algorithm Numbers]
-specifies which algorithms to use.
+A domain name (zone) should only use DNSKEY algorithms that are specified
+by [RFC 8624], section 3.1 and the [IANA registry] of *DNSSEC Algorithm
+Numbers* to be used for DNSSEC signing. A public domain name (zone) should not use
+private algorithms.
```
sandoche2k: I am not sure is it a good idea to have both the RFC and the registry as reference. The registry will be updated, whereas the RFC will not be..

docs/specifications/tests/DNSSEC-TP/dnssec05.md
```diff
@@ -27,29 +31,25 @@ The DNSKEY record is defined in [RFC 4034, section 2].
...
+      2. If the algorithm is reserved (algorithm
+         4, 9, 11, 123-252 or 255), output *[ALGORITHM_RESERVED]*.
+      3. If the algorithm is unassigned (algorithm
+         17-122), output *[ALGORITHM_UNASSIGNED]*.
+      4. If the algorithm is private algorithm
+         (algorithm 253-254), output *[ALGORITHM_PRIVATE]*.
+      5. If the algorithm is not meant for zone signing (algorithm
+         0-2), output *[ALGORITHM_NOT_ZONE_SIGN]*.
+      6. If the algorithm is not rekommended for zone signing (algorithm
+         5, 7 or 10), output *[ALGORITHM_NOT_RECOMMENDED]*.
```
sandoche2k: Why did you remove INDIRECT_KEY?

docs/specifications/tests/DNSSEC-TP/dnssec05.md
matsduf: I agree that it is complex, but I think we have to meet the stricter specification in the RFC at the same time as we should keep a reference to the registry. There algorithms that the registry covers that are not mentioned in the RFC.

When the registry is updated, there will be a simple update if they just adapt to the RFC (which I think they will) or else we have to decide how we should balance.

Independent of what we do, we have to follow the change of RFC and the update of the registry.

docs/specifications/tests/DNSSEC-TP/dnssec05.md
matsduf: I did not remove it. It is treated as "reserved" instead. It is quite unclear what it really is. I guess it is better to move it back.

PR #779, opened 2019-06-25
Updated DNSSEC01

matsduf:
* Solves issue zonemaster#762.
* Meet specification in new RFC 8624 with more strict requirements.
* Rewritten in new format, e.g.
  * Steps are specified in more details.
  * Messages are explicitly specified incl default severity level.

When this PR has been merged, an issue should be created in Zonemaster-Engine.

### DIFF
--- a/docs/specifications/tests/DNSSEC-TP/dnssec01.md
+++ b/docs/specifications/tests/DNSSEC-TP/dnssec01.md
@@ -1,41 +1,120 @@
-## DNSSEC01: Legal values for the DS hash digest algorithm
+# DNSSEC01: Legal values for the DS hash digest algorithm
 
-### Test case identifier
-**DNSSEC01** Legal values for the DS hash digest algorithm
+## Test case identifier
+**DNSSEC01**
 
-### Objective
+## Objective
 
-The allowed Digest Algorithms in a DS record published by the parent are
-published by IANA in [Delegation Signer (DS) Resource Record (RR) Type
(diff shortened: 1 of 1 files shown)


docs/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -1,41 +1,120 @@
...
+  (only if *Test Type* is undelegated).  
 
-1. Retrieve the DS RR set from the parent zone.
-2. From the DS RRs, extract the Digest type values.
-3. If any value from the Digest type is not included in the IANA table
-   (currently values 1 to 4 are valid), this test case fails.
+## Ordered description of steps to be taken to execute the test case
 
-### Outcome(s)
+1. If the *Test Type* is "undelegated, then:
```
sandoche2k: If we are going this path, then shouldn't we add an undelegated section for all test cases?

matsduf: Yes, in many cases, or maybe most cases. When we start using the [new methods] it can often be hidden from the explicit test case text.

See updated [Basic01] (in develop branch) which was approved 5 days ago in PR #642, which heavily refers to undelegated test.

I see no other way to handle it.

[new methods]: https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/MethodsNT.md

[Basic01]: https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/Basic-TP/basic01.md

PR #597, opened 2018-03-23
Update connectivity03

matsduf:
* Added support for ASN lookup against RIPE.
* Rewriting the steps.
* Adding the messages and default level.

This PR is the first step to solve issue dotse/zonemaster-engine#273. It also incorporates the messages into the Test Case.

PR #781, opened 2019-06-26
Update Connectivity03 (v2)

matsduf:
* Adding missing references
* Transformed into new format with more explicit steps
* Messages explicitly defined in specification
* New messages
* Added support for RIPE Ris Whois in addition to Cymru ASN look-up

This PR is the first step to solve issue zonemaster/zonemaster-engine#273.

This PR replaces PR #597.

When this is merged, a new issue must be created in Zonemaster-Engine.

### DIFF
--- a/docs/specifications/tests/Connectivity-TP/connectivity03.md
+++ b/docs/specifications/tests/Connectivity-TP/connectivity03.md
@@ -7,14 +7,14 @@
 ## Objective
 
 The objective in this test is to verify that all IP addresses of the domain's
-authoritative name servers are not announced from the same ASN (autonomous system 
+authoritative name servers are announced from different ASNs (autonomous system 
 number). See [RFC 1930] and [Wikipedia] for an explanation of AS (autonomous 
 system).
 
 This test is done separately on IPv4 and IPv6, and both must match the criterion.
 
-[RFC 2182], section 3.1 clearly specifies that distinct authoritative name 
-servers for a child domain should be placed in different topological and 
+[RFC 2182][RFC 2182#3.1], section 3.1, clearly specifies that distinct authoritative 
+name servers for a child domain should be placed in different topological and 
 geographical locations. The objective is to minimise the likelihood of a single 
 failure disabling all of them. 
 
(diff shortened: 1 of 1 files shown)


docs/specifications/tests/Connectivity-TP/connectivity03.md
```diff
@@ -1,58 +1,204 @@
...
-   fails.
 
-### Outcome(s)
+## Ordered description of steps to be taken to execute the test case
 
-If there is a AS which is different from the other retrieved AS, then the
-test succeeds.
+1. Obtain the total set of IP addresses of the name servers for the 
+   *Child Zone* using [Method4] and [Method5] and do:
+   1. Create a possibly empty set of the IPv4 addresses ("NS IPv4").
```
sandoche2k: Not clear why do you want to create an empty set of IP addresses

matsduf: It is more like saying that there might not be any IPv4 addresses. And if the set is not kind of created (but empty) then some could complain about the logic in the step below where there is a reference to the set "NS IPv4".

Do you think it makes it harder to follow or understand the steps?

mattias-p: Have you considered something like this?

> Split the total set into two subsets, one for IPv4 addresses ("NS IPv4") and one for IPv6 addresses ("NS IPv6"). (The subsets may be empty.)

matsduf:
@mattias-p and @sandoche2k, I have done some editorial updates after review. Please re-review.

matsduf:
@mattias-p, I think your comments have been handled. Please mark them as "resolved" if so. Please re-reveiw.

@vlevigneron, please review.

matsduf:
@vlevigneron, by next release do you mean v2019.2 or v2019.3?

vlevigneron:
@matsduf I mean v2019.2.

matsduf:
For implentation of the changes created by this PR issue zonemaster/zonemaster-engine#592 has been created.

Issue #528, opened 2017-11-16
Test that a signed zone include a DNSKEY for each algorithm present in the zone's DS RRset


matsduf:
From RFC 6840:

>       The DS RRset and DNSKEY RRset are used to signal which algorithms
>       are used to sign a zone.  The presence of an algorithm in either a
>       zone's DS or DNSKEY RRset signals that that algorithm is used to
>       sign the entire zone.
> 
>       A signed zone MUST include a DNSKEY for each algorithm present in
>       the zone's DS RRset and expected trust anchors for the zone.  The
>       zone MUST also be signed with each algorithm (though not each key)
>       present in the DNSKEY RRset.  It is possible to add algorithms at
>       the DNSKEY that aren't in the DS record, but not vice versa.  If
>       more than one key of the same algorithm is in the DNSKEY RRset, it
>       is sufficient to sign each RRset with any subset of these DNSKEYs.
>       It is acceptable to sign some RRsets with one subset of keys (or
>       key) and other RRsets with a different subset, so long as at least
>       one DNSKEY of each algorithm is used to sign each RRset.
>       Likewise, if there are DS records for multiple keys of the same
>       algorithm, any subset of those may appear in the DNSKEY RRset.

Create a new Test Case that requires that for each algorithm found in the DS RRset for the zone, there is also a DNSKEY with the same algorithm.

pawal:
Also see #344

matsduf:
Replaces #527

sandoche2k:
@matsduf @pawal 
As I understand: https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/DNSSEC-TP/dnssec02.md matches all DS RR algorithm with DNSKEY RR from the child

If we add the same to https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/DNSSEC-TP/dnssec07.md

then this issue could be resolved. Isn't it?

matsduf:
PR #782 will solve this issue.

matsduf:
Resolved by #782.

Issue #344, opened 2015-02-18
Several algorithms in DS or DNSKEY RRsets, Zonemaster should warn


pawal:
_From @bortzmeyer on February 18, 2015 16:15_
# Several algorithms in DS or DNSKEY RRsets, and consequences

The problem was noticed on cepn.asso.fr (the configuration changed
since). Some resolvers or DNS checkers complained about its setup,
while other accepted it. What was right?
## Setup

The zone had the following DNSKEY RRset:

```
cepn.asso.fr.           7808 IN DNSKEY 257 3 5 (
                                AwEAAaBtXBNAyFHVvRBB4K9z79+1YRXkUDyycyCzPRpm
                                Xi9lhB0Eg5vM3XlaS6OuN0dnFHItpZFNIDBDrPsN1OCf
                                1ULKWpD3KDl1mE7zRK2W0HXeu4WOoFpUcC/1h06W26DT
                                CkisntU9L8JfPi9osmI+CuzWZhdmyZt+hPvMpjmDthyh
                                MZpb//kNv7+TUeczCo4MExHxjHHIVH0vRmhfyo/J1KBe
                                6eS3G5lDbJEEFUdxuLyGQLaG2f6wlQxoHGnzvM+V/Mj8
                                yGHae//7Z5rMCdaiLJy03u5+l2WVVy954dsrFC6mkB5s
                                M4n8nvbo1d5ap7cI76dJi9X0IUJQohZk5b5eef0=
                                ) ; KSK; alg = RSASHA1; key id = 36778
cepn.asso.fr.           7808 IN DNSKEY 256 3 5 (
                                AwEAAc6AqnBoi+hfxMqtb0eokyqWT46Os5N6ZYoFm8Gb
                                t90EF3hTpuwDClEsulKSckhr4zFTDj3SvHc9krzeQEl5
                                UNCqmmZeMo/wsxKHTzIVU75fPrs1zOuM9m9zRNV4q9eG
                                Y0+I2h4D7E/WlPE7n57E0lmPOxK9g46xE8p9eX3bWVVK
                                FSm60VvginZfTzN3Zgt+peecrboEZnSzWvDVcHY2dq+o
                                w0UEekI1+nfwcIgEOn0Wh8B5Gx3pG5XkV3QvHVN514FH
                                eJLdsk0iFPHv1Xc0rLYWssFVS9s7Z8u0tEju6LshGaPQ
[...]

PR #782, opened 2019-06-28
Updated DNSSEC02

matsduf:
* Transformed into new format.
* Explicit messages in specification.
* Changed, added and removed messages.
* Logic updated:
  * All DS must match DNSKEY.
  * DS is validated against DNSKEY.
  * DNSKEY of DS must have signed DNSKEY RRset.
  * RRSIG is validated against DNSKEY.
  * DNSKEY is check for Zone key flag and SEP flag, respectively.

This PR will solve #528 and #344.

When this PR has been merged an issue will be created in Zonemaster-Engine.

### DIFF
--- a/docs/specifications/tests/DNSSEC-TP/dnssec02.md
+++ b/docs/specifications/tests/DNSSEC-TP/dnssec02.md
@@ -1,69 +1,145 @@
-## DNSSEC02: DS must match a DNSKEY in the designated zone
+# DNSSEC02: DS must match a valid DNSKEY in the child zone
 
-### Test case identifier
-**DNSSEC02** DS must match a DNSKEY in the designated zone
+## Test case identifier
+**DNSSEC02**
 
-### Objective
+## Objective
 
 DNS delegations from a parent to a child are secured with DNSSEC by
 publishing one or several Delgation Signer (DS) records in the parent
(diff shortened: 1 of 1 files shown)


docs/specifications/tests/DNSSEC-TP/dnssec02.md
```diff
@@ -1,69 +1,159 @@
...
+      1. Send the DS query over UDP and collect the response.
+      2. If there is no DNS response, then output *[NO_RESPONSE]*.
+      3. Else, if the RCODE is not NOERROR, then output
+         *[UNEXPECTED_RESPONSE_DS]*.
+      4. Else, go to next IP address if there is no DS in the
+         response.
+      5. Else, extract all DS records from the DNS response and
+         add them to *DS Records*.
+
+4. If *DS Records* is empty, exit this test case.
```
sandoche2k: Shouldn't the test case exit with an error message if DS records are empty

matsduf: No, because it is not an error not to have a DS. The purpose of this test case is to verify that DS, if any, matches a valid DNSKEY.

matsduf:

matsduf:
Issue zonemaster/zonemaster-engine#619 has been created for the update.

Issue #494, opened 2016-12-13
Create TC of DNSSEC key size


matsduf:
Create a TC that checks the DNSSEC key size matches the standard requirements.

See issue https://github.com/dotse/zonemaster-engine/issues/227

sandoche2k:
@matsduf Can you have a look at this issue

matsduf:
PR #783 solves this issue.

matsduf:
Resolved by #783

PR #783, opened 2019-06-28
Create DNSSEC14

matsduf:
Test case that verifies the key size. Solves issues #494 and
zonemaster/zonemaster-engine#227.

### DIFF
--- a/docs/specifications/tests/DNSSEC-TP/README.md
+++ b/docs/specifications/tests/DNSSEC-TP/README.md
@@ -21,6 +21,7 @@ This document uses the terminology defined in the
 |R76|Zone contains NSEC or NSEC3 records                          |[DNSSEC10](dnssec10.md)|
 |R79|If DS at parent, child zone must be signed                   |[DNSSEC11](dnssec11.md)|
 |R84|Test for DNSSEC Algorithm Completeness (DS->DNSKEY->RRSIG)   |[DNSSEC12](dnssec12.md)|
+|???|Check for valid DNSKEY key size                              |[DNSSEC14](dnssec14.md)|
 
 ## Default DNS query flags for all DNSSEC tests
 
--- a/docs/specifications/tests/DNSSEC-TP/dnssec14.md
+++ b/docs/specifications/tests/DNSSEC-TP/dnssec14.md
@@ -1,0 +1,116 @@
+# DNSSEC14: Check for valid DNSKEY key size
+
+## Test case identifier
+**DNSSEC14**
+
+## Objective
+
+Depending on algorithm, some DNSKEY has different minimum and maximum
+key sizes. This test case will validate the keys size. Algorithms
+that are deprecated, else are not suitable for DNSKEY ([RFC 8624] and 
+[IANA registry]) or are have fixed size are just ignored. See test case 
+[DNSSEC05] for test of algorithm.
+
+The algorithms in the table below are the ones tested by this test case.
+The key sizes in the table below is in bits.
+
+Algorithm | Min size  | Max size | Reference
+:---------|:----------|:---------|:----------------
+5         | 512 ?     | 4096 ?   | [RFC 3110]
+7         | 512 ?     | 4096 ?   | [RFC 5155]
+8         | 512       | 4096     | [RFC 5702]
+10        | 1024      | 4096     | [RFC 5702]
+
+
+## Inputs
+
+* "Child Zone" - The domain name to be tested. 
+* "Key Size Table" - The table above. 
+
+## Ordered description of steps to be taken to execute the test case
+
+1. Create a DNSKEY query with DO flag set for the apex of the
+   *Child Zone*.
+
+2. Retrieve all name server IP addresses for the
+   *Child Zone* using [Method4] and [Method5] ("NS IP").
+
+3. Create an empty set "DNSKEY RRs".
+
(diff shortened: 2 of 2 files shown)


matsduf:
@stirnim, you created issue zonemaster/zonemaster-engine#227 long time ago. Now we have a proposal of a test case (test) that meets your issue. Please review. I have some question marks. Maybe you know more.

stirnim:
Thank you @matsduf for creating this test case. I guess my original intend was that zonemaster shows an error output for RSA key sizes outside the allowed/standardized boundaries. I think this test case succeeds doing this. As for the question marks, not sure why they are there. For example, RFC 3110 states min RSA key size is 512 and max 4096 bits.

However, I think limiting the key size test to just show an **error** if outside the standardized boundaries is a shortcoming. It think it would be more useful, if it also **warns** on small key sizes. As there does not seem to be an RFC with such recommendations, I suggest to base the recommendation on [NIST SP 800-57 Part 1](https://csrc.nist.gov/publications/detail/sp/800-57-part-1/revised/archive/2007-03-01), [Part 3](https://csrc.nist.gov/publications/detail/sp/800-57-part-3/rev-1/final).

Specifically, these two documents make the following recommendation for RSA (taking DNSSEC into account):
KSK min size: 2048 bits (NIST SP 800-57 Part 1, Page 66)
ZSK min size: >1024 bits (NIST SP 800-57 Part 3, Page 51)

matsduf:
@stirnim, thank you for the update. It was very helpful and I like your suggestion. I followed the links that you provided, but I cannot see that there is an exception for ZSK any more (after year 2015). I also followed part 1 to revision 4. Please review the updated specification.

stirnim:
> I followed the links that you provided, but I cannot see that there is an exception for ZSK any more (after year 2015).

Rereading section "8.1.4 Special Considerations for Key Sizes" Page 51 ([NIST SP 800-57 Part 3, Rev. 1](https://csrc.nist.gov/publications/detail/sp/800-57-part-3/rev-1/final)), I agree with you. After 2015, the minimal ZSK key size is not >1024-bits but >= 2048-bits from NIST SP 800-57 Part 1. Thanks for spotting that.

As for your commit, there is a typo in the word "lenght" three times where it should be "length" 

> It is also recommended that an RSA based algorithm has a key lenght 
> of at least 2048 bit as stated in [Recommendation for key Management, 
> part 1, revision 4], table 2 on page 53 in section 5.6.1 and table 4 
> on page 55 in section 5.6.2 (also link from [NIST Publications 
> SP 800-57 Part 1 Rev. 4]

I do not see why the addition of _"and table 4 on page 55 in section 5.6.2 (also link from [NIST Publications SP 800-57 Part 1 Rev. 4]"_ is needed. Can we omit this? Thank you.

matsduf:
@stirnim, thank you for reviewing and finding the typos. I removed the double references to the NIST document and kept the most informative. I do not understand why you want to remove the reference to table 4. I think that is needed to support the requirement of 2048 bit key length. Please take a look again.

stirnim:
@matsduf, I agree with you. Table 2 focus on comparison on algorithms and key sizes whereas the focus of table 4 is on security strength with a time frame.

docs/specifications/tests/DNSSEC-TP/README.md
```diff
@@ -21,6 +21,7 @@ This document uses the terminology defined in the
 |R76|Zone contains NSEC or NSEC3 records                          |[DNSSEC10](dnssec10.md)|
 |R79|If DS at parent, child zone must be signed                   |[DNSSEC11](dnssec11.md)|
 |R84|Test for DNSSEC Algorithm Completeness (DS->DNSKEY->RRSIG)   |[DNSSEC12](dnssec12.md)|
+|???|Check for valid RSA DNSKEY key size                          |[DNSSEC14](dnssec14.md)|
```
sandoche2k: Shouldn't the test case be DNSSEC13? Why added after Test requirement 84?

docs/specifications/tests/DNSSEC-TP/dnssec14.md
```diff
@@ -0,0 +1,130 @@
...
+   1. If the algorithm of the DNSKEY is not listed in *Key Size 
+      Table*, go to next DNSKEY.
+   2. Else, if the algorithm is listed in *Key Size Table* and the
+      key size is smaller than specified, then output 
+      *[DNSKEY_TOO_SMALL_FOR_ALGO]*.
+   3. Else, if the algorithm is listed in *Key Size Table* and the
+      key size is smaller than 2048 bits, then output
+      *[DNSKEY_SMALLER_THAN_REC]*.
+   3. Else, if the algorithm is listed in *Key Size Table* and the
+      key size is largeer than specified, then output
```
sandoche2k: larger than

docs/specifications/tests/DNSSEC-TP/README.md
sandoche2k: Just saw DNSSEC 13. Pls. Ignore the above comment

docs/specifications/tests/DNSSEC-TP/dnssec14.md
matsduf: Corrected. Please approve again.

matsduf:

Issue #475, opened 2016-10-18
Does not check the lack of NSEC3 records ?


bortzmeyer:
Among the three name servers of .sx, one does not return NSEC3 records:

```
% dig @b.ns.sx A hfgygtfgsqdghfkjsdhgsdkffghd.sx

; <<>> DiG 9.10.3-P4-Debian <<>> @b.ns.sx A hfgygtfgsqdghfkjsdhgsdkffghd.sx
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 43115
;; flags: qr aa rd; QUERY: 1, ANSWER: 0, AUTHORITY: 2, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 4096
;; QUESTION SECTION:
;hfgygtfgsqdghfkjsdhgsdkffghd.sx. IN A

;; AUTHORITY SECTION:
sx.         86400 IN SOA a.ns.sx. sx.openregistry.com. (
                15437      ; serial
                14400      ; refresh (4 hours)
                3600       ; retry (1 hour)
                604800     ; expire (1 week)
                86400      ; minimum (1 day)
                )
sx.         86400 IN RRSIG SOA 7 1 86400 (
                20170415212004 20161017212004 39185 sx.
                YidMh+bq2M6rir6SYosVfGmGaX6OPnYSotsYQaL/q/1M
                RSKXnoLtwYroqxGaLvZNSs76Re7yj3jp9YxtZrZogSl6
                rcLVe+MQb72kdgIwJhFZ+3TA+qHNhkVYPzNdqBxF52h1
                04mdbpfiqzQ5PQkJJ2GO4fkeyN0ZRqhnpqejjaE= )

;; Query time: 42 msec
;; SERVER: 80.92.90.160#53(80.92.90.160)
;; WHEN: Tue Oct 18 09:29:06 CEST 2016
;; MSG SIZE  rcvd: 282
```

The others do:

```
% dig @a.ns.sx A hfgygtfgsqdghfkjsdhgsdkffghd.sx

; <<>> DiG 9.10.3-P4-Debian <<>> @a.ns.sx A hfgygtfgsqdghfkjsdhgsdkffghd.sx
; (2 servers found)
;; global options: +cmd
;; Got answer:
[...]

PR #786, opened 2019-07-15
Updated DNSSEC10

matsduf:
* Moved into new format.
* Added explicit messages.
* Added check on all servers.
* Check for inconsistent NSEC/NSEC3.

This PR is meant to solve issue #475.

When this PR is merged, an issue will be created in Zonemaster-Engine to have the implementation of DNSSEC10 updated to meet the updated specification. Any messages in today's implementation should be removed unless they are part of the specification.

### A note to the reviewer

The following text found in the proposal
```
(...)
   7. Else, if the authority section has one or more NSEC records
      but no NSEC3 records (one or more NSEC3 records but no NSEC
      records), then do:
      1. Add the name server IP to the *NSEC-Zone* set (*NSEC3-Zone*
         set).
      2. Retreive all NSEC (NSEC3) records from the response.
      (...)
(...)
```
is meant to be an abbreviation for
```
(...)
   7. Else, if the authority section has one or more NSEC records
      but no NSEC3 records, then do:
      1. Add the name server IP to the *NSEC-Zone* set.
      2. Retreive all NSEC records from the response.
      (...)
   8. Else, if the authority section has one or more NSEC3 records
       but no NSEC records), then do:
      1. Add the name server IP to the *NSEC3-Zone* set.
      2. Retreive all NSEC3 records from the response.
      (...)
(...)
```
Does it work or is it better to have it spelled out?

### DIFF
--- a/docs/specifications/tests/DNSSEC-TP/dnssec10.md
+++ b/docs/specifications/tests/DNSSEC-TP/dnssec10.md
@@ -1,48 +1,172 @@
-## DNSSEC10: Zone contains NSEC or NSEC3 records
+# DNSSEC10: Zone contains NSEC or NSEC3 records
 
-### Test case identifier
-**DNSSEC10** Zone contains NSEC or NSEC3 records
+## Test case identifier
+**DNSSEC10**
 
-### Objective
+## Objective
 
-Using DNSSEC, having NSEC or NSEC3 records provides a secure denial of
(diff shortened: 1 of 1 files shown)


matsduf:

matsduf:
Issue zonemaster/zonemaster-engine#618 for the implementation of these changes.

PR #794, opened 2019-08-01
Create Zone10

matsduf:
Create a new test case that will verify that all servers for a zone
returns exactly one SOA record.

This solves issue #666.

When this PR has been merged an issue will be created in Zonemaster-Engine.

PR #802, opened 2019-10-04
Updated Zone10

matsduf:
* Changed message (message tag) tog get a unique tag that does
  not collide with messages of other test cases in the same
  module.
* Raised the level to warning on that message.
* Sorted the messages in list.

Zone10 was created by #794.

The ongoing implementation of Zone10 (zonemaster-engine#606) showed a collision between different usages of the same message (message tag). This PR is to resolve that by change the message (message tag).


matsduf:
@vlevigneron, if you can approve, I can merge.

PR #831, opened 2020-03-19
Updated version of Firefox and Chrome

matsduf:
Updated the versions of Firefox and Chrome, respectively, supported by Zonemaster.


sandoche2k:
@pamasse can you review so that it can be merged

PR #837, opened 2020-04-02
Updated information on limitations with translation.

matsduf:
This information should be updated when all OSs have been fully tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,35 +133,19 @@ Ubuntu 16.04     | no
 Ubuntu 18.04     | yes
 
 
-## Localization and translation
+## Translation
 
-Zonemaster comes with translation for the following
-languages
+Zonemaster comes with translation to the following languages. Translation is
+available as methods in `Zonemaster::Engine`, `zonemaster-cli` (i.e. the
+Zonemaster-CLI interface to `Zonemaster::Engine`), Zonemaster-Backend
+`RPCAPI` interface to `Zonemaster::Engine`) and the Zonemaster-GUI interface
+to `RPCAPI`.
+
 
 * English (en, en_US.UTF-8)
 * French (fr, fr_FR.UTF-8)
 * Swedish (sv, sv_SE.UTF-8)
 * Danish (da, da_DK.UTF-8)
-
-The support of localization or translation is however not complete due to
-complexity in the implementation of `gettext` functions in different OSs.
-The table below documents in what OS the translation function works for
-Engine (i.e. the Zonemaster libraries), the CLI (i.e. the CLI interface
-to the Engine) and Backend (i.e. the RPCAPI interface to the Engine). If
-translation does not work properly in the Backend, it will not work in
-GUI either.
-
-Operating System | Engine | CLI | Backend
----------------- | -------|-----|--------
-CentOS 7         | ?      |  ?  |   ?
-CentOS 8         | ?      |  ?  |   ?
-Debian 9         | ?      |  ?  |   ?
-Debian 10        | ?      |  ?  |   ?
-FreeBSD 11.3     | yes    | yes | no
-FreeBSD 12.1     | yes    | yes | yes
-Ubuntu 16.04     | ?      |  ?  |   ?
-Ubuntu 18.04     | ?      |  ?  |   ?
-
 
 ## Zonemaster and its components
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -133,16 +133,35 @@ Ubuntu 16.04     | no
 Ubuntu 18.04     | yes
 
 
-## Localization
+## Localization and translation
```
mattias-p, line 136: IMHO Localization is a better heading because localization included translation.

README.md
```diff
@@ -133,16 +133,35 @@ Ubuntu 16.04     | no
...
+Zonemaster comes with translation for the following
+languages
+
+* English (en, en_US.UTF-8)
+* French (fr, fr_FR.UTF-8)
+* Swedish (sv, sv_SE.UTF-8)
+* Danish (da, da_DK.UTF-8)
+
+The support of localization or translation is however not complete due to
+complexity in the implementation of `gettext` functions in different OSs.
```
mattias-p, line 147: Should we really be shifting blame here? IMHO it would be better to just state what works and what doesn't. Maybe with a link to some issue in case anyone is interested.

README.md
```diff
@@ -133,16 +133,35 @@ Ubuntu 16.04     | no
...
+GUI either.
+
+Operating System | Engine | CLI | Backend
+---------------- | -------|-----|--------
+CentOS 7         | ?      |  ?  |   ?
+CentOS 8         | ?      |  ?  |   ?
+Debian 9         | ?      |  ?  |   ?
+Debian 10        | ?      |  ?  |   ?
+FreeBSD 11.3     | yes    | yes | no
+FreeBSD 12.1     | yes    | yes | yes
```
mattias-p, line 161: AFAICT Engine is relying on undefined behavior in the l10n APIs we're using. It happens to work on Linux but on FreeBSD they're implemented differently. CLI was working around this bug in Engine all along, and now Backend is working around it too. IMHO we should either integrate the workaround in Engine or say something other than yes here.

README.md
```diff
@@ -133,16 +133,35 @@ Ubuntu 16.04     | no
...
+translation does not work properly in the Backend, it will not work in
+GUI either.
+
+Operating System | Engine | CLI | Backend
+---------------- | -------|-----|--------
+CentOS 7         | ?      |  ?  |   ?
+CentOS 8         | ?      |  ?  |   ?
+Debian 9         | ?      |  ?  |   ?
+Debian 10        | ?      |  ?  |   ?
+FreeBSD 11.3     | yes    | yes | no
```
mattias-p, line 160: To the best of our knowledge it isn't possible to get translation working in Engine in all circumstances on FreeBSD 11. IMHO we shouldn't say "yes" for Engine here. Maybe "yes*" or something else.

Adding some complexity, I believe the point I make for FreeBSD 12 applies to FreeBSD 11 too.

README.md
```diff
@@ -133,16 +133,35 @@ Ubuntu 16.04     | no
...
+
+Zonemaster comes with translation for the following
+languages
+
+* English (en, en_US.UTF-8)
+* French (fr, fr_FR.UTF-8)
+* Swedish (sv, sv_SE.UTF-8)
+* Danish (da, da_DK.UTF-8)
+
+The support of localization or translation is however not complete due to
```
mattias-p, line 146: IMHO this part should go into the "Notable bugs and issues" section.

README.md
matsduf: I see no problem to point out translation because that is our focus. We do no other localization.

README.md
matsduf: I do not understand what you mean by "shifting blame". Do you want that specific sentence to be removed and replaced by a link to an issue containing what?

README.md
matsduf: OK, a "no" for Engine for FreeBSD?

README.md
matsduf: What part?

README.md
mattias-p: That's what I had in mind.

README.md
mattias-p: Translation does work in some contexts, e.g. CLI, so a simple no isn't accurate either. I suggest "partial" with a link to an issue. I'm not sure if there is an issue to link to. Do you want me to create one?

README.md
mattias-p: Line 146 and below.

README.md
matsduf: I can remove that sentence, but I have no issue to link to. It is OK.

matsduf: Updated.

README.md
matsduf: Updated.

README.md
matsduf: Updated.

README.md
matsduf: Updated.

PR #838, opened 2020-04-06
Updated to handle difference between FreeBSD Make and GNU Make

matsduf:
`share/Makefile` in Zonemaster-Engine and Zonemaster-CLI, respectively, only works with GNU Make. FreeBSD Make is not "GNU Make". `gmake` in FreeBSD (and seperately installed) is "GNU Make". The Release Process document is updated to hand that.

Also see zonemaster/zonemaster-engine#697 and zonemaster/zonemaster-cli#132.

### DIFF
--- a/docs/internal-documentation/maintenance/ReleaseProcess.md
+++ b/docs/internal-documentation/maintenance/ReleaseProcess.md
@@ -96,7 +96,14 @@ version              | version number of the new release
 
 ## 6. Verify that MANIFEST is up to date and that tarball can be built
 
-> This section is not relevant for Zonemaster-GUI.
+> This section is *NOT* relevant for Zonemaster-GUI.
+
+> `share/Makefile` in Zonemaster-Engine and Zonemaster-CLI, respectively,
+> only works with GNU Make. FreeBSD Make is not "GNU Make". `gmake` in
+> FreeBSD (and seperately installed) is "GNU Make".
+
+    gmake -C share touch-po all # Extra step in "Engine" in FreeBSD only.
+    gmake -C share all          # Extra step in "CLI" in FreeBSD only.
 
 Build generated files (if any) and verify that a distribution tarball can be 
 successfully built for each component that is to be updated in this release.
--- a/docs/internal-documentation/maintenance/ReleaseProcess.md
+++ b/docs/internal-documentation/maintenance/ReleaseProcess.md
@@ -98,12 +98,17 @@ version              | version number of the new release
 
 > This section is *NOT* relevant for Zonemaster-GUI.
 
-> `share/Makefile` in Zonemaster-Engine and Zonemaster-CLI, respectively,
-> only works with GNU Make. FreeBSD Make is not "GNU Make". `gmake` in
-> FreeBSD (and seperately installed) is "GNU Make".
+### 6.1 FreeBSD only
 
-    gmake -C share touch-po all # Extra step in "Engine" in FreeBSD only.
-    gmake -C share all          # Extra step in "CLI" in FreeBSD only.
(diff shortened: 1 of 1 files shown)


docs/internal-documentation/maintenance/ReleaseProcess.md
```diff
@@ -96,7 +96,14 @@ version              | version number of the new release
...
 
-> This section is not relevant for Zonemaster-GUI.
+> This section is *NOT* relevant for Zonemaster-GUI.
+
+> `share/Makefile` in Zonemaster-Engine and Zonemaster-CLI, respectively,
+> only works with GNU Make. FreeBSD Make is not "GNU Make". `gmake` in
+> FreeBSD (and seperately installed) is "GNU Make".
+
+    gmake -C share touch-po all # Extra step in "Engine" in FreeBSD only.
+    gmake -C share all          # Extra step in "CLI" in FreeBSD only.
```
mattias-p, line 106: Why not just `gmake all` for all components except GUI on BSD?

docs/internal-documentation/maintenance/ReleaseProcess.md
```diff
@@ -96,7 +96,14 @@ version              | version number of the new release
 
 ## 6. Verify that MANIFEST is up to date and that tarball can be built
 
-> This section is not relevant for Zonemaster-GUI.
+> This section is *NOT* relevant for Zonemaster-GUI.
+
+> `share/Makefile` in Zonemaster-Engine and Zonemaster-CLI, respectively,
+> only works with GNU Make. FreeBSD Make is not "GNU Make". `gmake` in
+> FreeBSD (and seperately installed) is "GNU Make".
```
mattias-p, line 103: I suggest the explanations be kept in the PR description. It's not difficult to trace your way back there for a given source line.

docs/internal-documentation/maintenance/ReleaseProcess.md
mattias-p, line 106: The commands to execute for BSD and GNU respectively should be more clearly marked up.

matsduf: It does not work and I am not sure it is a good idea.

Main Makefile is generated from Makefile.PL and in the FreeBSD case probably adapted to FreeBSD Make. In FreeBSD it is probably safer to run "make" than "gmake".

And running "gmake" on the main Makefile would not even help because it contains "MAKE=make" for the execution of share/Makefile. Then you would need to run "make MAKE=gmake", but then that "MAKE" might be used for something where gmake is not suitable.

docs/internal-documentation/maintenance/ReleaseProcess.md
matsduf: I could shorten the comment, but it is good to have something there.

docs/internal-documentation/maintenance/ReleaseProcess.md
matsduf: Will fix.

matsduf:
@mattias-p and @sandoche2k, please re-review.

docs/internal-documentation/maintenance/ReleaseProcess.md
mattias-p: My goodness. I'll look into making the Makefiles portable.

docs/internal-documentation/maintenance/ReleaseProcess.md
mattias-p: I'll accept that.

matsduf:
@sandoche2k, still OK with you?

PR #839, opened 2020-04-08
Updated treatment of DS SHA-1 in DNSSEC01

matsduf:
ERROR was felt to be too harsh due to the wide use, and due to the
fact that SHA-1 based DNSSEC algorithms only create WARNING as
default. Secondly, IANA still lists SHA-1 as algorithm to be used for DS.

Commit 1: Updated the treatment of algo 1 (SHA-1)
* Add an initial comment on the status of SHA-1
* Created new message tag
* Set default level to WARNING (instead of ERROR)

Commit 2: Removed trailing spaces

Commit 3:
* Sorted message tags in table.
* Sorted link list.
* Updated indentation in link list.

When this PR has been approved and merged, a new issue will be created in Zonemaster-Engine for the implementation to be updated.

### DIFF
--- a/docs/specifications/tests/DNSSEC-TP/dnssec01.md
+++ b/docs/specifications/tests/DNSSEC-TP/dnssec01.md
@@ -107,14 +107,14 @@ In other cases the outcome of this Test Case is "pass".
 
 Message                       | Default severity level
 :-----------------------------|:-----------------------------------
+DS_ALGORITHM_DEPRECATED       | ERROR
+DS_ALGORITHM_MISSING          | NOTICE
+DS_ALGORITHM_NOT_DS           | ERROR
+DS_ALGORITHM_OK               | INFO
+DS_ALGORITHM_RESERVED         | ERROR
+DS_ALGO_SHA1_DEPRECATED       | WARNING
 NO_RESPONSE_DS                | WARNING
 UNEXPECTED_RESPONSE_DS        | WARNING
-DS_ALGORITHM_NOT_DS           | ERROR
-DS_ALGORITHM_DEPRECATED       | ERROR
-DS_ALGORITHM_RESERVED         | ERROR
-DS_ALGORITHM_OK               | INFO
-DS_ALGORITHM_MISSING          | NOTICE
-DS_ALGO_SHA1_DEPRECATED       | WARNING
 
 ## Special procedural requirements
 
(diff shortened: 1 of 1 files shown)


docs/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -5,19 +5,23 @@
...
-If [RFC 8624] and the [IANA registry] disagree on the same DS digest 
-algorithm, the RFC takes precedence until the registry has a been 
+If [RFC 8624] and the [IANA registry] disagree on the same DS digest
+algorithm, the RFC takes precedence until the registry has a been
 updated with a reference to the RFC.
 
-The table of algorithms below is for reference only and is copied from [IANA 
+At the time of writing (2020-04-08), algorithm 1 (SHA-1) is still
+in wide use even though it is no longer considered to be secure
+([Wikipedia]).
```
sandoche2k, line 20: Is Wikipedia a reliable reference?

matsduf: Yes, Wikipedia has good checks of the content.

docs/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -15,6 +15,10 @@ If [RFC 8624] and the [IANA registry] disagree on the same DS digest
 algorithm, the RFC takes precedence until the registry has a been 
 updated with a reference to the RFC.
 
+At the time of writing (2020-04-08), algorithm 1 (SHA-1) is still
```
mattias-p, line 18: "At the time of writing (2020-04-08)" -> "As of 2020-04-08"?

docs/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -132,3 +141,5 @@ None.
 [DS_ALGORITHM_RESERVED]:   #outcomes
 [DS_ALGORITHM_OK]:         #outcomes
 [DS_ALGORITHM_MISSING]:    #outcomes
+[DS_ALGO_SHA1_DEPRECATED]:     #outcomes
+[Wikipedia]:                   https://en.wikipedia.org/wiki/SHA-1
```
mattias-p, line 144: Maybe change the link label to include "SHA-1"? If you want the link text to be something other than the link label that can be achieved using a [full reference link](https://spec.commonmark.org/0.29/#full-reference-link).

docs/specifications/tests/DNSSEC-TP/dnssec01.md
matsduf: Yes, that is an alternative. Both works. None is better than the other.

docs/specifications/tests/DNSSEC-TP/dnssec01.md
matsduf: If there were multiple references to different locations at Wikipedia I would agree (and done that way) but in this case there is only one reference to Wikipedia so I do not see the gain.

matsduf:
@vlevigneron, can you also review?

PR #847, opened 2020-04-29
Updated translation status in README

matsduf:
* Updated translation status
* Updated database versions
* Updated "notable bugs and issues"

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,19 +60,16 @@ and processor architecture listed below.
 
 ### Supported database engine versions
 
-Operating System | MySQL/MariaDB | PostgreSQL
----------------- | --------------| -----------
-CentOS 7         | 5.6 (My) ?    | 9.3 ?
-CentOS 8         | 8.0 (My) ?    | 9.6 ?
-Debian 9         | 10.1 (Ma)     | 9.6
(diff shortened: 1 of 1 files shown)


sandoche2k:
@matsduf for CentOS 8, the PSQL versions available are 9.6, 10 and 12
For translation, it works for CLI for both CentOS 7 & 8
Same from the GUI. If it works in the GUI, then the backend should be Ok, Isn't it ?

matsduf:
> @matsduf for CentOS 8, the PSQL versions available are 9.6, 10 and 12

Which version has been tested? The installation instruction says 9.3.

> For translation, it works for CLI for both CentOS 7 & 8
> Same from the GUI. If it works in the GUI, then the backend should be Ok, Isn't it ?

Yes.

mattias-p:
I've successfully tested translation for both CLI and Backend on both Debian 9 and Debian 10.

matsduf:
@sandoche2k, the README is not complete since I need the information on what databases and versions that you have tested. (It must also match the installation instruction.)

sandoche2k:
For CentOS7 it is 9.3 (which was forced by me) and for CentOS 8, by default it is 10.6

PR #853, opened 2020-05-19
Preparation for release v2019.2.2

matsduf:


matsduf:
@mattias-p and @sandoche2k, a re-review is needed after the second commit.

PR #859, opened 2020-05-31
Create Basic04

matsduf:
This test case will test some basic nameserver and zone functionality. Since it is in "basic" it will always run. 

Many test cases will query the name servers in the delegation or the
name servers appointed by the NS records in the zone for the NS or SOA
record or both. Reporting problem is crucial, but instead of letting
several test cases report the same problems found, all test cases are
made dependent on this test case and this case only will report any
problems found in the following areas:

* Name Server not responding to a query without EDNS over UDP.
* Name Server responding to TCP but not UDP.
* Name Server not including SOA record of *Child Zone* in the answer
  section in the response on a SOA query for *Child Zone*.
* Name Server not including NS record of *Child Zone* in the answer
  section in the response on an NS query for *Child Zone*.
* Name Server not setting the AA flag in a response with SOA or NS in
  answer section.
* Name Server responding with unexpected RCODE (any except "NOERROR")
  on query for SOA or NS for *Child Zone*.

This test case is assumed to be run before Basic02 to be able to capture the odd, but existing, situation where all name servers listen to TCP, but not UDP.

When this test case has been created, other test cases could be updated so that any message reporting the same error could be suppressed or having its level lowered to DEBUG or DEBUG2. 

All message tags in this test case start with "B04_" to firstly make them test case local, and secondly make it easier for the user to find where it comes from. There is no intention that any Zonemaster code should do anything "magical" with that prefix.

### DIFF
--- a/docs/specifications/tests/Basic-TP/basic04.md
+++ b/docs/specifications/tests/Basic-TP/basic04.md
@@ -1,0 +1,124 @@
+# BASIC04: Test of basic nameserver and zone functionality
+
+## Test case identifier
+**BASIC04**
+
+## Objective
+
+Many test cases will query the name servers in the delegation or the
+name servers appointed by the NS records in the zone for the NS or SOA
+record or both. Reporting problem is crucial, but instead of letting
+several test cases report the same problems found, all test cases are
+made dependent on this test case and this case only will report any
+problems found in the following areas:
+
+* Name Server not responding to a query without EDNS over UDP.
+* Name Server responding to TCP but not UDP.
+* Name Server not including SOA record of *Child Zone* in the answer
+  section in the response on a SOA query for *Child Zone*.
+* Name Server not including NS record of *Child Zone* in the answer
+  section in the response on an NS query for *Child Zone*.
+* Name Server not setting the AA flag in a response with SOA or NS in
+  answer section.
+* Name Server responding with unexpected RCODE (any except "NOERROR")
+  on query for SOA or NS for *Child Zone*.
+
+
+## Inputs
+
+* "Child Zone" - The domain name to be tested.
+
+
+## Ordered description of steps to be taken to execute the test case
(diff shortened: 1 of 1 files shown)


matsduf:
@vlevigneron, any comment on this PR before I merge?

matsduf:
@vlevigneron, good point. I had the impression that Basic was always run before all other tests, but as I understand it now, the special thing about Basic is that it is run first and that the test will break if some of the Basic test cases fail. I do not think we should change that, especially not as of this PR. The main use case is when all test cases are run.

If we should change anything it should probably be in the way CLI runs selected test cases or modules, but that would be a different issue/PR.

I will do a slight update to the wording of this PR (test case) so that you do not get the impression that this test case is automatically run before other modules. I will also look at my other PR where I refer to Basic04.

vlevigneron:
@matsduf No, as I told you, Basic is only called in ONE case, when we run all the tests (method Zonemaster::Engine::test_zone()) which is the use case you are talking about. If we plan to run it in this only case, I'll hav to find a way to prevent that same tests are executed twice (first in Basic04, then in test case methods).

matsduf:
> @matsduf No, as I told you, Basic is only called in ONE case, when we run all the tests (method Zonemaster::Engine::test_zone()) which is the use case you are talking about. If we plan to run it in this only case, I'll hav to find a way to prevent that same tests are executed twice (first in Basic04, then in test case methods).

@vlevigneron, my idea is that we should *remove* those tests from the other test cases. Most test cases that I have updated (available PRs) have the equivalent tests remove. I do not think that we should invent something magic.

Basic04 should, in the future, be the only test case that complain about unavailability. If you look at the other PRs you can see that I have tried to avoid to complain about the same thing several times.

matsduf:
@vlevigneron, can you approve this?

PR #864, opened 2020-06-04
Contact and mailing lists

matsduf:
* Public information on mailing lists
* Updated contact information

A new page with mailing list and contact information is created. The main README has been updated to point to that page.

The mail address has been changed.

One idea is that the new page should be linked to from GUI instead of the mail address there.


matsduf:
@mattias-p and @sandoche2k, please review and approve again. This is the same, but against another branch.

PR #879, opened 2020-07-24
Update Basic04

matsduf:
Corrected specification, but no change of test logic. Basic04 will only be run if selected or if all test cases are run. There is nothing magic for when it is run.

This update will not affect zonemaster/zonemaster-engine#763.

### DIFF
--- a/docs/specifications/tests/Basic-TP/basic04.md
+++ b/docs/specifications/tests/Basic-TP/basic04.md
@@ -8,9 +8,9 @@
 Many test cases will query the name servers in the delegation or the
 name servers appointed by the NS records in the zone for the NS or SOA
 record or both. Reporting problem is crucial, but instead of letting
-several test cases report the same problems found, all test cases are
-made dependent on this test case and this case only will report any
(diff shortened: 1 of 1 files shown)


docs/specifications/tests/Basic-TP/basic04.md
```diff
@@ -101,7 +101,7 @@ If either IPv4 or IPv6 transport is disabled, ignore the evaluation of the
 result of any test using this transport protocol and log a message reporting
 the ignored result.
 
-This test case is always run and is run before [Basic02].
+This test case is run before [Basic02].
```
mattias-p, line 104: This is an intercase dependency so it should be listed under that section.

matsduf:
@mattias-p, @vlevigneron and @sandoche2k, please review this PR.

sandoche2k (approved):

matsduf:
@mattias-p, can you review again?

matsduf:
@mattias-p, please review this again.

Issue #795, opened 2019-08-02
Do TestRequirements fill any function anymore?


matsduf:
Among the documents we have [TestRequirements] with listed and numbered requirements separated from the [test cases]. The requirement list was an important input when Zonemaster was created as an input to what test cases to create. Today we do not start by updating [TestRequirements], rather we have an issue pointing out something that should be tested, and then we create the test case. And finally we try to find something to write in [TestRequirements].

~~I suggest that we move [TestRequirements] to the history and instead have the test cases as the starting point.~~

*Updated:* Update the [TestRequirements] and remove references to legacy tools.

I suggest that we create one directory file of all test cases and put that above all test case folders. The README.md in the test case folders are instead static files that points to the new directory file.

All our files are markdown files (.md) and we have scripts that try to use the files as database, which does not work very well. Let us simplify all this by only updating the test case documents and the new directory file.

- [x] Create test case directory file based on existing test cases.
- [x] Update the test case folder README files to the new directory file.
- [x] Create a concatenated README of all test cases at top level.
- [x] Review each requirement and determine if that requirement statement is included in the purpose section of the relevent test case, and if not work it in if possible.
- [x] Remove references to the TestRequirements in README files.
- [x] ~~Move the TestRequirements to the history.~~ Update TestRequirements
- [x] Remove unneeded scripts.

*Updated*: PR #881 resolves this issue.

If this is done #657 can be closed.

[test cases]: https://github.com/zonemaster/zonemaster/tree/master/docs/specifications/tests
[TestRequirements]: https://github.com/zonemaster/zonemaster/blob/master/docs/requirements/TestRequirements.md

pawal:
If you still want to follow IEEE 829-2008 for creating a test plan, you want to keep an updated list of requirements for the tests.

matsduf:
That standard is "superseded" by IEEE, and it is not an open standard. I see no strong value to try to follow a standard that few people know the details of.

matsduf:
Issue description updated at F2F and agreed to proceed.

matsduf:
Resolved by #881

PR #881, opened 2020-07-30
Requirements and specification update

matsduf:
The current requirements document ([TestRequirements.md], current version) was created 7 years ago, and most of was written then 6-7 years ago. I heavily refer to the legacy tools DNSCheck and ZoneCheck, which are hardly used anymore. The requirements statements, in the document, are also inconsistently written.

## Requirements document
With this PR the requirements document has gone through a complete renovation. 
* Firstly, the requirement statements have been updated. All requirements in the original document are there, but often re-written to be clear requirements, i.e. should-statements. New requirements are added, primarily fetched from current test case specifications. Most requirements have also got a reference to an RFC or another source. Most requirements have also been linked to test cases (test case specifications). With this update it will be easy to add new requirements as they come.
* Secondly, all references to the legacy tools have been removed. They have no relevance anymore.

## Documents on legacy tools
In the document structure there were a number of documents just covering the legacy tools. They have no function and have been removed.

## Documents and scripts not used
There were documents never used or updated and scripts never used that have been removed.

## Updated tools
The tools kept have been updated to better meet the needs. The documentation of the tools have also been updated. The purpose of the tools is to generate document, but if that is not done at every release then the documents fall behind.

## Main test case README
The main [test case README] (current version) was also written 6 years ago as a document for the development period. In this PR it has been updated to include a complete list of all test case specifications (test case ID and headline of test case). That part will be updated with the help of one of the tools. Now it will be easy to see all test cases at the same time.

## Level README
All the current level README file contains a list of that level's test cases, but not with the headline of the test case, but with the requirement it was supposed to test, which was confusing. Now they are reduced to be static files with a link to the new, main test case README (see above).

Go to [requirements-and-specification-update] to see the source of this PR, and to read and check links as the document will be when this PR has (hopefully) been merged.

## TO DO (not included in this PR)
* Update the release process to include check of TestRequirements.md plus the generation of TestMessages.md and main test case README.
* Update the scripts (tools) to make the previous bullet easier to do.
* Update the scripts to include references to requirements in the main test case README.

[requirements-and-specification-update]: https://github.com/matsduf/zonemaster/tree/requirements-and-specification-update

[TestRequirements.md]: https://github.com/zonemaster/zonemaster/blob/master/docs/requirements/TestRequirements.md
[test case README]: https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/README.md


matsduf:
This PR resolves #657 and #795.

Issue #649, opened 2018-11-15
Create a list of all Test Cases


matsduf:
The specification for profile (zonemaster/zonemaster-engine#447) in Engine wants to have a list of all implemented test cases, but we have no such list. The file [TestMessages.md] is the closest we have, but would be confusing to point to.

* Create a file that lists all implemented test cases.

Name on file-to-be: TestCases.md
Directory: docs/specifications/tests (relative the Zonemaster repository root)

This issue must be solved before profile is released.

[TestMessages.md]: https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/TestMessages.md

mattias-p:
This also needs to be reconciled with the [Master Test Plan](https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/MasterTestPlan.md).

matsduf:
Use updated version of docs/requirements/TestRequirements.md instead.

matsduf:
PR #882 will resolve this issue.

PR #882, opened 2020-07-31
Resolve issue #649 by creating a list of implemented test cases

matsduf:
Creates a tool to create a list of implemented test cases. The list is also added by this PR. This resolves issue #649.

This PR is also related to zonemaster/zonemaster-engine#768, which could be merged when this PR is merged.


utils/generateImplementedTestCases.pl
```diff
@@ -0,0 +1,34 @@
...
+use 5.16.0;
+use warnings;
+use Zonemaster::Engine;
+use File::Basename;
+
+# page header
+print '<!-- File generated by ', basename ($0), ", script in zonemaster/zonemaster util directory.\n";
+print "Use that script to generate a new file for each release of Zonemaster when \n";
+print "Zonemaster-Engine also has been updated.-->\n\n";
+print "# Implemented Test Cases\n\n";
```
mattias-p, line 12: IMO it would be better to just call this Test Cases. The word "implemented" adds more confusion than clarity. We shouldn't be documenting unimplemented test cases in the master branch anyway. By extension I think the word implemented should be removed every from this PR.

utils/generateImplementedTestCases.pl
```diff
@@ -0,0 +1,34 @@
...
+print "Index of Text Case specifications is also found in [README](README.md).\n";
+print "[Zonemaster-Engine] is the repository of the implementation of the Test Cases (the methods).\n";
+print "\n\n";
+
+# table header
+print "| Module | Method (implemented test case) | Test Case Specification |\n";
```
mattias-p, line 18: IMO we shouldn't be referring to test cases as "methods" in the specifications.

utils/generateImplementedTestCases.pl
```diff
@@ -0,0 +1,34 @@
...
+
+# table content
+foreach my $module ( 'Basic', sort { fc $a cmp fc $b } Zonemaster::Engine::Test->modules ) {
+    my $full = "Zonemaster::Engine::Test::$module";
+    my $ref  = $full->metadata;
+    my $modulelink = "https://metacpan.org/pod/Zonemaster::Engine::Test::$module";
+    for my $method (sort keys %$ref) {
+	my $testcase = uc ($method);
+	my $testcaselink = "${module}-TP/${method}.md";
+	printf "| [%s](%s) | %s | [%s](%s) |\n", $module, $modulelink, $method, $testcase, $testcaselink;
```
mattias-p, line 29: If you could change the indentation to use spaces, that would be nice.

mattias-p:
I have a couple of gripes with naming, but apart from that I think this looks great.

I guess it would be nice with an additional "description" column would be nice, but that can be added later.

mattias-p:
Oh, and the POD-link in Profile.md should probably be updated as well.

utils/generateImplementedTestCases.pl
matsduf: Test cases appear in two shapes in Zonemaster, as test case specifications in Zonemaster/Zonemaster and as implemented test cases (test cases in code) in Zonemaster-Engine. It is that distinction show. In Zonemaster/Zonemaster, as the generated document will appear in, test case usually refer to the specification.

If you could suggest something else that clarifies that this does not refer to the specification I could use that instead.

utils/generateImplementedTestCases.pl
matsduf: "Method" here refer to the Perl code.

utils/generateImplementedTestCases.pl
matsduf: I will fix. They must have come with some copy-and-paste.

utils/generateImplementedTestCases.pl
mattias-p: My suggestion is that we never release changes to the specifications without the corresponding functionality in the implementation.

Having a set of defined test case specifications and a page of links to implemented subset of test cases would work for the situation where we add new test cases - though I don't see how this model benefits our users. But this model breaks down when it comes to changes to existing test cases. A more reasonable model is, IMHO, to merge spec and implementation at the same time.

utils/generateImplementedTestCases.pl
mattias-p: At least it's (locally) consistent that way. Though this is part of the specification, so I think it would be more reasonable to have it speak the language of the specification.

Also, we should be able to change the implementation without invalidating the specification. E.g. we should be able to switch programming languages to Python or Haskell, or switch from OO-style Perl to procedural style Perl.

I think the appropriate terminology here would be:
```
| Test Level | Test Case Identifier | Test Case Specification |
```

Though we seem to be using "Test Level" and "Test Plan" kind of interchangeably, with "Test Plan" begin more common. So maybe "Test Plan" would be better than "Test Level" at this point (even though I believe it to be less technically correct).

utils/generateImplementedTestCases.pl
matsduf: > My suggestion is that we never release changes to the specifications without the corresponding functionality in the implementation.

I agree with you, but don't you miss my point that I want to make a distinction between test case specification and test case implementation in Perl code. Or do I misunderstand you?

utils/generateImplementedTestCases.pl
matsduf: > At least it's (locally) consistent that way. Though this is part of the specification, so I think it would be more reasonable to have it speak the language of the specification.

This is located with the specification, but links to the implementation in Perl (the first column). The second column has the name of the Perl subroutine (method) that implements the test case. The third column has the test case identifier and a link to the specification.

> Also, we should be able to change the implementation without invalidating the specification. E.g. we should be able to switch programming languages to Python or Haskell, or switch from OO-style Perl to procedural style Perl.

In that case this would just be a small part of the updates we would have to do.

mattias-p (approved):
This is a clear improvement from what we had before.

I still don't think it's okay for the specification to refer to the implementation, but we don't need to resolve that right now.

sandoche2k (approved):

PR #890, opened 2020-09-24
Update main README for v2020.1

matsduf:
There are still data that we have to have verified, but if we cannot confirm now I suggest that we merge as a draft version and then update again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,9 +71,9 @@ Debian 9         | 10.1            | 9.6
 Debian 10        | 10.3            | 11.7
 FreeBSD 11.4     | 10.5.5 (###)    | 12.4
 FreeBSD 12.1     | 10.5.5 (###)    | 12.4
-Ubuntu 16.04.7   | 10.0            | 9.5
-Ubuntu 18.04.5   | 10.1            | 10.12
-Ubuntu 20.04.1   | 10.3 (#)        | 12.4 (#)
+Ubuntu 16.04     | 10.0            | 9.5
+Ubuntu 18.04     | 10.1            | 10.12
+Ubuntu 20.04     | 10.3 (#)        | 12.4 (#)
 
 ###) The latest stable in the highest release serie. There is no
 default version for MariaDB in FreeBSD.
@@ -97,9 +97,9 @@ Debian 9         | 5.24
 Debian 10        | 5.28
 FreeBSD 11.4     | 5.32
 FreeBSD 12.1     | 5.32
-Ubuntu 16.0.7    | 5.22
-Ubuntu 18.04.5   | 5.26
-Ubuntu 20.04.1   | 5.30
+Ubuntu 16.04     | 5.22
+Ubuntu 18.04     | 5.26
+Ubuntu 20.04     | 5.30
 
 Zonemaster requieres Perl version 5.14.2 or higher. Zonemaster has been
 tested with the default version of Perl in the OSs as listed in the table above.
--- a/README.md
+++ b/README.md
@@ -57,9 +57,9 @@ and processor architecture listed below.
 * Debian 10
 * FreeBSD 11.4
 * FreeBSD 12.1
-* Ubuntu 16.04.7
-* Ubuntu 18.04.5
-* Ubuntu 20.04.1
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,23 +55,33 @@ and processor architecture listed below.
...
-Ubuntu 18.04     | 10.1 (MariaDB)  | 10.12
+CentOS 7         | 5.5 (#)         | 9.3 (##)
+CentOS 8         | ????            | 10.6 (##)
+Debian 9         | 10.1            | 9.6
+Debian 10        | 10.3            | 11.7
+FreeBSD 11.4     | 10.5.5 (###)    | 12.4
+FreeBSD 12.1     | 10.5.5 (###)    | 12.4
+Ubuntu 16.04.7   | 10.0            | 9.5
+Ubuntu 18.04.5   | 10.1            | 10.12
+Ubuntu 20.04.1   | 10.3 (#)        | 12.4 (#)
```
sandoche2k, line 76: The third decimal for Ubuntu is not needed, I think.

matsduf: I think functionally the third level on Ubuntu is more like the second level of FreeBSD. You cannot refer to version Ubuntu 18, because 18.04 and 18.10 are two different versions, like FreeBSD 11 and FreeBSD 12.

Maybe we should just say FreeBSD 11, but there could be more than subversion of 11 at the same time. Currently we have both 11.3 and 11.4 as supported versions (even though 11.3 is soon deprecated).

I will remove the subversion of Ubuntu. We do not specify it for neither Debian nor CentOS.

README.md
```diff
@@ -53,23 +55,33 @@ and processor architecture listed below.
...
+CentOS 8         | ????            | 10.6 (##)
+Debian 9         | 10.1            | 9.6
+Debian 10        | 10.3            | 11.7
+FreeBSD 11.4     | 10.5.5 (###)    | 12.4
+FreeBSD 12.1     | 10.5.5 (###)    | 12.4
+Ubuntu 16.04     | 10.0            | 9.5
+Ubuntu 18.04     | 10.1            | 10.12
+Ubuntu 20.04     | 10.3 (#)        | 12.4 (#)
+
+###) The latest stable in the highest release serie. There is no
```
mattias-p, line 78: I'm not sure I agree with the choice of syntax here.

mattias-p (dismissed):
It's good to have the versions of OSes and dependencies updated in the develop branch.

README.md
mattias-p: I don't think it helps to be too granular about what releases we support. No one will dig through archives anyway just to download the exact point-release we specified in our documentation. And if you update your OS, as you should, you still get something slightly different anyway.

matsduf: I have updated the file and only indicate one level on Linux OSs. (I consider 18.04 for Ubuntu to  one level, comparable with 9 for Debian.) For FreeBSD I have kept the second level, as we did last time. Is that fine?

README.md
matsduf: Me neither. They are only meant to be in the draft version of the file, not the final version for the release. I wrote them for you, the reviewer of this PR.

matsduf:
> It's good to have the versions of OSes and dependencies updated in the develop branch.

@mattias-p, I am not sure if I understand your comment.

matsduf:
@sandoche2k, can you review this again?

sandoche2k (dismissed):

README.md
```diff
@@ -53,23 +55,33 @@ and processor architecture listed below.
...
 * Debian 9
 * Debian 10
-* FreeBSD 11.3
+* FreeBSD 11.4
 * FreeBSD 12.1
-* Ubuntu 16.04
-* Ubuntu 18.04
+* Ubuntu 16.04.7
+* Ubuntu 18.04.5
+* Ubuntu 20.04.1
```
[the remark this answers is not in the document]
matsduf: I missed that. I have removed it. Please add your formal approval again, @sandoche2k and @mattias-p.

sandoche2k (approved):

mattias-p (approved):

PR #897, opened 2020-10-27
Fix typos

pnax:
Small fixes in the documentation.

### DIFF
--- a/docs/internal-documentation/maintenance/ReleaseProcess-preparation.md
+++ b/docs/internal-documentation/maintenance/ReleaseProcess-preparation.md
@@ -44,7 +44,7 @@ Makefile.PL must be updated.
      below and that the value at each path matches the listed requirement.
 
 
-## 6. Verify that MANIFEST and MANIFEST.SKIP are up to date
+## 4. Verify that MANIFEST and MANIFEST.SKIP are up to date
 
 > This section is *NOT* relevant for Zonemaster-GUI or Zonemeaster/Zonemaster.
 
--- a/docs/internal-documentation/maintenance/ReleaseProcess-preparation.md
+++ b/docs/internal-documentation/maintenance/ReleaseProcess-preparation.md
@@ -23,7 +23,7 @@ is up to date with the [supported Perl versions] in main README.md.
  * zonemaster-cli - [.travis.yml][cli.travis]
  * zonemaster-backend - [.travis.yml][backend.travis]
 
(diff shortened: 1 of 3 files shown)


docs/internal-documentation/maintenance/ReleaseProcess-preparation.md
```diff
@@ -23,7 +23,7 @@ is up to date with the [supported Perl versions] in main README.md.
  * zonemaster-cli - [.travis.yml][cli.travis]
  * zonemaster-backend - [.travis.yml][backend.travis]
 
-(zonemaster-gui uses no Perl and (zonemaster/zonemaster uses no Travis.)
+> Zonemaster-GUI uses no Perl and Zonemaster/Zonemaster uses no Travis.
```
mattias-p, line 26: I think lower case zonemaster/zonemaster is more correct here. That's how the organization and repository are spelled on Github.

pnax: As I found it spelt with the first letter uppercased in the document I thought they were used without distinction. The idea behind this modification was to be aligned with the rest of the document :

in [section 3](https://github.com/zonemaster/zonemaster/blob/develop/docs/internal-documentation/maintenance/ReleaseProcess-preparation.md#3-verify-that-metayml-will-have-all-the-correct-data)
```
> This section is not relevant for Zonemaster-GUI or Zonemaster/Zonemaster.
```

and in [section 6]() (which should be section 4 I think)
```
> This section is NOT relevant for Zonemaster-GUI or Zonemeaster/Zonemaster.
```

Do you want me to update the other occurrences of `Zonemaster/Zonemaster` to `zonemaster/zonemaster` ? Or is the lower case only required in this particular case ?

mattias-p: Oh, I didn't realize. I don't recall seeing Zonemaster/Zonemaster before, but it's actually used in a couple of places outside of this document, and in fact it's more common than zonemaster/zonemaster (unless you count URLs). I guess we ought to look through all of our documentation and come up with a preferred way of referring to the main repository. Because we're not doing it in a consistent way.

I guess Zonemaster/Zonemaster is fine for now. At least it's in good company.

pnax: Ok. I leave it like that for now. And I just fixed the section number.

mattias-p (approved):

PR #904, opened 2020-11-05
Updated instructions for release process, build enviroment and zip file creation

matsduf:


sandoche2k (approved):

docs/internal-documentation/maintenance/ReleaseProcess-create-test-distribution.md
```diff
@@ -87,23 +87,28 @@ MANIFEST.SKIP, i.e. no missing or extra files:
 
 > This section is relevant for Zonemaster-GUI only.
 
-The requirements are nodejs and npm. There are available from the [Node.js]
-official website. Minimal version of Nodejs is 10.0 but install the last LTS
-version available. It was tested on Ubuntu 18.04.
+For this you need a [build environment for nodjs], on which you create
```
pnax, line 90: I think this should be `nodejs` (and maybe `Nodejs`)

docs/internal-documentation/maintenance/ReleaseProcess-release.md
```diff
@@ -172,23 +174,27 @@ MANIFEST.SKIP, i.e. no missing or extra files:
 
 > This section is relevant for Zonemaster-GUI only.
 
-The requirements are nodejs and npm. There are available from the [Node.js]
-official website. Minimal version of Nodejs is 10.0 but install the last LTS
-version available. It was tested on Ubuntu 18.04.
+For this you need a [build environment for nodjs], on which you create
```
pnax, line 177: same : `nodejs` or `Node.js` or `Nodejs` ?

pnax (commented):
I think `nodjs` is wrong. But I don't know if it should be `nodejs`, `Nodejs` or `Node.js` ?

Therefore :
* shouldn't the title of the file `Ubuntu nodjs Build Environment` be `Ubuntu nodejs Build Environment` ? (or something else based on the choice made)
* line 6 in file `Ubuntu nodjs Build Environment` : `Nodjs and npn` -> `Nodejs and npm`

matsduf:
Please note that the milestone is v2020.2 so we could look at it after the release of v2020.1.

docs/internal-documentation/maintenance/ReleaseProcess-create-test-distribution.md
matsduf: Good catch. And that was not the only place were I had it wrong. They write i `Node.js` and I guess there is no reason not to follow that. I'll correct.

docs/internal-documentation/maintenance/ReleaseProcess-release.md
matsduf: `Node.js` I think.

matsduf:
@sandoche2k, @pnax and @mattias-p, please review this while testing and release process are in fresh memory.

sandoche2k (dismissed):

docs/internal-documentation/maintenance/ReleaseProcess-release.md
```diff
@@ -299,9 +299,70 @@ The releases pages:
...
+
+The versions of Zonemaster::LDNS and Zonemaster::Engine are defined in the
+format `vX.Y.Z` and it is important how this is written in `Makefile.PL`.
+
+For lowest risk of error follow the following steps:
+
+1. Expand the version number with leading zeros on second ("Y") and third
+   ("Z") level. Remove the leading `v` and the second and third dots.
+   E.g. `v2.1.2` = `2.001002`.
+2. Decide if the third level can be skipped, e.g. `v2.1.2` > `2.001`.
```
mattias-p, line 317: Since we're using semantic versioning the policy should be to always restrict the version requirement to only include the major and minor numbers.

docs/internal-documentation/maintenance/ReleaseProcess-release.md
```diff
@@ -82,34 +82,32 @@ The GUI has no Perl. Update the following files in the *develop branch*:
...
+### Zonemaster-CLI and Zonemaster-Backend
 
-In [Zonemaster-CLI Makefile.PL] set the lowest version of Zonemaster-LDNS and
-Zonemaster-Engine that Zonemaster-CLI requires. If unsure, set the versions
-included in this release since that is what has been tested. E.g.
+In [Zonemaster-CLI Makefile.PL] and in [Zonemaster-Backend Makefile.PL],
+respectively, set the lowest version of Zonemaster-LDNS and
+Zonemaster-Engine that Zonemaster-CLI and Zonemaster-Backend, respectively,
+require. Usually, set the versions included in this release since that is
+what has been tested. E.g.
```
mattias-p, line 104: If we say "usually" here, we should at least give an example of what constitutes an unusual situation. Such an example is when Backend/CLI doesn't use any of the features introduced in the dependency since its last minor version bump.

docs/internal-documentation/maintenance/ReleaseProcess-release.md
```diff
@@ -82,34 +82,32 @@ The GUI has no Perl. Update the following files in the *develop branch*:
...
-### Zonemaster-CLI
+### Zonemaster-CLI and Zonemaster-Backend
 
-In [Zonemaster-CLI Makefile.PL] set the lowest version of Zonemaster-LDNS and
-Zonemaster-Engine that Zonemaster-CLI requires. If unsure, set the versions
-included in this release since that is what has been tested. E.g.
+In [Zonemaster-CLI Makefile.PL] and in [Zonemaster-Backend Makefile.PL],
+respectively, set the lowest version of Zonemaster-LDNS and
+Zonemaster-Engine that Zonemaster-CLI and Zonemaster-Backend, respectively,
+require. Usually, set the versions included in this release since that is
```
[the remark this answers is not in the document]
matsduf: Will fix.

docs/internal-documentation/maintenance/ReleaseProcess-release.md
matsduf: I do not agree. We should normally set what we have tested to lower the risk of problem.

docs/internal-documentation/maintenance/ReleaseProcess-release.md
matsduf: To save us problem we should steer the users to install versions that constitute a Zonemaster product version, because that is what we have tested. If in a new product version includes updated Engine but the same CLI as the previous version, then we do not update what version CLI will require.

matsduf:
@mattias-p, @sandoche2k  and @pnax, please review again.

mattias-p (approved):

sandoche2k (approved):

PR #906, opened 2020-11-09
Fix merging develop into master

mattias-p:
The current steps to merge develop into master fail when new files have been added since last release. The steps fail because the `git commit --include` command only considers files tracked by the master branch. This leaves untracked files in the working tree that cause the final `git merge` command to fail.

The solution here is to make the `git reset` command keep all changes in the index. This ensures that the `git commit` command includes all the changes. The squashed commit will be complete and the working directory will be clean, clearing the way for the `git merge` command.


matsduf (approved):
When I test in Zonemaster-Engine it works as expected, I think. See https://github.com/zonemaster/zonemaster-engine/pull/837

sandoche2k (approved):

PR #919, opened 2021-01-09
Updates in preparation for release v2020.2

matsduf:
SQLite version has to be verified for all OSs execept FreeBSD.

Firefox and Chrome versions for Windows and Ubuntu, respectively, have to be verified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,30 +51,27 @@ and processor architecture listed below.
 
 * CentOS 7
 * CentOS 8
-* Debian 9
 * Debian 10
 * FreeBSD 11.4
 * FreeBSD 12.2
-* Ubuntu 16.04
 * Ubuntu 18.04
 * Ubuntu 20.04
 
 ### Supported database engine versions
 
-Operating System | MariaDB         | PostgreSQL
----------------- | ----------------| -----------
-CentOS 7         | 5.5             | 9.3 
-CentOS 8         | 10.3            | 10.14 
-Debian 9         | 10.1            | 9.6
-Debian 10        | 10.3            | 11.7
-FreeBSD 11.4     | 5.7 (#)         | 12.4
-FreeBSD 12.2     | 5.7 (#)         | 12.4
-Ubuntu 16.04     | 10.0            | 9.5
-Ubuntu 18.04     | 10.1            | 10.14
-Ubuntu 20.04     | 10.3            | 12.4
+Operating System | MariaDB | PostgreSQL | SQLite
+---------------- | --------| -----------|------------------------
+CentOS 7         | 5.5     | 9.3        | 3.7.17 (?)
+CentOS 8         | 10.3    | 10.14      | 3.34 (?)
+Debian 10        | 10.3    | 11.7       | 3.27.2 (?)
+FreeBSD 11.4     | 5.7 (#) | 12.4       | 3.32.3 (##)
+FreeBSD 12.2     | 5.7 (#) | 12.4       | 3.32.3 (##)
+Ubuntu 18.04     | 10.1    | 10.14      | 3.22.0 (?)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -130,18 +122,7 @@ installation instructions given for Zonemaster have been followed. A test of the
...
-CentOS 8         | yes
-Debian 9         | no
-Debian 10        | yes
-FreeBSD 11.4     | yes
-FreeBSD 12.1     | yes
-Ubuntu 16.04     | no
-Ubuntu 18.04     | yes
-Ubuntu 20.04     | yes
-
+All supported OSs, execept CentOS 7, supports algorithms 15 and 16.
```
pnax, line 125: `execept` -> `except`
`supports` -> `support`

matsduf: Fixed.

README.md
```diff
@@ -51,30 +51,27 @@ and processor architecture listed below.
...
+CentOS 7         | 5.5     | 9.3        | 3.7.17 (?)
+CentOS 8         | 10.3    | 10.14      | 3.34 (?)
+Debian 10        | 10.3    | 11.7       | 3.27.2 (?)
+FreeBSD 11.4     | 5.7 (#) | 12.4       | 3.32.3 (##)
+FreeBSD 12.2     | 5.7 (#) | 12.4       | 3.32.3 (##)
+Ubuntu 18.04     | 10.1    | 10.14      | 3.22.0 (?)
+Ubuntu 20.04     | 10.3    | 12.4       | 3.33.0 (?)
 
-#) FreeBSD uses MySQL, not MariaDB.
+(#) FreeBSD uses MySQL, not MariaDB.
```
mattias-p, line 72: Both MariaDB and MySQL are available on FreeBSD.

README.md
```diff
@@ -51,30 +51,27 @@ and processor architecture listed below.
...
 
+(##) FreeBSD bundles SQLite in Perl DBD::SQLite (version 1.66).
```
mattias-p, line 74: When is this footnote useful?

README.md
matsduf: Yes, but it does not work to use MariaDB with Zonemaster. After the release I can look at it to see how we could resolve it.

README.md
matsduf: If you install SQLite 3 as package on FreeBSD you will get version 3.34.0, but the bundled SQLite (3.32.3) is still used.

README.md
mattias-p: Are MySQL and MariaDB incompatible on FreeBSD?

matsduf: I tested to use MariaDB, but DBI::mysql requires that MySQL is installed and that cannot be installed at the same time as MariaDB.

mattias-p: You mean DBD::mysql? I checked /usr/ports/databases/p5-DBD-mysql/Makefile and /usr/ports/databases/p5-DBD-MariaDB/Makefile, and both have USES mysql. Here is a fair amount of guessing and speculation: Could it be that both MySQL and MariaDB implement "mysql" in FreeBSD, that FreeBSD locks in the meaning of "mysql" to the first implementation of it that you install, and that you had MySQL installed before you tried using DBD::mysql and MariaDB together?

matsduf: I can try again at next test section. I did not make work last time.

matsduf: @mattias-p, in FreeBSD:
* Backend requires DBD-mysql
* DBD-mysql requires MySQL-client
* MariaDB-server requires MariaDB-client
* MariaDB-client and MySQL-client cannot both be installed

mattias-p: Too bad. Dependencies can be a real mess. It was worth a shot though.

mattias-p (approved):

PR #922, opened 2021-01-27
Verify SQLite version for Debian an Ubuntu

pnax:
Verify and update the Readme file.


matsduf (approved):
Great!

@mattias-p, can you check what version CentOS runs?

PR #938, opened 2021-04-16
Create DNSSEC15 (Existence of CDS and CDNSKEY)

matsduf:
Also see #939 (PR to create DNSSEC16), #941 (PR to create DNSSEC17) and [New-Test-Cases-for-CDS-CDNSKEY] (a temporary document).

[New-Test-Cases-for-CDS-CDNSKEY]: https://github.com/matsduf/zonemaster/blob/notes-for-new-DNSSEC-test-cases-for-CDS-CDNSKEY/docs/specifications/tests/DNSSEC-TP/New-Test-Cases-for-CDS-CDNSKEY.md

PR #939, opened 2021-04-21
Create DNSSEC16 (Validate CDS) and DNSSEC17 (Validate CDNSKEY)

matsduf:
Now updated to create two test cases, DNSSEC16 for validating CDS and DNSSEC17 for validating CDNSKEY. Previously there was one test case for both.

Also see #938 (PR to create DNSSEC15), #941 (PR to create DNSSEC18 and [New-Test-Cases-for-CDS-CDNSKEY] (a temporary document).

[New-Test-Cases-for-CDS-CDNSKEY]: https://github.com/matsduf/zonemaster/blob/notes-for-new-DNSSEC-test-cases-for-CDS-CDNSKEY/docs/specifications/tests/DNSSEC-TP/New-Test-Cases-for-CDS-CDNSKEY.md

### DIFF
--- a/docs/specifications/tests/DNSSEC-TP/dnssec16.md
+++ b/docs/specifications/tests/DNSSEC-TP/dnssec16.md
@@ -49,9 +49,9 @@ DS16_MIXED_DELETE_CDS                | ERROR   | "Delete" CDS record is mixed wi
 
 1.  Create the following empty sets:
     1.  Name server IP address and associated CDS RRset and its RRSIG
-        records ("CDS RRsets"). RRSIG records may be absent.
+        records ("CDS RRsets"). The set of RRSIG records may be empty.
     2.  Name server IP address and associated DNSKEY RRset and its 
-        RRsig records ("DNSKEY RRsets"). RRSIG records may be absent.
+        RRSIG records ("DNSKEY RRsets"). The set of RRSIG records may be empty.
     3.  Name server IP address ("No DNSKEY RRset").
     4.  Name server IP address ("Mixed Delete CDS").
     5.  Name server IP address ("Delete CDS").
@@ -107,15 +107,16 @@ DS16_MIXED_DELETE_CDS                | ERROR   | "Delete" CDS record is mixed wi
 7.  For each name server IP in the *CDS RRsets* set do:
 
     1. If the CDS RRset is non-empty, get the RRset and its RRSIG
-       records, else go to next name server IP address.
+       records, else go to next name server IP address. The set of the RRSIG
+       records may be empty.
     2. If any CDS record is a "delete" CDS, then do:
        1. If there is more than a single CDS record then add the name
           server IP to the *Mixed Delete CDS* set.
        2. Else, add the name server IP address to the *Delete CDS*
           set.
        3. Go to next name server IP.
-    3. Get the DNSKEY and its RRSIG records, if any, from the
-       *DNSKEY RRsets* for the same name server IP.
+    3. Get the DNSKEY and its RRSIG records from the *DNSKEY RRsets* for the same
+       name server IP. The set of RRSIG records may be empty.
     4. If there are no DNSKEY records, then do:
        1. Add name server IP address to the *No DNSKEY RRset* set.
        2. Go to next name server IP.
--- a/docs/specifications/tests/DNSSEC-TP/dnssec16.md
+++ b/docs/specifications/tests/DNSSEC-TP/dnssec16.md
@@ -41,7 +41,7 @@ DS16_CDS_WITHOUT_DNSKEY              | ERROR   | CDS RRset exists, but no DNSKEY
 DS16_CDS_INVALID_RRSIG               | ERROR   | CDS RRset is signed with an invalid RRSIG.
 DS16_CDS_MATCHES_NO_DNSKEY           | WARNING | CDS record does not match any DNSKEY in DNSKEY RRset.
 DS16_CDS_SIGNED_BY_UNKNOWN_DNSKEY    | ERROR   | CDS RRset is signed but not by a key in DNSKEY RRset.
-DS16_DELETE_CDS                      | INFO    | CDS RRset have a "delete" CDS record as a single record.
+DS16_DELETE_CDS                      | INFO    | CDS RRset has a "delete" CDS record as a single record.
 DS16_DNSKEY_NOT_SIGNED_BY_CDS        | WARNING | DNSKEY RRset is not signed by the key or keys that the CDS records point to.
 DS16_MIXED_DELETE_CDS                | ERROR   | "Delete" CDS record is mixed with normal CDS record.
 
@@ -178,7 +178,7 @@ The outcome of this Test Case is "fail" if there is at least one message
 with the severity level *[ERROR]* or *[CRITICAL]*.
 
 The outcome of this Test Case is "warning" if there is at least one message
-with the [severity level] *[WARNING]*, but no message with severity level
+with the severity level *[WARNING]*, but no message with severity level
 *ERROR* or *CRITICAL*.
 
 In other cases, no message or only messages with severity level 
--- a/docs/specifications/tests/DNSSEC-TP/dnssec16.md
+++ b/docs/specifications/tests/DNSSEC-TP/dnssec16.md
@@ -1,0 +1,340 @@
+# DNSSEC16: Validate CDS and CDNSKEY
+
+## Test case identifier
+**DNSSEC16**
+
+## Objective
+
+CDS and CDNSKEY record types are defined in [RFC 7344] and [RFC 8878].
+Both record types are optional in a zone. The objective of this test
+case is to verify that they are they are valid. This Test case is
+only relevant if the zone has either CDS or CDNSKEY record or both.
+
+If an CDS record is included in the zone, the corresponding CDNSKEY
+record should also be included ([RFC 7344][RFC 7344, section 4],
+section 4).
+
+The CNS and CDNSKEY RRsets should be consistent between all name
+servers for the zone in question.
+
+If there are both CDS RRs and CDNSKEY RRs in the zone they must match
+in content ([RFC 7344][RFC 7344, section 4], section 4). It means that
+both must be derived from the same DNSKEY or both being "delete" CDS
+and CDNSKEY.
+
+If a name server has issues covered by [Basic04] (basic name server
+issues) no messages will be outputted from this test case.
+
+It is assumed that the Child Zone has been tested by [DNSSEC15] and
+that the servers give the same responses.
+
+## Inputs
+
+* "Child Zone" - The domain name to be tested.
+
+## Summary
+* If no CDS or CDNSKEY is found, the test case is nor run.
+* [ERROR] message if there is no DNSKEY RRset.
+* [WARNING] message if a CDS record does not match any
+  DNSKEY in DNSKEY RRset (except for "delete" CDS).
+* [WARNING] message if a CDNSKEY record does not match any
+  DNSKEY in DNSKEY RRset (except for "delete" CDNSKEY).
+* [ERROR] message if CDS RRset is not signed.
+* [ERROR] message if CDNSKEY RRset is not signed.
+* [ERROR] message if CDS RRset is signed with an invalid RRSIG.
+* [ERROR] message if CDNSKEY RRset signed with an invalid RRSIG.
+* [ERROR] message if CDS RRset is signed but not by a key in DNSKEY
+  RRset.
+* [ERROR] message if CDNSKEY RRset is signed but not by a key in DNSKEY
+  RRset.
+* [WARNING] message if CDS RRset is signed by a key in the DNSKEY RRset
+  but not with one that has signed the DNSKEY RRset.
+* [WARNING] message if CDNSKEY RRset is signed by a key in the DNSKEY
+  RRset but not with one that has signed the DNSKEY RRset.
+* [INFO] message if CDS RRset have a "delete" CDS record as a single
+  record.
+* [INFO] message if CDNSKEY RRset have a "delete" CDNSKEY record as a
+  single record.
+* [ERROR] message if "delete" CDS record is mixed with normal CDS
+  record.
+* [ERROR] message if "delete" CDNSKEY record is mixed with normal
+  CDNSKEY record.
+
+## Ordered description of steps to be taken to execute the test case
+
+1.  Create a CDS query with EDNS enabled and the DO bit set for the
+    apex of the *Child Zone*.
+
+2.  Create a CDNSKEY query with EDNS enabled and the DO bit set for
+    the apex of the *Child Zone*.
+
+3.  Create a DNSKEY query with EDNS enabled and the DO bit set for
+    the apex of the *Child Zone*.
+
+4.  Retrieve all name server IP addresses for the *Child Zone* using
+    [Get-Del-NS-IPs] and [Get-Zone-NS-IPs] ("NS IP").
+
+5.  Create the following empty sets:
+    1.  Name server IP address and associated CDS RRset and its RRSIG
+        records ("CDS RRsets"). A name server IP can hold an empty 
+        RRset or no RRSIG records.
+    2.  Name server IP address and associated CDNSKEY RRset and its
+        RRSIG redords ("CDNSKEY RRsets"). A name server IP can hold an
+        empty RRset or no RRSIG records.
+    3.  Name server IP address and associated DNSKEY RRset and its 
+        RRsig records ("DNSKEY RRsets"). A name server IP can hold an empty
+        RRset or no RRSIG records.
+    4.  Name server IP address ("No DNSKEY RRset").
+    5.  Name server IP address ("Mixed Delete CDS").
+    6.  Name server IP address ("Delete CDS").
+    7.  Name server IP address and associated CDS key tag 
+        ("No Match CDS With DNSKEY").
+    8.  Name server IP address ("CDS Not Signed").
+    9.  Name server IP address and key tag 
+        ("CDS Signed Unknown DNSKEY").
+    10. Name server IP address and key tag ("CDS Invalid RRSIG").
+    11. Name server IP address and key tag
+        ("CDS RRSIG Not Signed DNSKEY").
+    12. Name server IP address ("Mixed Delete CDNSKEY").
+    13. Name server IP address ("Delete CDNSKEY").
+    14. Name server IP address and associated CDNSKEY key tag 
+        ("No Match CDNSKEY With DNSKEY").
+    15. Name server IP address ("CDNSKEY Not Signed").
+    16. Name server IP address and key tag 
+        ("CDNSKEY Signed Unknown DNSKEY").
+    17. Name server IP address and key tag ("CDNSKEY Invalid RRSIG").
+    18. Name server IP address and key tag
+        ("CDNSKEY RRSIG Not Signed DNSKEY").
+
+6.  Repeat the following steps for each name server IP address in 
+    *NS IP*:
+
+    1. Send the CDS query over UDP to the name server IP address.
+       1. If no DNS response is returned, then go to next name server 
+          IP.
+       2. Else, if AA bit is not set in the DNS response, then go to
+          next name server IP.
+       3. Else, if the RCODE in the DNS response is not *NOERROR*, then
+          go to next name server IP.
+       4. Else, if the DNS response contains at least one CDS record
+          in the answer section, then add the name server IP and the
+          CDS RRset to the *CDS RRsets* set. Also include any assciated
+          RRSIG records.
+    2. Send the CDNSKEY query over UDP to the name server IP address.
+       1. If no DNS response is returned, then go to next name server
+          IP.
+       2. Else, if AA bit is not set in the DNS response, then go to
+          next name server IP.
+       3. Else, if the RCODE in the DNS response is not *NOERROR*, then
+          go to next name server IP.
+       4. Else, if the DNS response contains at least one CDNSKEY
+          record in the answer section, then add the name server IP and
+          the CDNSKEY RRset from the answer section to the 
+          *CDNSKEY RRsets* set. Also include any assciated RRSIG records.
+    3. Send the DNSKEY query over UDP to the name server IP address.
+       1. If no DNS response is returned, then go to next name server
+          IP.
+       2. Else, if AA bit is not set in the DNS response, then go to
+          next name server IP.
+       3. Else, if the RCODE in the DNS response is not *NOERROR*, then
+          go to next name server IP.
+       4. Else, if the DNS response contains at least one DNSKEY
+          record in the answer section, then add the name server IP and
+          the CDNSKEY RRset from the answer section to the 
+          *DNSKEY RRsets* set. Also include any assciated RRSIG records.
+    4. Go to next name server IP.
+
+7.  If neither of the *CDS RRsets* and *CDNSKEY RRsets* sets,
+    respectively, has any RRset then terminate this test case.
+
+8.  For each name server IP in the *CDS RRsets* set do:
+
+    1. Extract the CDS and its RRSIG records, if any.
+    2. If any CDS record is a "delete" CDS, then do:
+       1. If there is more than a single CDS record then add the name
+          server IP to the *Mixed Delete CDS* set.
+       2. Else, add the name server IP address to the *Delete CDS*
+          set.
+       2. Go to next name server IP.
+    3. Extract the DNSKEY and its RRSIG records, if any, from the 
+       *DNSKEY RRsets* for the same name server IP.
+    4. If there are no DNSKEY records, then do:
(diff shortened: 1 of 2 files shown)


docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,340 @@
...
+
+It is assumed that the Child Zone has been tested by [DNSSEC15] and
+that the servers give the same responses.
+
+## Inputs
+
+* "Child Zone" - The domain name to be tested.
+
+## Summary
+* If no CDS or CDNSKEY is found, the test case is nor run.
```
ulrichwisser, line 36: nor -> not

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,340 @@
...
+* [ERROR] message if there is no DNSKEY RRset.
+* [WARNING] message if a CDS record does not match any
+  DNSKEY in DNSKEY RRset (except for "delete" CDS).
+* [WARNING] message if a CDNSKEY record does not match any
+  DNSKEY in DNSKEY RRset (except for "delete" CDNSKEY).
```
ulrichwisser, line 41: Add test: [ERROR] if no CDS/CDNSKEY record matches any DNSKEY record.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,340 @@
+# DNSSEC16: Validate CDS and CDNSKEY
+
+## Test case identifier
+**DNSSEC16**
+
+## Objective
+
+CDS and CDNSKEY record types are defined in [RFC 7344] and [RFC 8878].
```
ulrichwisser, line 8: RFC 8078

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,340 @@
...
+* [ERROR] message if CDS RRset is signed with an invalid RRSIG.
+* [ERROR] message if CDNSKEY RRset signed with an invalid RRSIG.
+* [ERROR] message if CDS RRset is signed but not by a key in DNSKEY
+  RRset.
+* [ERROR] message if CDNSKEY RRset is signed but not by a key in DNSKEY
+  RRset.
+* [WARNING] message if CDS RRset is signed by a key in the DNSKEY RRset
+  but not with one that has signed the DNSKEY RRset.
+* [WARNING] message if CDNSKEY RRset is signed by a key in the DNSKEY
+  RRset but not with one that has signed the DNSKEY RRset.
```
ulrichwisser, line 53: Actually RFC7344 says "MUST be signed with a key that is represented in both the current DNSKEY and DS RRsets,"

But that only applies if a DS set exists.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
matsduf: Fixed.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
matsduf: Fixed.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
matsduf: And that could be the case here.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
matsduf: There will be warning messages for every CDS and CDNSKEY record. That should be enough, shouldn't it?

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -47,10 +49,6 @@ that the servers give the same responses.
   RRset.
 * [ERROR] message if CDNSKEY RRset is signed but not by a key in DNSKEY
   RRset.
-* [WARNING] message if CDS RRset is signed by a key in the DNSKEY RRset
-  but not with one that has signed the DNSKEY RRset.
-* [WARNING] message if CDNSKEY RRset is signed by a key in the DNSKEY
-  RRset but not with one that has signed the DNSKEY RRset.
```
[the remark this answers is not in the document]
matsduf: In this test case the DS records are not considered. That comes in DNSSEC17 (#941).

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,340 @@
...
+       2. Else, add the name server IP address to the *Delete CDS*
+          set.
+       2. Go to next name server IP.
+    3. Extract the DNSKEY and its RRSIG records, if any, from the 
+       *DNSKEY RRsets* for the same name server IP.
+    4. If there are no DNSKEY records, then do:
+       1. Add name server IP address to the *No DNSKEY RRset* set.
+       2. Go to next name server IP.
+    5. Repeat the following steps for each CDS record:
+       1. Compare the key tag from the CDS record with the caculated
```
mattias-p, line 163: `caculated` -> `calculated`

mattias-p (commented):
I found a bunch of spelling errors and minor things.

But I really like how you approach this, despite the sheer size of it.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,340 @@
...
+    1. Send the CDS query over UDP to the name server IP address.
+       1. If no DNS response is returned, then go to next name server 
+          IP.
+       2. Else, if AA bit is not set in the DNS response, then go to
+          next name server IP.
+       3. Else, if the RCODE in the DNS response is not *NOERROR*, then
+          go to next name server IP.
+       4. Else, if the DNS response contains at least one CDS record
+          in the answer section, then add the name server IP and the
+          CDS RRset to the *CDS RRsets* set. Also include any assciated
```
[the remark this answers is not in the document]
matsduf: Will fix.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,340 @@
...
+          the CDNSKEY RRset from the answer section to the 
+          *DNSKEY RRsets* set. Also include any assciated RRSIG records.
+    4. Go to next name server IP.
+
+7.  If neither of the *CDS RRsets* and *CDNSKEY RRsets* sets,
+    respectively, has any RRset then terminate this test case.
+
+8.  For each name server IP in the *CDS RRsets* set do:
+
+    1. Extract the CDS and its RRSIG records, if any.
```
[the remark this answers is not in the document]
matsduf: Will fix.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,340 @@
...
+    2. If any CDS record is a "delete" CDS, then do:
+       1. If there is more than a single CDS record then add the name
+          server IP to the *Mixed Delete CDS* set.
+       2. Else, add the name server IP address to the *Delete CDS*
+          set.
+       2. Go to next name server IP.
```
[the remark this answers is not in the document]
matsduf: Will fix.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
matsduf: Will fix.

matsduf:
I merged the table in "Outcome(s)" into the summary to make it more comprehensible on suggestion by @mattias-p. Some message tags and sets have been renamed. Also old methods are used, not MethodsNT.

@mattias-p, @ulrichwisser, @vlevigneron and @romu42, please rereview!

matsduf:
@sandoche2k, I have updated the format and introduced a summary section. It replaces the table in "Outcome(s)". Can you review?

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,323 @@
...
+          *DNSKEY Not Signed By CDS* set.
+    6. If there are no RRSIG records for the CDS RRset, then add the
+       name server IP address to the *CDS Not Signed* set.
+    7. Else, for each RRSIG (CDS) do:
+       1. If the key tag of the RRSIG does not match any DNSKEY then
+          add the name server IP address and key tag to the
+          *CDS Signed By Unknown DNSKEY* set.
+       2. Else, if the RRSIG cannot be validated by the DNSKEY it
+          refers to by key tag, then add the name server IP and RRSIG
+          key tag to the *CDS Invalid RRSIG* set.
```
ulrichwisser, line 184: 6/7 must be done for CDS delete, too.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,323 @@
...
+          *DNSKEY Not Signed By CDNSKEY* set.
+    6. If there are no RRSIG records for the CDNSKEY RRset, then add
+       the name server IP address to the *CDNSKEY Not Signed* set.
+    7. Else, for each RRSIG (CDNSKEY) do:
+       1. If the key tag of the RRSIG does not match any DNSKEY then
+          add the name server IP address and key tag to the
+          *CDNSKEY Signed By Unknown DNSKEY* set.
+       2. Else, if the RRSIG cannot be validated by the DNSKEY it
+          refers to by key tag, then add the name server IP and RRSIG
+          key tag to the *CDNSKEY Invalid RRSIG* set.
```
ulrichwisser, line 219: Same as above. 6/7 needs to be done for CDS delete, too

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,323 @@
...
+    8. Go to next name server IP address.
+
+10. If the *No DNSKEY RRset* set is non-empty, then output
+    *[DS16_CDS_CDNSKEY_WITHOUT_DNSKEY]* with all name server IP addresses
+    in the set.
```
ulrichwisser, line 224: Positive statement "if all DNSKEY RRsets are empty ....". Much easier to understand

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,323 @@
+# DNSSEC16: Validate CDS and CDNSKEY
+
+## Test case identifier
+**DNSSEC16**
+
+## Objective
+
+CDS and CDNSKEY record types are defined in [RFC 7344] and [RFC 8078].
+Both record types are optional in a zone. The objective of this test
+case is to verify that they are are valid. This Test case is
```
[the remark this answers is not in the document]
matsduf: Will update.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,323 @@
...
+only relevant if the zone has either CDS or CDNSKEY record or both.
+
+If a CDS record is included in the zone, the corresponding CDNSKEY
+record should also be included ([RFC 7344][RFC 7344, section 4],
+section 4).
+
+The CNS and CDNSKEY RRsets should be consistent between all name
```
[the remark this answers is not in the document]
matsduf: Will be resolved by removing the paragraph not relevant here.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,323 @@
...
+       3. Else, if the RCODE in the DNS response is not *NOERROR*, then
+          go to next name server IP.
+       4. Else, if the DNS response contains at least one DNSKEY
+          record in the answer section, then add the name server IP and
+          the CDNSKEY RRset from the answer section to the 
+          *DNSKEY RRsets* set. Also include any associated RRSIG records.
+    4. Go to next name server IP.
+
+7.  If neither of the *CDS RRsets* and *CDNSKEY RRsets* sets,
+    respectively, has any RRset then terminate this test case.
```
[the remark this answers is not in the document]
matsduf: No, not as the steps are defined now. I does help when you read or run it manually since you can see directly that you can break.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,323 @@
...
+
+8.  For each name server IP in the *CDS RRsets* set do:
+
+    1. Get the CDS and its RRSIG records, if any.
```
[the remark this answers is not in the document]
matsduf: It should rather be:
```
1. If the CDS RRset is non-empty, get the RRset and its RRSIG
   records, else go to next name server IP address.
```

I will update.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,323 @@
...
+    2. If any CDS record is a "delete" CDS, then do:
+       1. If there is more than a single CDS record then add the name
+          server IP to the *Mixed Delete CDS* set.
+       2. Else, add the name server IP address to the *Delete CDS*
+          set.
+       3. Go to next name server IP.
+    3. Get the DNSKEY and its RRSIG records, if any, from the
```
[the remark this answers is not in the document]
matsduf: As above.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,323 @@
...
+       *DNSKEY RRsets* for the same name server IP.
+    4. If there are no DNSKEY records, then do:
+       1. Add name server IP address to the *No DNSKEY RRset* set.
+       2. Go to next name server IP.
+    5. Repeat the following steps for each CDS record:
+       1. Compare the key tag from the CDS record with the calculated
+          key tags for the DNSKEY records.
+       2. If the CDS record does not match any DNSKEY record then add
+          the name server IP address and key tag to 
+          *No Match CDS With DNSKEY* set.
```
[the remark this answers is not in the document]
matsduf: Will fix.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
matsduf: It is, but further down.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
matsduf: It is, but further up.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
matsduf: The *No DNSKEY RRset* set is defined as a set of IP addresses.

matsduf:
@romu42, @mattias-p, @ulrichwisser and @vlevigneron, DNSSEC16 was too large. It has now been split up into two test cases, one for CDS and one for CDNSKEY. Please rereview.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,220 @@
...
+
+    1. If the CDS RRset is non-empty, get the RRset and its RRSIG
+       records, else go to next name server IP address.
+    2. If any CDS record is a "delete" CDS, then do:
+       1. If there is more than a single CDS record then add the name
+          server IP to the *Mixed Delete CDS* set.
+       2. Else, add the name server IP address to the *Delete CDS*
+          set.
+       3. Go to next name server IP.
+    3. Get the DNSKEY and its RRSIG records, if any, from the
```
mattias-p, line 117: What's the meaning of this "if any" clause again? Does it mean any DNSKEY and RRSIG records, or any RRSIG records for this DNSKEY since that set may be absent? Could it be present but empty? Anyway, what happens if the "if any" clause is false?

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,220 @@
...
+    non-empty, then output *[DS16_MIXED_DELETE_CDS]* with all
+    name server IP addresses in the set.
+
+10. If the *Delete CDS* set is non-empty, then output
+    *[DS16_DELETE_CDS]* with all name server IP addresses.
+
+11. If the *No Match CDS With DNSKEY* set is non-empty then for each
+    key tag in the set output *[DS16_CDS_MATCHES_NO_DNSKEY]* with the
+    CDS key tag and the name server IP addresses in the set per key
+    tag.
```
mattias-p, line 157: This could be simplified without changing the behavior, right? It'd be shorter, less complex and easier to parse.
`If the *No Match CDS With DNSKEY* set is non-empty then for each key tag in the set output`
->
`For each key tag in the *No Match CDS With DNSKEY* output`

Also, first it talks about "the key tag", then "the CDS key tag" and then "the key tag" again. Those three are all the same key tag, right?

What's going on with the tag arguments? It looks pretty complicated. Is the problem that you add them per name server IP but here you want to get them per key tag?

This goes for steps 12 and 15 as well.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,220 @@
...
+    3. Get the DNSKEY and its RRSIG records, if any, from the
+       *DNSKEY RRsets* for the same name server IP.
+    4. If there are no DNSKEY records, then do:
+       1. Add name server IP address to the *No DNSKEY RRset* set.
+       2. Go to next name server IP.
+    5. Repeat the following steps for each CDS record:
+       1. Compare the key tag from the CDS record with the calculated
+          key tags for the DNSKEY records.
+       2. If the CDS record does not match any DNSKEY record then add
+          the name server IP address and key tag to the
```
mattias-p, line 126: Maybe clarify here that it is the CDS key tag.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,220 @@
...
+13. If the *CDS Not Signed* set is non-empty then output 
+    *[DS16_CDS_UNSIGNED]* with all name server IP addresses in the set.
+
+14. If the *CDS Signed By Unknown DNSKEY* set is non-empty then output
+    *[DS16_CDS_SIGNED_BY_UNKNOWN_DNSKEY]* with the name server IP
+    addresses in the set.
+
+15. If the *CDS Invalid RRSIG* set is non-empty then for each key tag
+    in the set output *[DS16_CDS_INVALID_RRSIG]* with the RRSIG key
+    tag and the name server IP addresses in the set per key tag.
```
mattias-p, line 173: Have you considered moving this up to just after step 12? That way steps with the same shape come just after each other.

mattias-p (commented):
I have only looked at DNSSEC16.

matsduf:
@mattias-p, I applied your comments in the same way on both DNSSEC16 and DNSSEC17.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,220 @@
...
+Message Tag outputted                | [Default level] | Description of when message tag is outputted
+:------------------------------------|:--------|:-----------------------------------------
+DS16_CDS_UNSIGNED                    | ERROR   | CDS RRset is not signed.
+DS16_CDS_WITHOUT_DNSKEY              | ERROR   | CDS RRset exists, but no DNSKEY RRset.
+DS16_CDS_INVALID_RRSIG               | ERROR   | CDS RRset is signed with an invalid RRSIG.
+DS16_CDS_MATCHES_NO_DNSKEY           | WARNING | CDS record does not match any DNSKEY in DNSKEY RRset.
+DS16_CDS_SIGNED_BY_UNKNOWN_DNSKEY    | ERROR   | CDS RRset is signed but not by a key in DNSKEY RRset.
+DS16_DELETE_CDS                      | INFO    | CDS RRset has a "delete" CDS record as a single record.
+DS16_DNSKEY_NOT_SIGNED_BY_CDS        | WARNING | DNSKEY RRset is not signed by the key or keys that the CDS records point to.
+DS16_MIXED_DELETE_CDS                | ERROR   | "Delete" CDS record is mixed with normal CDS record.
```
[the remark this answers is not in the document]
mattias-p: Okay, that sounds reasonable.

docs/specifications/tests/DNSSEC-TP/dnssec17.md
```diff
@@ -0,0 +1,225 @@
+# DNSSEC16: Validate CDNSKEY
```
mattias-p, line 1: `DNSSEC16` -> `DNSSEC17`

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,224 @@
+# DNSSEC16: Validate CDS
+
+## Test case identifier
+**DNSSEC16**
+
+## Objective
+
+CDS and CDNSKEY record types are defined in [RFC 7344] and [RFC 8078].
```
mattias-p, line 8: Should we really be speaking of CDNSKEY in the objective section of DNSSEC16? It makes sense to mention it in the scope section. But having it in the Objective section makes it look like this test case also concerns itself with CDNSKEY.

And conversely in DNSSEC17.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,220 @@
...
+
+2.  Create a DNSKEY query with EDNS enabled and the DO bit set for
+    the apex of the *Child Zone*.
+
+3.  Retrieve all name server IP addresses for the *Child Zone* using
+    [Method4] and [Method5] ("NS IP").
+
+4.  Create the following empty sets:
+    1.  Name server IP address and associated CDS RRset and its RRSIG
+        records ("CDS RRsets"). RRSIG records may be absent.
```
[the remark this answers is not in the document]
mattias-p: Yes. My suggestion is to make it so that the RRSIG RRset is required, but that it may be empty. That way you only have one kind of absence to think about when you consider the state space that can be represented within the defined data structures. Having a smaller state space makes to consider makes the steps below easier to understand.

As authors of the test specifications we may know that the steps don't make use of the possibility for the RRSIG RRset to be present but empty. But for someone working through the test case who doesn't understand it yet, this distinction is another thing to keep in mind. And AFAICT for on good reason.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
mattias-p: Okay, so "if any" isn't about control flow. It's only there to let the reader know that they shouldn't worry even if there is no RRSIG RRset associated with the name server IP. Right?

How do you feel about this wording?

```
Get the DNSKEY and its RRSIG records from the *DNSKEY RRsets* for the same name server IP. The set of RRSIG records may be empty.
```

To me that would be much better.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
matsduf: I think it is. They are very connected. It says "The objective of this test
case is to verify that the CDS RRset is valid" and "For tests of the
CDNSKEY, see test case [DNSSEC17]".

docs/specifications/tests/DNSSEC-TP/dnssec16.md
matsduf: It is ok. I have done change.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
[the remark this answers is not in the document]
mattias-p: Oh, that's interesting! You're thinking of an RRset as something different than a set of records of the same record type? Is there a meaningful distinction?

In any case the empty set is a very simple yet powerful idea that makes many things easier.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -0,0 +1,225 @@
...
+          set.
+       3. Go to next name server IP.
+    3. Get the DNSKEY and its RRSIG records from the *DNSKEY RRsets* for the same
+       name server IP. The set of RRSIG records may be empty.
+    4. If there are no DNSKEY records, then do:
+       1. Add name server IP address to the *No DNSKEY RRset* set.
+       2. Go to next name server IP.
+    5. Repeat the following steps for each CDS record:
+       1. Compare the key tag from the CDS record with the calculated
+          key tags for the DNSKEY records.
```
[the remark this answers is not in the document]
matsduf: No, the key tag is calculated from the DNSKEY RR. If you run this manually, `dig` can provide it to you. `ldns` provides a method to do it.

Let us save that possible change to the next iteration.

PR #941, opened 2021-04-23
Create DNSSEC17 (Validate trust from DS to CDS and CDNSKEY)

matsduf:
Also see #938 (PR to create DNSSEC15), #939 (PR to create DNSSEC16) and [New-Test-Cases-for-CDS-CDNSKEY] (a temporary document).

[New-Test-Cases-for-CDS-CDNSKEY]: https://github.com/matsduf/zonemaster/blob/notes-for-new-DNSSEC-test-cases-for-CDS-CDNSKEY/docs/specifications/tests/DNSSEC-TP/New-Test-Cases-for-CDS-CDNSKEY.md

### DIFF
--- a/docs/specifications/tests/DNSSEC-TP/dnssec17.md
+++ b/docs/specifications/tests/DNSSEC-TP/dnssec17.md
@@ -1,0 +1,213 @@
+# DNSSEC17: Validate trust from DS to CDS and CDNSKEY
+
+## Test case identifier
+**DNSSEC17**
+
+## Objective
+
+CDS and CDNSKEY record types are defined in [RFC 7344] and [RFC 8078].
+Both record types are optional in a zone. The objective of this test
+case is to verify that they are there is a correct chain of trust
+from DS, in the parent zone to the CDS and CDNSKEY RRsets
+([RFC 7344][RFC 7344, section 4.1], section 4.1).
+
+This Test case is only relevant if 
+* The *Child Zone* has either CDS or CDNSKEY record or both.
+* The parent zone has a DS RRset for the *Child Zone*.
+
+If a name server has issues covered by [Basic04] (basic name server
+issues) no messages will be outputted from this test case.
+
+It is assumed that the Child Zone has been tested by [DNSSEC15] and
+[DNSSEC16] and that the servers still give the same responses.
+
+## Inputs
+
+* "Child Zone" - The domain name to be tested.
+* "Test Type" - The test type with value "undelegated" or "normal".
+* "Undelegated DS" - The DS record or records submitted (only if
+  Test Type is undelegated).
+
+## Summary
+* If no CDS or CDNSKEY records are found, this test case is not run.
+* If no DS records are found at parent, this test case is not run.
+* [ERROR] message if the CDS RRset is not signed with the DNSKEY that
+  a DS record points at.
+* [ERROR] message if the CDNSKEY RRset is not signed with the DNSKEY
+  that a DS record points at.
+
+## Ordered description of steps to be taken to execute the test case
+
+1.  Create a CDS query with EDNS enabled and the DO bit set for the
+    apex of the *Child Zone*.
+
+2.  Create a CDNSKEY query with EDNS enabled and the DO bit set for
+    the apex of the *Child Zone*.
+
+3.  Create a DNSKEY query with EDNS enabled and the DO bit set for
+    the apex of the *Child Zone*.
+
+4.  Create a DS query with EDNS enabled and DO flag set for the name of
+    the *Child Zone*.
+
+5.  Create the following empty sets:
+    1.  Name server IP address and associated CDS RRset and its RRSIG
+        records ("CDS RRsets"). A name server IP can hold an empty 
+        RRset or no RRSIG records.
+    2.  Name server IP address and associated CDNSKEY RRset and its
+        RRSIG redords ("CDNSKEY RRsets"). A name server IP can hold an
+        empty RRset or no RRSIG records.
+    3.  Name server IP address and associated DNSKEY RRset
+        ("DNSKEY RRsets"). A name server IP can hold an empty RRset.
+    4.  DS record set ("DS Records").
+    5.  Name server IP and DS record key tag ("DS No Match CDS RRSIG").
+    6.  Name server IP and DS record key tag 
+        ("DS No Match CDNSKEY RRSIG").
+
+6. If the *Test Type* is "undelegated, then:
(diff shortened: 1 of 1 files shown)


docs/specifications/tests/DNSSEC-TP/dnssec17.md
```diff
@@ -0,0 +1,213 @@
...
+* "Undelegated DS" - The DS record or records submitted (only if
+  Test Type is undelegated).
+
+## Summary
+* If no CDS or CDNSKEY records are found, this test case is not run.
+* If no DS records are found at parent, this test case is not run.
+* [ERROR] message if the CDS RRset is not signed with the DNSKEY that
+  a DS record points at.
+* [ERROR] message if the CDNSKEY RRset is not signed with the DNSKEY
+  that a DS record points at.
```
ulrichwisser, line 37: not "the DNSKEY" but "a DNSKEY". The DS set can point out more than one key.

docs/specifications/tests/DNSSEC-TP/dnssec17.md
```diff
@@ -0,0 +1,213 @@
...
+       server IP.
+    3. For each DS record in *DS Records* do:
+       1. If the DS record does not point at a DNSKEY record then add
+          name server IP and DS key tag to the 
+          *DS No Match CDNSKEY RRSIG*.
+       2. Else, if the DNSKEY that the DS record points at does not
+          match any RRSIG for CDNSKEY RRset then add name server IP and
+          DS key tag to the *DS No Match CDNSKEY RRSIG*.
+    4. Go to next name server IP address.
+
```
ulrichwisser, line 159: Overall question: How is "match" defined for keys? Only matching alg/keytag or are signatures actually verified?

docs/specifications/tests/DNSSEC-TP/dnssec17.md
matsduf: Yes, but the text is "a DS record points at", so it is per DS record.

matsduf: It is meant to be "the DNSKEY record". I can clarify.

docs/specifications/tests/DNSSEC-TP/dnssec17.md
matsduf: Presently is probably only verifying key tag.

docs/specifications/tests/DNSSEC-TP/dnssec17.md
```diff
@@ -0,0 +1,213 @@
...
+          go to next name server IP.
+       4. Else, if the DNS response contains at least one DNSKEY
+          record in the answer section, then add the name server IP and
+          the CDNSKEY RRset from the answer section to the 
+          *DNSKEY RRsets* set.
+    4. Go to next name server IP.
+
+11. If neither of the *CDS RRsets* and *CDNSKEY RRsets* sets,
+    respectively, has any RRset then terminate this test case.
+
```
ulrichwisser, line 129: If  all *CDS RRsets* and *CDNSKEY RRsets* are empty then terminate this test case.

docs/specifications/tests/DNSSEC-TP/dnssec17.md
ulrichwisser: I think this should be validate.

docs/specifications/tests/DNSSEC-TP/dnssec17.md
```diff
@@ -0,0 +1,213 @@
...
+       server IP.
+    3. For each DS record in *DS Records* do:
+       1. If the DS record does not point at a DNSKEY record then add
+          name server IP and DS key tag to the 
+          *DS No Match CDS RRSIG*.
+       2. Else, if the DNSKEY that the DS record points at does not
+          match any RRSIG for CDS RRset then add name server IP and
+          DS key tag to the *DS No Match CDS RRSIG*.
+    4. Go to next name server IP address.
+
```
ulrichwisser, line 145: This test will create false negatives.
Only one of the DS records has to match the RRSIG, not all.
Same applies below for CDNSKEY.

docs/specifications/tests/DNSSEC-TP/dnssec17.md
matsduf: Will update wording.

docs/specifications/tests/DNSSEC-TP/dnssec17.md
matsduf: > A DS RR SHOULD point to a DNSKEY RR that is present in the child's
> apex DNSKEY RRset, and the child's apex DNSKEY RRset SHOULD be
> signed by the corresponding private key."

https://tools.ietf.org/html/rfc4035#section-2.4

docs/specifications/tests/DNSSEC-TP/dnssec17.md
matsduf: I agree, and that is an improvement that maybe can be included for this release for this test case.

matsduf:
@romu42, @ulrichwisser, @mattias-p and @vlevigneron, please rereview.

docs/specifications/tests/DNSSEC-TP/dnssec17.md
ulrichwisser: Your code tests if all DS records point to an RRSIG.
RFC says "A DS" needs to point to the DNSKEY, sign the dnskey set and the cds set.
Your need to test backwards:
For all RRSIG of the CDS set check that there is an RRSIG with the same key on the dnskey set (from the same server ip) and check that there is a DS that points to the same key.

matsduf: My interpretation is that the CDS RRset should be signed by the private key of every DNSKEY that any DS points to.

ulrichwisser: This test case doesn't test 4035, but https://tools.ietf.org/html/rfc7344#4.1
BTW, I haven't seen a test for continuity.

matsduf: > This test case doesn't test 4035, but https://tools.ietf.org/html/rfc7344#4.1

True, and "with a key" could be interpreted as "with at least one key". I will rework.

> BTW, I haven't seen a test for continuity.

That is part of the proposal for DNSSEC16 and DNSSEC17 in #939.

matsduf:
@ulrichwisser, please re-review after my update.

matsduf:
@romu42, please re-review.

PR #945, opened 2021-05-11
Codify test planning

mattias-p:
## Purpose

In preparation for v2020.2 we decided to update the format of our release testing. We went through with the release roughly according to those plans. Then we had a postmortem meeting where we concluded that the approach we used was a big improvement over how we did things before. This PR aims to codify our conclusions on test planning from that postmortem.

## Scope

I've tried to keep the document light on details focusing on the conclusions we made. I've tried to leave out other details of how we did things. Not because they were bad but because they weren't explicitly highlighted as reasons for success. Also, it should be said that I wrote this document according to my own notes that I made at the time, and that was review by @matsduf without remarks shortly after the meeting.

## Changes

I've added a new test planning phase in the release process. Initially I considered integrating the test planning in the preparation phase, but it doesn't fit there for two reasons. First the preparation phase is defined to optionally take place before the end of development, and second the preparation steps are all performed as pull request but the test planning is performed as a meeting.

### DIFF
--- a/docs/internal-documentation/maintenance/ReleaseProcess-test-planning.md
+++ b/docs/internal-documentation/maintenance/ReleaseProcess-test-planning.md
@@ -7,7 +7,7 @@ This gives us greater faith in our releases.
 
 ## Release testing start up-meeting
 
-1. Compile a list of all pull requests with user-facing changes.
+1. Compile a list of all pull requests that should be tested.
 
 2. For each test item, discuss how it could and should be tested.
    Make sure there is a heading titled "How to test this PR" in the pull request
--- a/docs/internal-documentation/maintenance/ReleaseProcess-test-planning.md
+++ b/docs/internal-documentation/maintenance/ReleaseProcess-test-planning.md
@@ -1,0 +1,18 @@
+Release process - Test Planning
+===============================
+
+The purpose of test planning is to raise the quality by making the release
+testing more structured, getting better coverage and finding more bugs.
+This gives us greater faith in our releases.
+
+## Release testing start up-meeting
+
+1. Compile a list of all pull requests with user-facing changes.
+
+2. For each test item, discuss how it could and should be tested.
+   Make sure there is a heading titled "How to test this PR" in the pull request
+   description.
+   Add/update notes under this heading to reflect the conclusions made.
+
+3. Discuss prioritization of the test items and who should test what.
+   Note down all the conclusions.
--- a/docs/internal-documentation/maintenance/ReleaseProcess-test-planning.md
+++ b/docs/internal-documentation/maintenance/ReleaseProcess-test-planning.md
@@ -12,6 +12,7 @@ This gives us greater faith in our releases.
 2. For each test item, discuss how it could and should be tested.
    Make sure there is a heading titled "How to test this PR" in the pull request
    description.
+   Add it if needed.
    Add/update notes under this heading to reflect the conclusions made.
 
 3. Discuss prioritization of the test items and who should test what.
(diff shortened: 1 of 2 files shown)


docs/internal-documentation/maintenance/ReleaseProcess-test-planning.md
```diff
@@ -0,0 +1,18 @@
+Release process - Test Planning
+===============================
+
+The purpose of test planning is to raise the quality by making the release
+testing more structured, getting better coverage and finding more bugs.
```
matsduf, line 5: It does not have to be bugs (software errors), it could be spelling errors or other things. And there might not be any bugs or issues to find. I think "any issues" is better than "more bugs".

docs/internal-documentation/maintenance/ReleaseProcess-test-planning.md
```diff
@@ -0,0 +1,18 @@
...
+This gives us greater faith in our releases.
+
+## Release testing start up-meeting
+
+1. Compile a list of all pull requests with user-facing changes.
+
+2. For each test item, discuss how it could and should be tested.
+   Make sure there is a heading titled "How to test this PR" in the pull request
+   description.
```
matsduf, line 14: If not, should that be added in this process? If so, state that.

docs/internal-documentation/maintenance/ReleaseProcess.md
```diff
@@ -11,20 +11,27 @@ process, definitely before acceptance testing starts, but also before
 development has been completed.
 
 
-## 2. [Create test distributions]
+## 2. [Test planning]
+
+This phase should be performed after the end of development but before testing
+process, definitely before acceptance testing starts, but also before
+development has been completed.
```
matsduf, line 18: "after the end of development" and "before development has been completed". Should it be done twice? If so, state it in the right order, and in two different bullets.

matsduf (commented):
I think ReleaseProcess-test-planning.md could expanded and be more detailed.

docs/internal-documentation/maintenance/ReleaseProcess-test-planning.md
mattias-p: I see your point. But I really like to have the word "more" in there to reflect that we would find some problems with unstructured testing, but we find more of them with structured testing. How do you feel about "more problems" instead of "more bugs"?

matsduf: Keep more.

docs/internal-documentation/maintenance/ReleaseProcess.md
mattias-p: Something is very wrong here. I'll make an update.

docs/internal-documentation/maintenance/ReleaseProcess-test-planning.md
mattias-p: IMHO we could deal with that situation outside of this process and continue this process once the situation has been dealt with.

mattias-p:
> I think ReleaseProcess-test-planning.md could expanded and be more detailed.

Yeah, I agree there's room for further improvement. In this PR I just want to tie up some loose ends from the last round of testing. Do you think I missed something important we concluded in the postmortem we had for last release?

docs/internal-documentation/maintenance/ReleaseProcess.md
mattias-p: I fixed it.

matsduf (dismissed):

pnax (dismissed):
edit: already done

~In the `Release Process` document, please update the next sentence either by removing the number of sections or updating it to 4.~

> ~This process has been split into three parts which are described in their own documents.~

pnax (approved):

PR #950, opened 2021-05-25
Updates CONSISTENCY specifications to lower level of no response

matsduf:
BASIC04 has been implemented to report on problem with responses from nameservers. The Consistency specifications and their implementation, repeats reporting the same thing with no benefit. With this PR the Consistency test case will refer that to BASIC04.

A PR to lower the level in the profile is created in Zonemaster-Engine (https://github.com/zonemaster/zonemaster-engine/pull/923).

* Adds reference to BASIC04, which tests all name servers for non-response.
* Lower default level to DEBUG for messages that report on non-response.

### DIFF
--- a/docs/specifications/tests/Consistency-TP/consistency01.md
+++ b/docs/specifications/tests/Consistency-TP/consistency01.md
@@ -122,28 +122,18 @@ When comparing SOA serial it must be done using the arithmetic defined in
 
 None
 
+
 [Basic04]:                    ../Basic-TP/basic04.md
-[RFC 1034]: https://tools.ietf.org/html/rfc1035
-
-[RFC 1035]: https://tools.ietf.org/html/rfc1035
-
-[RFC 1982]: https://tools.ietf.org/html/rfc1982 
-
-[Method4]: ../Methods.md#method-4-obtain-glue-address-records-from-parent
-
-[Method5]: ../Methods.md#method-5-obtain-the-name-server-address-records-from-child
-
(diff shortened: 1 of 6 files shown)


mattias-p:
Do we have time to test the changes in Engine for v2021.1? I think we should consider moving this to v2021.2.

matsduf:
> Do we have time to test the changes in Engine for v2021.1? I think we should consider moving this to v2021.2.

The code changes are simple. It is just an update of profile.json (zonemaster/zonemaster-engine#923). It really improves the output and makes use of Basic04. I think we can include it.

docs/specifications/tests/Consistency-TP/consistency01.md
```diff
@@ -117,6 +122,7 @@ When comparing SOA serial it must be done using the arithmetic defined in
 
 None
 
+[Basic04]:                    ../Basic-TP/basic04.md
```
mattias-p, line 125: The indentation here is off.

There are more of these below as well as in the sibling pull requests.

mattias-p (approved):

docs/specifications/tests/Consistency-TP/consistency01.md
matsduf: When the link section was not formatted I just ignored the formatting. I have now sorted and formatted the link sections in the specifications in this PR.

@mattias-p, please re-review.

mattias-p (approved):

pnax (approved):

PR #952, opened 2021-05-25
Updates DNSSEC specifications to lower level of no response

matsduf:
BASIC04 has been implemented to report on problem with responses from nameservers. The DNSSEC specifications and their implementation, repeats reporting the same thing with no benefit. With this PR the Consistency test case will refer that to BASIC04.

A PR to lower the level in the profile will be created in Zonemaster-Engine (https://github.com/zonemaster/zonemaster-engine/pull/923).

* Adds reference to BASIC04, which tests all name servers for non-response.
* Lower default level to DEBUG for messages that report on non-response.
* ~~DNSSEC14 has not been updated. See discussion in zonemaster/zonemaster-engine#920~~
* DNSSEC14 is included.

Relates to #950 and #951

PR #955, opened 2021-05-25
Updates Zone10 specifications to lower level of no response

matsduf:
BASIC04 has been implemented to report on problem with responses from nameservers. The Zone10 specification and its implementation, repeats reporting the same thing with no benefit. With this PR the Zone10 test case will refer that to BASIC04.

A PR to lower the level in the profile is created in Zonemaster-Engine (https://github.com/zonemaster/zonemaster-engine/pull/923).

* Adds reference to BASIC04, which tests all name servers for non-response.
* Lower default level to DEBUG for messages that report on non-response.

This PR relates to #950, #951, #952, #953, #954

### DIFF
--- a/docs/specifications/tests/Zone-TP/zone10.md
+++ b/docs/specifications/tests/Zone-TP/zone10.md
@@ -62,7 +62,7 @@ MULTIPLE_SOA                  | ERROR
 NO_RESPONSE                   | DEBUG
 NO_SOA_IN_RESPONSE            | DEBUG
 ONE_SOA                       | INFO
-WRONG_SOA                     | ERROR
+WRONG_SOA                     | DEBUG
 
 
 ## Special procedural requirements
--- a/docs/specifications/tests/Zone-TP/zone10.md
+++ b/docs/specifications/tests/Zone-TP/zone10.md
@@ -13,6 +13,12 @@ This test case will verify that the zone of the domain to be tested return
 exactly one SOA record.
 
 
+## Scope
+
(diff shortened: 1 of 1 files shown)


docs/specifications/tests/Zone-TP/zone10.md
```diff
@@ -53,8 +59,8 @@ In other cases the outcome of this Test Case is "pass".
 Message                       | Default severity level
 :-----------------------------|:-----------------------------------
 MULTIPLE_SOA                  | ERROR
-NO_RESPONSE                   | WARNING
-NO_SOA_IN_RESPONSE            | WARNING
+NO_RESPONSE                   | DEBUG
+NO_SOA_IN_RESPONSE            | DEBUG
 ONE_SOA                       | INFO
 WRONG_SOA                     | ERROR
```
mattias-p, line 65: Shouldn't WRONG_SOA also be demoted to DEBUG?

mattias-p (commented):
It feels a bit like all of Zone10 should be deprecated and replaced by Basic04. (With Basic04 being extended to cover the remaining messages in Zone10.)

mattias-p:
Oh, and maybe the title should be fixed as well: `multipe` -> `multiple`.

docs/specifications/tests/Zone-TP/zone10.md
matsduf: Yes. Will fix.

mattias-p (dismissed):

pnax (approved):

PR #956, opened 2021-05-26
Removes test case DNSSEC14 since it is not implemented

matsduf:
Test case DNSSEC14 is not implemented. This RP removes it from develop branch to make it disappear from master branch at release. The test case will be added back to develop branch with a new PR. See zonemaster/zonemaster-engine#920.

The following files also lists the test case, but that will be removed at release through file generation.

./docs/specifications/tests/ImplementedTestCases.md
./docs/specifications/tests/README.md
./docs/specifications/tests/TestMessages.md


matsduf:
Wrong.

PR #959, opened 2021-05-27
Updates to refer to latest versions

matsduf:
Updates to latest versions according to https://github.com/nvm-sh/nvm and https://nodejs.org/en/download/


pnax:
> Does this mean we should repeat any testing we've done on GUI with the new tooling versions?

This is what I am currently doing. Well TBH I haven't started testing the GUI before this PR, so I started directly with this new version.

pnax (approved):
I've build the GUI with the latest version installed it and ran some tests. It seems to work.

mattias-p (approved):
I've built the GUI dist file on a Ubuntu 20.04 prepared with this instruction. Well, actually I just ran these commands on top of the old setup.

I've made installations based on a dist file build with version 0.38.0/14.17 on both CentOS 7 and 8. The smoke test passed on both installations.

PR #964, opened 2021-06-09
Updates copyright statements (zm/zm)

matsduf:
## Purpose

In the copyright statement a range of years is given. That requires updates, but year is not needed.

Nothing is said about copyright for contributors.

## Changes

Removes year from copyright and "translates" the name of Internetstiftelsen. Adds statement about contributor's copyright.

Removes copyright statement from file not needing it.

## How to test this PR

No need to test.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,8 +1,10 @@
 ### Code license
 
-Copyright (c) 2013-2020, Internetstiftelsen (<https://internetstiftelsen.se/en/>)
-Copyright (c) 2013-2020, AFNIC (<https://www.afnic.fr/en/>)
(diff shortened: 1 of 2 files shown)


pnax (approved):

mattias-p (approved):
Looks good to me.

But could you say something about why these changes are the right ones?

matsduf:
> But could you say something about why these changes are the right ones?

Today year is included which requires updates every year, but year does not add anything and is not required. To simplify, this PR removes the year.

We take the copyright, but contributors should correctly be respected, as suggested by one of the IIS lawyers.

Issue #957, opened 2021-05-26
Test specification "DNSSEC16: Validate CDS" should require SEP key for CDS RRSIG


huguei:
On step 7.vii of test specification "DNSSEC16: Validate CDS", it should be a requirement that at least one CDS RRSIG record should be made with the "KSK" of the zone, or more precisely, with a key with the SEP bit that corresponds to the DS from the parent. Not just any key.

This requirement is from RFC7344, section 4.1, where it says "Signer: MUST be signed with a key that is represented in both the
current DNSKEY **and DS RRsets**,...".

To maintain this test case autonomous from parent data, and to allowing to work for the unsigned case, I propose to at least test that there's a valid RRSIG over CDS, made with a valid key from the DNSKEY set **that has the SEP bit on**. The responsibility of ensuring that the key with the SEP bit is the same as the parent's DS, it should be from the same parent when accepting the rollover or initial enrollment, in another test or action.

So, I propose this change for 7.vii:

```
vii. Else, for each RRSIG (CDS) do:
         a. If the key tag of the RRSIG does not match any DNSKEY **with the SEP bit on,** then add the name server IP address and key tag to the CDS Signed By Unknown DNSKEY set.
         b. Else, if the RRSIG cannot be validated by the DNSKEY it refers to by key tag, then add the name server IP and RRSIG key tag to the CDS Invalid RRSIG set.

```

matsduf:
Thank you for your comment. Validation against the DS records is planned for test case DNSSEC18, for which there is draft specification in #941. Neither [DNSSEC16] (CDS) or [DNSSEC17] (CDNSKEY) look at the DS records.

According to the specifications it is mandatory to have KSK+ZSK. You could have a combined CSK instead. And also the SEP is not mandatory. You could use a DNSKEY without SEP set and use it as KSK (or CSK). That said, it is still recommended to have the SEP set on the keys that are used for DS.

I will make a proposal of an update of DNSSEC16 and DNSSEC17 to include verification if the CDS and CDNSKEY records, respecitively, point at DNSKEY records with SEP set. -- In the case of CDNSKEY that the SEP is set in itself.

[DNSSEC16]: https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/DNSSEC-TP/dnssec16.md
[DNSSEC17]: https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/DNSSEC-TP/dnssec16.md

matsduf:
[...]

PR #966, opened 2021-06-14
Adds test of DNSKEY bits in DNSSEC16 and DNSSEC17

matsduf:
## Purpose

The DNSKEY used for CDS or CDNSKEY must have bit 7 set (key for zone signing). And it should have bit 15 (SEP) set.

## Context

This PR addresses issue #957.

## Changes

Tests for bits 7 and 15 are added, and if not set a message is outputted. Those messages are added to the specification. This PR changes the specifications, not the implementation.

## How to test this PR

This PR should be reviewed before merge.

### DIFF
--- a/docs/specifications/tests/DNSSEC-TP/dnssec16.md
+++ b/docs/specifications/tests/DNSSEC-TP/dnssec16.md
@@ -36,11 +36,13 @@ give an incomplete report of the CDS and CDNSKEY status of
 
 Message Tag outputted                | [Default level] | Description of when message tag is outputted
 :------------------------------------|:--------|:-----------------------------------------
-DS16_CDS_UNSIGNED                    | ERROR   | CDS RRset is not signed.
-DS16_CDS_WITHOUT_DNSKEY              | ERROR   | CDS RRset exists, but no DNSKEY RRset.
 DS16_CDS_INVALID_RRSIG               | ERROR   | CDS RRset is signed with an invalid RRSIG.
+DS16_CDS_MATCHES_NON_SEP_DNSKEY      | NOTIFY  | CDS record matches a DNSKEY with SEP bit (bit 15) unset.
+DS16_CDS_MATCHES_NON_ZONE_DNSKEY     | ERROR   | CDS record matches a DNSKEY with zone bit (bit 7) unset.
 DS16_CDS_MATCHES_NO_DNSKEY           | WARNING | CDS record does not match any DNSKEY in DNSKEY RRset.
 DS16_CDS_SIGNED_BY_UNKNOWN_DNSKEY    | ERROR   | CDS RRset is signed but not by a key in DNSKEY RRset.
+DS16_CDS_UNSIGNED                    | ERROR   | CDS RRset is not signed.
+DS16_CDS_WITHOUT_DNSKEY              | ERROR   | CDS RRset exists, but no DNSKEY RRset.
 DS16_DELETE_CDS                      | INFO    | CDS RRset has a "delete" CDS record as a single record.
 DS16_DNSKEY_NOT_SIGNED_BY_CDS        | WARNING | DNSKEY RRset is not signed by the key or keys that the CDS records point to.
 DS16_MIXED_DELETE_CDS                | ERROR   | "Delete" CDS record is mixed with normal CDS record.
@@ -55,14 +57,18 @@ DS16_MIXED_DELETE_CDS                | ERROR   | "Delete" CDS record is mixed wi
     3.  Name server IP address ("No DNSKEY RRset").
     4.  Name server IP address ("Mixed Delete CDS").
     5.  Name server IP address ("Delete CDS").
-    6.  Name server IP address and associated CDS key tag 
+    6.  Name server IP address and associated CDS key tag
         ("No Match CDS With DNSKEY").
     7.  Name server IP address and associated CDS key tag
+        ("CDS points to non-zone DNSKEY").
+    8.  Name server IP address and associated CDS key tag
+        ("CDS points to non-SEP DNSKEY").
+    9.  Name server IP address and associated CDS key tag
         ("DNSKEY Not Signed By CDS").
-    8.  Name server IP address ("CDS Not Signed").
-    9.  Name server IP address and key tag
+    10. Name server IP address ("CDS Not Signed").
+    11. Name server IP address and key tag
         ("CDS Signed By Unknown DNSKEY").
-    10. Name server IP address and key tag ("CDS Invalid RRSIG").
+    12. Name server IP address and key tag ("CDS Invalid RRSIG").
 
 2.  Create a CDS query with EDNS enabled and the DO bit set for the
     apex of the *Child Zone*.
@@ -126,10 +132,18 @@ DS16_MIXED_DELETE_CDS                | ERROR   | "Delete" CDS record is mixed wi
        2. If the CDS record does not match any DNSKEY record then add
           the name server IP address and CDS record key tag to the
           *No Match CDS With DNSKEY* set.
-       3. Else, if there is no RRSIG for the DNSKEY RRset created by
-          the DNSKEY record that the CDS record points at then add the
-          name server IP address and key tag of CDS record to the
-          *DNSKEY Not Signed By CDS* set.
+       3. Else, if bit 7 of the flag field of the DNSKEY that the DS record
+          points to is set to 0 (nil) then add the name server IP address and
+          CDS record key tag to the *CDS points to non-zone DNSKEY* set.
+       4. Else, do:
+          *  If there is no RRSIG for the DNSKEY RRset created by the DNSKEY
+             record that the CDS record points at then add the name server IP
+             address and key tag of CDS record to the
+             *DNSKEY Not Signed By CDS* set.
+          *  If bit 15 of the flag field of the DNSKEY that the DS record
+             points to is set to 0 (nil) then add the name server IP address
+             and the key tag of the CDS record to the
+             *CDS points to non-SEP DNSKEY* set.
     6. If there are no RRSIG records for the CDS RRset, then add the
        name server IP address to the *CDS Not Signed* set.
     7. Else, for each RRSIG (CDS) do:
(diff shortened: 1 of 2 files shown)


matsduf:
@huguei, please review this PR.

huguei:
I think it should be added "The DNSKEY used for **signing** CDS or CDNSKEY...", because now it's confusing if it's the signing or the one referred by CDS itself.

matsduf:
> I think it should be added "The DNSKEY used for **signing** CDS or CDNSKEY...", because now it's confusing if it's the signing or the one referred by CDS itself.

That is the DNSKEY that the current DS record points at and that is covered by [DNSSEC02]. The key that CDS and CDNSKEY points at is the DS-to-be, so this test case should verify that key has bit 8 and 15, respectively, set. I have updated the description above.

Please review the updated specification to see if you think it meets the requirements of the RFCs.

[DNSSEC02]: https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/DNSSEC-TP/dnssec02.md

huguei:
Ok, I see. Yes, I find this test very useful, thanks!
However, it doesn't address the ticket #957 . I think it will be better addressed in test case DNSSEC16 7.vii, in the RRSIG checks. Thanks!

matsduf:
@huguei, now I have included test for "self-signing" as you request. Please review.

@ulrichwisser, @mattias-p and @pnax, please review.

huguei:
Works for our use case, thanks a lot @matsduf !

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -106,40 +115,52 @@ DS16_MIXED_DELETE_CDS                | ERROR   | "Delete" CDS record is mixed wi
...
           key tags for the DNSKEY records.
        2. If the CDS record does not match any DNSKEY record then add
           the name server IP address and CDS record key tag to the
           *No Match CDS With DNSKEY* set.
-       3. Else, if there is no RRSIG for the DNSKEY RRset created by
-          the DNSKEY record that the CDS record points at then add the
-          name server IP address and key tag of CDS record to the
-          *DNSKEY Not Signed By CDS* set.
-    6. If there are no RRSIG records for the CDS RRset, then add the
+       3. Else, if bit 7 of the flag field of the DNSKEY that the DS record
```
mattias-p, line 138: `flag field` -> `flags field`

There are more of these.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -106,40 +115,52 @@ DS16_MIXED_DELETE_CDS                | ERROR   | "Delete" CDS record is mixed wi
...
+             address and key tag of CDS record to the
+             *DNSKEY Not Signed By CDS* set.
+          *  If there is no RRSIG for the CDS RRset created by the DNSKEY
+             record that the CDS record points at then add the name server IP
+             address and key tag of CDS record to the *CDS Not Signed By CDS*
+             set.
+          *  If bit 15 of the flag field of the DNSKEY that the DS record
+             points to is set to 0 (nil) then add the name server IP address
+             and the key tag of the CDS record to the
+             *CDS points to non-SEP DNSKEY* set.
```
mattias-p, line 153: There's no corresponding step for this in DNSSEC17. Is that intentional?

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -106,40 +115,52 @@ DS16_MIXED_DELETE_CDS                | ERROR   | "Delete" CDS record is mixed wi
...
+          CDS record key tag to the *CDS points to non-zone DNSKEY* set.
+       4. Else, do:
+          *  If there is no RRSIG for the DNSKEY RRset created by the DNSKEY
+             record that the CDS record points at then add the name server IP
+             address and key tag of CDS record to the
+             *DNSKEY Not Signed By CDS* set.
+          *  If there is no RRSIG for the CDS RRset created by the DNSKEY
+             record that the CDS record points at then add the name server IP
+             address and key tag of CDS record to the *CDS Not Signed By CDS*
+             set.
```
mattias-p, line 149: There's a lot going on in these two steps. Would you consider splitting each of them up into separate steps for the conditions and action?

Also I believe the condition clauses would be easier on the reader if they were reversed. I.e. so you start out from the CDS record and end up at the RRSIG set. (I assume that the condition is whether or not the RRSIG set is empty. Is that right?)

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -106,40 +115,52 @@ DS16_MIXED_DELETE_CDS                | ERROR   | "Delete" CDS record is mixed wi
 
 7.  For each name server IP in the *CDS RRsets* set do:
 
-    1. If the CDS RRset is non-empty, get the RRset and its RRSIG
-       records, else go to next name server IP address. The set of the RRSIG
-       records may be empty.
-    2. If any CDS record is a "delete" CDS, then do:
+    1. If the CDS RRset is empty go to next name server IP address.
+    2. Get the CDS RRset and its RRSIG records from the *CDS RRsets* set
+       for the name server IP. The RRSIG records may be absent.
```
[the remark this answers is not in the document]
matsduf: Yes, it means that the RRSIG records may be absent, there might be no RRSIG records for the CDS RRset.

No, at 2 the CDS RRset is non-empty.

RRSIG records are special. They do not constitute an RRset. They are a kind of attachments to the RRset that they sign. Here the RRset is the CDS RRset, and above in 5.ii.d (as md is rendered) the RRset and the associated RRSIG records are collected. It is possible that we have an CDS RRset but no RRSIG records for the RRset. If there is no CDS RRset we do not fetch any RRSIG records for the non-existent RRset.

I do not want to call it an RRSIG set, because that would sound like a RRSIG RRset, which it is not.

If you do see it as a set, then I do not see the difference between an empty set and a absent set.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
matsduf: I found 2 instances. I will correct those.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -106,40 +115,52 @@ DS16_MIXED_DELETE_CDS                | ERROR   | "Delete" CDS record is mixed wi
...
        2. If the CDS record does not match any DNSKEY record then add
           the name server IP address and CDS record key tag to the
           *No Match CDS With DNSKEY* set.
-       3. Else, if there is no RRSIG for the DNSKEY RRset created by
-          the DNSKEY record that the CDS record points at then add the
-          name server IP address and key tag of CDS record to the
-          *DNSKEY Not Signed By CDS* set.
-    6. If there are no RRSIG records for the CDS RRset, then add the
+       3. Else, if bit 7 of the flag field of the DNSKEY that the DS record
+          points to is set to 0 (nil) then add the name server IP address and
```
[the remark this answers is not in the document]
matsduf: "Nil" is just another word for "zero".

docs/specifications/tests/DNSSEC-TP/dnssec16.md
```diff
@@ -106,40 +115,52 @@ DS16_MIXED_DELETE_CDS                | ERROR   | "Delete" CDS record is mixed wi
...
+          CDS record key tag to the *CDS points to non-zone DNSKEY* set.
+       4. Else, do:
+          *  If there is no RRSIG for the DNSKEY RRset created by the DNSKEY
```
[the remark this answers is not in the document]
matsduf: Yes, it could be easier to refer to some line, but I think it is easier to read with the bullet when it is a few levels down. I can change to ordered list to see how it looks like.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
matsduf: No, it is more complex than that. I will take a look at it.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
matsduf: Yes, look for "CDNSKEY is non-SEP key". When you have a CDS you have to find the corresponding DNSKEY to determine if the SEP is set or not. With CDNSKEY you at look at the record itself.

docs/specifications/tests/DNSSEC-TP/dnssec17.md
```diff
@@ -106,40 +115,52 @@ DS17_MIXED_DELETE_CDNSKEY            | ERROR   | "Delete" CDNSKEY record is mixe
...
+             the CDNSKEY record to the *No Match CDNSKEY With DNSKEY* set.
+          4. Else, do:
+             * If there is no RRSIG for the DNSKEY RRset created by
+               the DNSKEY record matching the CDNSKEY record then add the
+               name server IP address and key tag of CDNSKEY record to the
+               *DNSKEY Not Signed By CDNSKEY* set.
+             *  If there is no RRSIG for the CDNSKEY RRset created by the DNSKEY
+                record matching the CDNSKEY record then add the name server IP
+                address and key tag of CDS record to the
+                *CDNSKEY Not Signed By CDNSKEY* set.
```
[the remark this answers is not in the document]
matsduf: I will correct.

matsduf:
@mattias-p, I have addressed all the issues that you brought up. Please re-review.

docs/specifications/tests/DNSSEC-TP/dnssec16.md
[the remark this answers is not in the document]
mattias-p: > RRSIG records are special. They do not constitute an RRset. They are a kind of attachments to the RRset that they sign.

Thanks for explaining! I wasn't aware of that.

> If you do see it as a set, then I do not see the difference between an empty set and a absent set.

If the set is absent that means there is no set. The same way having the number zero is different from having no number and having the empty string is different not having a string.

The new wording is good from a programmer's perspective.

docs/specifications/tests/DNSSEC-TP/dnssec17.md
```diff
@@ -106,40 +115,53 @@ DS17_MIXED_DELETE_CDNSKEY            | ERROR   | "Delete" CDNSKEY record is mixe
...
-          *DNSKEY Not Signed By CDNSKEY* set.
-    6. If there are no RRSIG records for the CDNSKEY RRset, then add
-       the name server IP address to the *CDNSKEY Not Signed* set.
-    7. Else, for each RRSIG (CDNSKEY) do:
-       1. If the key tag of the RRSIG does not match any DNSKEY then
-          add the name server IP address and key tag to the
+    6. Repeat the following steps for each CDNSKEY record unless it is a "delete"
+       CDNSKEY record:
+       1. If bit 7 of the flags field of the CDNSKEY record is unset (value 0)
+          then add the name server IP address and the key tag derived from
```
mattias-p, line 135: Does the word "derived" here (and twice more below) refer to the key tag calculation described in [RFC 4034 Appendix B](https://datatracker.ietf.org/doc/html/rfc4034#appendix-B)? In that case maybe "calculated" would be a more helpful word?

mattias-p (approved):
It think this looks really good. There's still one thing that I felt isn't obvious when I read through the changes.

docs/specifications/tests/DNSSEC-TP/dnssec17.md
matsduf: Good point. I'll update.

mattias-p (approved):
I've looked at this from an editorial point of view and I think it looks really good.

matsduf:
@vlevigneron, could you review this PR?

matsduf:
@vlevigneron, can I merge this and create an issue for update of Engine?

PR #1016, opened 2021-11-22
Use latest stable version of Node

mattias-p:
## Purpose

I ran into problems when I used the Node version specified in the Ubuntu-Node.js-build-environment document. I don't remember the exact error message or when it happened but the solution was to upgrade to a newer version of Node, so I picked the latest stable. I guess it's probably a good idea to specify that in the build environment document instead.

I'd very much like a second opinion on this from someone who's more up to speed with Node development.

## Context

This was discovered during releaste testing of v2021.2.
I looked for anything about raising the minimum supported version of Node in the files in the root directory of zonemaster-gui (including the change log), but I couldn't find anything.
The README just prescribes the installation of the latest version.

I did find a4708f3a64ad4bee4fcbaf9ae4d97fd53f6d2bed where Node.js 10 is declared as minimum version, I couldn't find any such declaration in our current documentation.

## Changes

The alias `stable` is used instead of specific versions in Ubuntu-Node.js-build-environment.

## How to test this PR

1. Start with a machine without Node.js installed,
2. Follow the Ubuntu-Node.js-build-environment document.
3. Follow the ReleaseProcess-create-test-distribution document to "produce a distribution zip file".

### DIFF
--- a/docs/internal-documentation/distrib-testing/Ubuntu-Node.js-build-environment.md
+++ b/docs/internal-documentation/distrib-testing/Ubuntu-Node.js-build-environment.md
@@ -22,8 +22,8 @@ been tested on Ubuntu 18.04, which we use here.
 4. Install Node.js by using [NVM], a node version manager.
    ```sh
    curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh | bash
-   nvm install 14.17.0
-   nvm use 14.17.0
+   nvm install stable
+   nvm use stable
    ```
 
 5. After installation, log out and log in again to handle [known issue], or just:


matsduf (commented):
It sounds like a good idea.

blacksponge (approved):
I am fine with that

matsduf (approved):

blacksponge:
We may need to rethink that change, https://app.travis-ci.com/github/zonemaster/zonemaster-gui/builds/242413674

blacksponge:
The issue seems to be related to some unmaintained modules. I suggest we don't change the supported node version for this release (we can still bump it to the latest lts, version 16, I have been tested this one a lot as this is the one I have in my dev environment) and replace the problematic dependencies for the next one (the issue is mainly related to protractor, used in e2e tests, that is no longer actively maintained, and by extension other additional testing modules like protractor-image-comparison).

matsduf:
Should we revert? We could look at this after the release.

matsduf:
> The issue seems to be related to some unmaintained modules. I suggest we don't change the supported node version for this release (we can still bump it to the latest lts, version 16,

Please create a PR that updates the file with relevant instructions.

PR #1025, opened 2021-12-05
Updated PostgreSQL version for FreeBSD

matsduf:
## Purpose

Matches the update in Zonemaster-Backend. See zonemaster/zonemaster-backend#925 and zonemaster/zonemaster-backend#926.


mattias-p (approved):
LGTM.

But which PostgreSQL version did we test v2021.2 against?

matsduf:
> LGTM.
> 
> But which PostgreSQL version did we test v2021.2 against?

The one in the update.

Issue #1028, opened 2021-12-16
Proposal of new test case: verify SOA MNAME


matsduf:
The SOA MNAME is supposed to be the master server of the zone. Currently Zonemaster does not do much tests of that server. This issue proposes a new test case with the following verifications of SOA MNAME:

* That the server responds to DNS queries.
* That the server is authoritative to the zone.
* That the SOA serial is identical to the NS name servers.

From [RFC 1035]:
> MNAME           The <domain-name> of the name server that was the
>                original or primary source of data for this zone.

From [RFC 1996]:
>    Primary Master  master server at the root of the zone transfer
                   dependency graph.  The primary master is named in the
                   zone's SOA MNAME field and optionally by an NS RR.
                   There is by definition only one primary master server
                   per zone.

From [RFC 2136]:
> Primary Master  master server at the root of the AXFR/IXFR
                      dependency graph.  The primary master is named in
                      the zone's SOA MNAME field and optionally by an NS
                      RR.  There is by definition only one primary master
                      server per zone.

From [RFC 2181]:
>    It is quite clear in the specifications, yet seems to have been
   widely ignored, that the MNAME field of the SOA record should contain
   the name of the primary (master) server for the zone identified by
   the SOA.  It should not contain the name of the zone itself.  That
   information would be useless, as to discover it, one needs to start
   with the domain name of the SOA record - that is the name of the
   zone.

[RFC 1035]: https://datatracker.ietf.org/doc/html/rfc1035#section-3.3.13
[RFC 1996]: https://datatracker.ietf.org/doc/html/rfc1996#section-2
[RFC 2136]: https://datatracker.ietf.org/doc/html/rfc2136#section-1
[RFC 2181]: https://datatracker.ietf.org/doc/html/rfc2181#section-7.3

tgreenx:
Rather than a new test case, this issue now relates to updating the specification of Zone01 with further testing.

tgreenx:
Resolved by #1032

PR #1000, opened 2021-10-24
Adds specification of default query and response handling

matsduf:
## Purpose

This PR adds a specification for the default setting of the DNS queries that the test cases emit. Test cases are to refer to this specification and use the default setting or specify the deviation.

1. By having a shared specification the test cases will be more consistent.
2. The test case specifications do not have to specify the DNS query when default values are used.

The specification also has default handling of the response, and that will give consistent handling in the test cases.

PR #870, opened 2020-07-04
Updated test case Zone09

matsduf:
Re-written test case Zone09

* Added exception for root zone, TLDs and zones under .ARPA to be required (expected) to have MX.
* Updated the steps to be explicit.
* Updated the steps to cover all name servers for the zone.
* Added explicit messages.
* Ignored address records (A/AAAA) from being mail domain target.
* Considered "Null MX"
* Assumes that Basic04 is run and disregard basic name server problems.

This PR will resolve issue #841 and include some validation of "Null MX" (issue #754).

Further updates:

* Adapts to draft template (zonemaster#971)
* Messages are updated.
* Informational messages removed to keep down the amount of messages.

When this update has been merged, the implementation must also be updated.

-----
Updates 2022-07-15

The PR was rebased on latest develop branch.  All commits before 12085c1aff68926c9b2eda7b835de369501acb3b are the same changes that were done 2022-01-17 and earlier.

Added updates:

* Adds reference to "Query and responds defaults".
* Updates the message tags to make them shorter.
* Updates the message tag table to make it follow proposed
  updates to the specification template.
  * Adds message ID
  * Adds arguments to message ID
* Editorial updates.

PR #1032, opened 2022-01-04
Update specification of Zone01 - #1028

tgreenx:
## Purpose

This PR proposes an update to the specification of Zone01 test as described in #1028. The corresponding implementation is provided in another PR ([here](https://github.com/zonemaster/zonemaster-engine/pull/1035)).

## Context

Addresses #1028.


matsduf:
We should try to have consistent format of the test case specifications. For that purpose there is a template for the specification, [Template01.md]. Also look at recently created or updated test case specifications such as [DNSSEC18] and [draft Zone09].

There is also a specification of DNS query (#1000) that has been reviewed, but it requires a final approval. With that the specification can be further standardised.

[Template01.md]: https://github.com/zonemaster/zonemaster/blob/develop/docs/internal-documentation/templates/specifications/tests/Template01.md
[DNSSEC18]: https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/DNSSEC-TP/dnssec18.md
[draft Zone09]: https://github.com/zonemaster/zonemaster/pull/870

tgreenx:
I have updated this (new) specification to meet the template's requirements. I have done what I can to detail the Test Procedure but that will surely require more work, help appreciated @matsduf .

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,166 @@
...
+      4. There is no SOA record with owner name matching the query.
+   3. From the DNS response, store the MNAME and its associated SERIAL.
+      1. Go to next name server IP if at least one of the following criteria is met:
+         1. If there is no MNAME record, then add name server IP to the *No MNAME Record* set. 
+         2. If there is no SERIAL record, then add name server IP to the *No SERIAL Record* set. 
+         3. If the name in the MNAME record is the same as *Child Zone*, then add name server IP to the *MNAME Is Zone Name* set.
+   4. Send *SOA Query* to the name server in the MNAME ("MNAME Nameserver").
+      1. If the SOA response has [RCODE Name] "NoError" then:
+         1. If the AA flag is not set, then output *Z11_MNAME_NOT_AUTHORITATIVE* with name server name and IP.
+      2. Else, output *Z11_MNAME_NO_RESPONSE* with name server name and IP and go to next name server IP.
```
matsduf, line 91: Wouldn't it be more efficient to collect all MNAME found in the loop in 4 starting on line 76, and then send a query to each of them? In most cases there will be only one. (Consistency is covered by Consistency06).

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,166 @@
...
+   5. Obtain the set of name server IP addresses using [Method5] ("NS IP").
```
matsduf, line 92: This set, *NS IP* will be identical to *Name Server IP* from 2 on lines 66-67. Why again?

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,166 @@
...
+   6. For each name server in *NS IP*, send *SOA Query* to the name server and collect the response.
+      1. If there is a response, compare the SERIAL of that name server to the one from *MNAME Nameserver*.
+         1. If they match, then add name server IP to the *MNAME Is Master* set.
+         2. Else, then add name server IP to the *MNAME Not Master* set.
```
matsduf, line 96: I suggest that the information is collected in a set and compared at the end.

It is not possible to state if MNAME is master or not by just comparing the SOA serial. The master server is expected to have the highest serial number, if they are different. And highest must be calculated with respect to the special serial number arithmetic  [RFC1982].

[RFC1982]: https://datatracker.ietf.org/doc/html/rfc1982

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,166 @@
...
+   7. Obtain the set of name server IP names using [Method2] ("Glue Names").
+      1. If *MNAME Nameserver* is not part of the *Glue Names* set for the zone, then output *Z11_MNAME_NOT_IN_GLUE*.
```
matsduf, line 98: Why not checking against the NS RRset in the zone (method 3) or the combined NS RRset?

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,166 @@
...
+   with the name server IP addresses from the set.
+
+7. If the set *MNAME Is Zone Name* is non-empty, then output *Z11_MNAME_IS_ZONE_NAME*
+   with the name server IP addresses from the set.
+
+8. If the set *MNAME Not Master* is non-empty, then output *Z11_MNAME_NOT_MASTER*
+   with the name server IP addresses from the set.
+
+9. If the set *MNAME Is Master* is non-empty, then output *Z11_MNAME_IS_MASTER*
+   with the name server IP addresses from the set.
```
matsduf, line 113: Make them as links, `*[Z11_NO_MNAME_RECORD]*` etc.

matsduf (changes requested):

docs/specifications/tests/Zone-TP/zone11.md
tgreenx: Fixed

docs/specifications/tests/Zone-TP/zone11.md
tgreenx: Fixed

docs/specifications/tests/Zone-TP/zone11.md
tgreenx: Fixed

docs/specifications/tests/Zone-TP/zone11.md
tgreenx: Removed

docs/specifications/tests/Zone-TP/zone11.md
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,166 @@
...
+   1. Send *SOA Query* to the name server and collect the response.
+   2. Go to next name server IP if at least one of the following criteria is met:
+      1. There is no DNS response.
+      2. [RCODE Name] of the response is not "NoError".
+      3. The AA flag is not set in the response.
+      4. There is no SOA record with owner name matching the query.
+   3. From the DNS response, store the MNAME and its associated SERIAL.
+      1. Go to next name server IP if at least one of the following criteria is met:
+         1. If there is no MNAME record, then add name server IP to the *No MNAME Record* set. 
+         2. If there is no SERIAL record, then add name server IP to the *No SERIAL Record* set.
```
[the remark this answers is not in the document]
tgreenx: Removed

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,166 @@
...
+what is specified for [DNS Response] in the same specification.
+
+1. Create a [DNS Query] with query type SOA and query name *Child Zone* ("SOA Query")
+
+2. Obtain the set of name server IP addresses using [Method5]
+   ("Name Server IP").
+
+3. Create the following empty sets:
+   1. Name server IP address ("No MNAME Record")
+   2. Name server IP address ("No Serial Record")
```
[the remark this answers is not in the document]
tgreenx: Fixed

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,166 @@
...
+
+In this section and unless otherwise specified below, the terms "[DNS Query]"
+follow the specification for DNS queries as specified in [DNS Query and Response Defaults].
+The handling of the DNS responses on the DNS queries follow, unless otherwise specified below,
+what is specified for [DNS Response] in the same specification.
+
+1. Create a [DNS Query] with query type SOA and query name *Child Zone* ("SOA Query")
+
+2. Obtain the set of name server IP addresses using [Method5]
+   ("Name Server IP").
```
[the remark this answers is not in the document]
tgreenx: Fixed

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,166 @@
...
+Z11_NO_MNAME_RECORD           | WARNING | ns_ip_list           | No MNAME records found from name servers "{ns_ip_list}".
+Z11_NO_SERIAL_RECORD          | WARNING | ns_ip_list           | No SERIAL records found from name servers "{ns_ip_list}".
+Z11_MNAME_IS_ZONE_NAME        | WARNING | ns_ip_list           | MNAME name server points to the zone tested. Returned from name servers "{ns_ip_list}".
+Z11_MNAME_NOT_AUTHORITATIVE   | WARNING | ns                   | MNAME name server points to a non authoritative name server ("{ns}").
+Z11_MNAME_NO_RESPONSE         | NOTICE  | ns                   | MNAME name server points to a non responsive name server ("{ns}").
+Z11_MNAME_NOT_IN_GLUE         | NOTICE  |                      | MNAME name server could not be found in the parent zone name servers.
+Z11_MNAME_NOT_MASTER          | WARNING | ns_ip_list           | MNAME name server has a mis-matching SERIAL. Returned from name servers "{ns_ip_list}".
+Z11_MNAME_IS_MASTER           | INFO    | ns_ip_list           | MNAME name server is master. Returned from name servers "{ns_ip_list}".
```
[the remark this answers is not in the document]
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,166 @@
...
+## Summary
+
+Message Tag                   | Level   | Arguments            | Message ID for message tag
+:---------------------------- |:--------|:---------------------|:---------------------------------------------------------------------------------------
+Z11_NO_MNAME_RECORD           | WARNING | ns_ip_list           | No MNAME records found from name servers "{ns_ip_list}".
+Z11_NO_SERIAL_RECORD          | WARNING | ns_ip_list           | No SERIAL records found from name servers "{ns_ip_list}".
+Z11_MNAME_IS_ZONE_NAME        | WARNING | ns_ip_list           | MNAME name server points to the zone tested. Returned from name servers "{ns_ip_list}".
+Z11_MNAME_NOT_AUTHORITATIVE   | WARNING | ns                   | MNAME name server points to a non authoritative name server ("{ns}").
+Z11_MNAME_NO_RESPONSE         | NOTICE  | ns                   | MNAME name server points to a non responsive name server ("{ns}").
+Z11_MNAME_NOT_IN_GLUE         | NOTICE  |                      | MNAME name server could not be found in the parent zone name servers.
```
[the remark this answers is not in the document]
tgreenx: Fixed

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,166 @@
...
+* [Terminology](#terminology)
+
+## Objective
+
+The SOA MNAME record must be a fully qualified master nameserver.
+ 
+The purpose of this test is to test:
+    - that the server is authoritative to the zone.
+    - that the MNAME is not the zone name.
+    - that the SOA serial is identical to the zone name servers.
```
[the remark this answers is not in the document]
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,166 @@
...
+* [Intercase dependencies](#Intercase-dependencies)
+* [Terminology](#terminology)
+
+## Objective
+
+The SOA MNAME record must be a fully qualified master nameserver.
+ 
+The purpose of this test is to test:
+    - that the server is authoritative to the zone.
+    - that the MNAME is not the zone name.
```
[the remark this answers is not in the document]
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,166 @@
...
+* [Inputs](#Inputs)
+* [Summary](#Summary)
+* [Test procedure](#Test-procedure)
+* [Outcome(s)](#Outcomes)
+* [Special procedural requirements](#Special-procedural-requirements)
+* [Intercase dependencies](#Intercase-dependencies)
+* [Terminology](#terminology)
+
+## Objective
+
```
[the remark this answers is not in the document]
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,169 @@
...
+## Inputs
+
+* "Child Zone" - The domain name to be tested.
+
+## Summary
+
+Message Tag                   | Level   | Arguments            | Message ID for message tag
+:---------------------------- |:--------|:---------------------|:---------------------------------------------------------------------------------------
+Z11_MULTIPLE_MNAME            | WARNING | ns_ip_list           | Distinct MNAMEs found, but only one is expected. Returned from name servers "{ns_ip_list}".
```
matsduf, line 50: This will overlap message from Consistency06. I think we should avoid having the same message more than once.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,169 @@
...
+Z11_MNAME_NOT_AUTHORITATIVE   | WARNING | ns                   | MNAME name server points to a non authoritative name server ("{ns}").
+Z11_MNAME_NO_RESPONSE         | NOTICE  | ns                   | MNAME name server points to a non responsive name server ("{ns}").
```
matsduf, line 52: "non authoritative" --> "non-authoritative"
"non responsive" --> "non-responsive"

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,169 @@
...
+Z11_MNAME_NOT_IN_DELEGATION   | INFO    | nsname               | MNAME name server {nsname} is not listed as NS record for the zone.
```
matsduf, line 53: It is permitted to have more NS in the zone than in the delegation as long as the latter is a subset. If MNAME is in one of the NS in the zone I see no reason to output a message.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,169 @@
...
+Z11_MNAME_NOT_MASTER          | WARNING | ns_ip_list           | MNAME name server doesn't have the highest SERIAL. Returned from name servers "{ns_ip_list}".
```
matsduf, line 54: I suggest "does not" to be more formal.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,169 @@
...
+follow the specification for DNS queries as specified in [DNS Query and Response Defaults].
+The handling of the DNS responses on the DNS queries follow, unless otherwise specified below,
+what is specified for [DNS Response] in the same specification.
+
+1. Create a [DNS Query] with query type SOA and query name *Child Zone* ("SOA Query")
+
+2. Obtain the set of name server IP addresses using [Method4] and [Method5] ("Name Server IP").
+
+3. Create the following empty sets:
+   1. Name server IP address, MNAME, SERIAL ("MNAME Nameserver")
```
matsduf, line 76: This should be `1. Name server name ("MNAME Nameserver")` to match the population in 4.3 below.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,169 @@
...
+      1. There is no DNS response.
+      2. [RCODE Name] of the response is not "NoError".
+      3. The AA flag is not set in the response.
+      4. There is no SOA record with owner name matching the query.
+   3. From the DNS response, add the MNAME field to the *MNAME Nameserver* set.
+   4. From the DNS response, add the SERIAL field and name server IP address to the *SERIAL Nameservers* set.
+   5. Go to next name server.
+
+5. If the set *MNAME Nameserver* contains more than one element, then output *[Z11_MULTIPLE_MNAME]*
+   with the name server IP addresses from the set, and terminate these procedures.
```
matsduf, line 93: Besides that it overlaps with Consistency06 I wonder why it stops here? I should be possible to check more than one MNAME.

tgreenx:
Now that I look more at [Zone01](https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/Zone-TP/zone01.md) in details (specification and implementation), it appears to be greatly overlapping with this PR, especially when you consider Zone01 title name.  And so I can't seem to justify to add this new test case rather than updating Zone01 instead. Wasn't the goal of Zone01 to do just what we are trying to do now? What do you think?

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,169 @@
...
+
+6. Send *SOA Query* to the name server in *MNAME Nameserver*.
```
matsduf, line 95: Here you have a hidden DNS lookup that I think should be explicit.
1. Then name might be in-bailiwick, and then we must query at least one of the name servers for the zone.
2. The name might not exist, NXDOMAIN.
3. There might not be any address record, NODATA.
4. There might be both IPv4 and IPv6 address for the name.
5. There might be more than one IPv4 or IPv6 address for the name.
6. The address or addresses might be not routable (e.g. private).
7. The address or addresses might not be reachable.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,169 @@
...
+   1. If there is a DNS response, with [RCODE Name] "NoError", then:
+      1. If the AA flag is not set, then output *[Z11_MNAME_NOT_AUTHORITATIVE]* with name server name and IP address.
+      2. Else, add the SERIAL field to the *MNAME Nameserver* set.
```
matsduf, line 98: Since each MNAME can have more than one IP address it could end up with more than one Serial too.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,169 @@
...
+   2. Else, output *[Z11_MNAME_NO_RESPONSE]* with name server name and IP.
+
+7. For each SERIAL value in the *SERIAL Nameservers* do:
+   1. Compare the value with the one in the *MNAME Nameserver* set.
```
matsduf, line 102: "using the arithmetic in RFC 1982"

matsduf (changes requested):

matsduf:
> Now that I look more at [Zone01](https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/Zone-TP/zone01.md) in details (specification and implementation), it appears to be greatly overlapping with this PR, especially when you consider Zone01 title name. And so I can't seem to justify to add this new test case rather than updating Zone01 instead. Wasn't the goal of Zone01 to do just what we are trying to do now? What do you think?

I think you are right. I missed that when writing the issue. I agree it is better to update Zone01.

matsduf:
> I have renamed this PR, its content, and the associated issue. Thanks for the comments, I will address them when I can.

Also look at the Objective of the current Zone01. It might have references that make sense. The more references the better.

docs/specifications/tests/Zone-TP/zone11.md
tgreenx: Fixed

docs/specifications/tests/Zone-TP/zone11.md
tgreenx: Fixed

docs/specifications/tests/Zone-TP/zone11.md
tgreenx: This set is also used in step 6.1.2 to store the SERIAL field of the MNAME name server. And the IP address is an implicit requirement, from a coding perspective, for sending the query (step 6) to the MNAME name server.

docs/specifications/tests/Zone-TP/zone11.md
tgreenx: Ok, sure, we can test all of them and see what emerges.

docs/specifications/tests/Zone-TP/zone11.md
tgreenx: Fixed

docs/specifications/tests/Zone-TP/zone11.md
tgreenx: On second thoughts I will indeed remove the IP address from this set since it is covered by a DNS lookup from step 6.

docs/specifications/tests/Zone-TP/zone11.md
tgreenx: Reworded

docs/specifications/tests/Zone-TP/zone11.md
tgreenx: Fixed

docs/specifications/tests/Zone-TP/zone11.md
tgreenx: Fixed

docs/specifications/tests/Zone-TP/zone11.md
matsduf: The user will get the message from Consistency06.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,181 @@
...
+   4. From the DNS response, add the SERIAL field and name server IP address to the *SERIAL Nameservers* set.
+   5. Go to next name server.
+
+5. If the set *MNAME Nameservers* is empty, then terminate these procedures.
+
+6. Else, for each name server name in *MNAME Nameservers* do:
+   1. Make a recursive DNS lookup by sending *A Query*, starting with one of the root servers for the name server name.
+      1. If there is a DNS response, with [RCODE Name] "NoError", then add the name server IP address(es) from the A record to the *MNAME Nameservers* set.
+      2. Else Make a recursive DNS lookup by sending *AAAA Query*, starting with one of the root servers for the name server name.
+         1. If there is a DNS response, with [RCODE Name] "NoError", then add the name server IP address(es) from the AAAA record to the *MNAME Nameservers* set.
```
matsduf, line 100: Here you are resolving the IP addresses of MNAME. What should be looked for are A and AAAA records, not the RCODE. The name could be anything. I suggest a "DNS Lookup" as in the proposed update of Basic01 (#1082).

```
* "DNS Lookup" - The term is used when a recursive lookup is used, though
any changes to the DNS tree introduced by an [undelegated test] must be
respected.
```

And if the name is "localhost" then we should not do any lookup, rather output Z01_MNAME_LOCALHOST (see my suggestion above).

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,181 @@
...
+   2. If at least one IP address from the previous steps (6.1.1 or 6.1.2.1) was found, then:
```
matsduf, line 101: If there is no such IP address, then Z01_MNAME_NOT_RESOLVE would fit (see suggestion above).

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,181 @@
...
+
+## Special procedural requirements
+
+If either IPv4 or IPv6 transport is disabled, skip sending queries over that
+transport protocol. A message will be outputted reporting that the transport
+protocol has been skipped.
+
+## Intercase dependencies
+
+The syntax of the SOA MNAME field is tested in [SYNTAX07].
```
matsduf, line 148: I suggest that to be under "Scope".

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,181 @@
...
-this test case fails.
+Note that the SOA SERIAL comparison must follow [RFC1982].
 
-### Special procedural requirements
+## Scope
 
-None.
+It is assumed that *Child Zone* has been tested and reported by [Basic04]. This
+test case will just ignore non-responsive name servers or name servers not
+giving a correct DNS response for an authoritative name server.
```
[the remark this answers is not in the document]
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone01.md
tgreenx: Fixed

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,181 @@
...
+   2. Go to next name server IP if at least one of the following criteria is met:
+      1. There is no DNS response.
+      2. [RCODE Name] of the response is not "NoError".
+      3. The AA flag is not set in the response.
+      4. There is no SOA record with owner name matching the query.
+   3. From the DNS response, add the name server name from the MNAME field to the *MNAME Nameservers* set.
+   4. From the DNS response, add the SERIAL field and name server IP address to the *SERIAL Nameservers* set.
+   5. Go to next name server.
+
+5. If the set *MNAME Nameservers* is empty, then terminate these procedures.
```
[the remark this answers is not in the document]
tgreenx: Either because, for some reason, (i) the set *Name Server IP* is empty (no name servers found), or (ii) no name server responded with a valid response (any conditions from step 4.1.2)

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,181 @@
...
+
+## Summary
+
+Message Tag                   | Level   | Arguments            | Message ID for message tag
+:---------------------------- |:--------|:---------------------|:---------------------------------------------------------------------------------------
+Z01_MNAME_NOT_AUTHORITATIVE   | WARNING | ns                   | MNAME name server points to a non-authoritative name server ("{ns}").
+Z01_MNAME_NO_RESPONSE         | NOTICE  | ns                   | MNAME name server points to a non-responsive name server ("{ns}").
+Z01_MNAME_NOT_IN_ZONE         | INFO    | nsname               | MNAME name server {nsname} is not listed as NS record for the zone.
+Z01_MNAME_NOT_MASTER          | WARNING | ns_ip_list           | MNAME name server does not have the highest SERIAL. Returned from name servers "{ns_ip_list}".
+Z01_MNAME_IS_MASTER           | DEBUG   | ns_ip_list           | MNAME name server does appear to be master. Returned from name servers "{ns_ip_list}".
```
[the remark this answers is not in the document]
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone01.md
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone01.md
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,209 @@
...
+   5. Go to next name server.
+
+5. If the set *MNAME Is Localhost* is non-empty, then output *[Z01_MNAME_IS_LOCALHOST]* with name servers IP address(es).
+
+6. If the set *MNAME Is Dot* is non-empty, then output *[Z01_MNAME_IS_DOT]* with name servers IP address(es).
+
+7. If the set *MNAME Nameservers* is empty, then terminate these procedures.
+
+8. Else, for each name server name in *MNAME Nameservers* do:
+   1. Do a [DNS Lookup] of name server name (A and AAAA) and add the name server IP address(es) found to the *MNAME Nameservers* set.
```
matsduf, line 116: Maybe "...to the same entry as the name server name".

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,209 @@
...
+   2. If at least one IP address from the previous step was found, then:
+      1. For each name server IP for the name server name do:
```
matsduf, line 118: Should there be exceptions for IP addresses 127.0.0.1 and ::1? I.e. another message and no query?

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,209 @@
...
+            1. If there is a DNS response with [RCODE Name] "NoError", then:
+               1. If the AA flag is not set, then output *[Z01_MNAME_NOT_AUTHORITATIVE]* with name server name and IP address.
+               2. Else, add the SERIAL field to the *MNAME Nameservers* set.
+            2. Else if there is a DNS response with [RCODE Name] other than "NoError", then output *[Z01_MNAME_UNEXPECTED_RCODE]* with [RCODE Name], name server name and IP address.
+            3. Else, output *[Z01_MNAME_NO_RESPONSE]* with name server name and IP address.
+         2. Go to next name server IP.
+   3. Else, output *[Z01_MNAME_NOT_RESOLVE]* with name server name.
+   4. Go to next name server name.
+
+9. If there is no SERIAL in the set *MNAME Nameservers*, then terminate these procedures.
```
matsduf, line 129: I cannot see that it is possible. An SOA record always contains both MNAME and SOA SERIAL.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,209 @@
...
+
+10. For each SERIAL in *SERIAL Nameservers* do:
+   1. For each SERIAL in *MNAME Nameservers* do:
+      1. Compare both SERIAL values (using the arithmetic in [RFC1982]).
+         1. If the one from *SERIAL Nameservers* is not higher, then add name server IP address to the *MNAME Is Master* set.
+         2. Else, add name server IP address to the *MNAME Not Master* set.
+      2. Go to next SERIAL.
+
+11. If the set *MNAME Not Master* is non-empty, then output *[Z01_MNAME_NOT_MASTER]*
+   with the name server IP addresses from the set.
```
matsduf, line 139: I think it requires additional MNAME name and IP address.

matsduf (changes requested):
It is easier to handle the text in general if lines are kept to 80 characters. From the template:

> Limit all lines to 80 characters with the possible exception of tables such as the one in the summary section.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,209 @@
...
+[RFC2181], section 7.2, clarifies that "it is quite clear in the specifications, yet seems to have been widely ignored, that the MNAME field of the SOA record should contain the name of the primary (master) server for the zone identified by the SOA. It should not contain the name of the zone itself. That information would be useless, as to discover it, one needs to start with the domain name of the SOA record - that is the name of the zone".
 
-If the SOA MNAME field is not an authoritative DNS server for the domain name,
-this test case fails.
+There exists an unstandardized practice to set the SOA MNAME to "." to mean no server at all, indicating that there is no default server for dynamic updates. There is at least one old and expired Internet-Draft that attempted to standardize that behavior, i.e. [draft-jabley-dnsop-missing-mname]. If the SOA MNAME is an empty name (".") this Test Case will not try to connect to a server
+behind it since there will never be a server behind that name, as the purpose is most definitely to follow that practice. Instead, a special message will be outputted.
+ 
+This Test Case will check that:
+ - the SOA MNAME contains the master name server.
```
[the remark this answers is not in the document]
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,209 @@
...
+ - this master name server is authoritative to the zone.
```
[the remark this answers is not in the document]
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,209 @@
...
+ - the SOA SERIAL of the MNAME is identical or higher than the ones found from the name servers in the delegation.
```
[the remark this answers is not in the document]
tgreenx: Updated, with the most relevant name server list.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,209 @@
...
+ - the master name server is listed as part of the delegation of the zone.
```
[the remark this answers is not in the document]
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,209 @@
...
 
-### Special procedural requirements
+Note that the SOA SERIAL comparison must follow [RFC1982].
```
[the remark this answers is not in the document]
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,209 @@
...
+      1. There is no DNS response.
+      2. [RCODE Name] of the response is not "NoError".
+      3. The AA flag is not set in the response.
+      4. There is no SOA record with owner name matching the query.
+   3. From the DNS response, extract the name server name from the MNAME field:
+      1. If the name is "localhost", then add name server IP address to the *MNAME Is Localhost* set and go to next name server.
+      2. Else if the name is ".", then add name server IP address to the *MNAME Is Dot* set and go to next name server.
+      3. Else, add name server name to the *MNAME Nameservers* set.
+   4. From the DNS response, add the SERIAL field and name server IP address to the *SERIAL Nameservers* set.
+   5. Go to next name server.
```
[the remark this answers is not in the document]
tgreenx: Yes. I thought it would not be useful to do the SERIAL check when the MNAME refers to an unexpected name server name. Plus step 8 cannot be applied to such name server names. If you think otherwise I would be interested to know why.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,209 @@
...
+
+In this section and unless otherwise specified below, the terms "[DNS Query]"
+follow the specification for DNS queries as specified in [DNS Query and Response Defaults].
+The handling of the DNS responses on the DNS queries follow, unless otherwise specified below,
+what is specified for [DNS Response] in the same specification.
+
+1. Create a [DNS Query] with query type SOA and query name *Child Zone* ("SOA Query").
+
+2. Create the following empty sets:
+   1. Name server IP address, name, SERIAL ("MNAME Nameservers")
```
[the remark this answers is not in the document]
tgreenx: Updated for both SOA SERIAL and SOA MNAME.

docs/specifications/tests/Zone-TP/zone01.md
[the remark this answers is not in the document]
matsduf: The SOA SERIAL that you collect in 4.4 is not the SOA SERIAL that is found in the SOA MNAME server, but the SOA SERIAL found in the servers listed in **Name Server IP**. I do not see why that should be affected by the SOA MNAME value.

docs/specifications/tests/Zone-TP/zone01.md
tgreenx: This is here just in case the DNS lookup from step 8 failed, for any reason.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,209 @@
...
+            2. Else if there is a DNS response with [RCODE Name] other than "NoError", then output *[Z01_MNAME_UNEXPECTED_RCODE]* with [RCODE Name], name server name and IP address.
+            3. Else, output *[Z01_MNAME_NO_RESPONSE]* with name server name and IP address.
+         2. Go to next name server IP.
+   3. Else, output *[Z01_MNAME_NOT_RESOLVE]* with name server name.
+   4. Go to next name server name.
+
+9. If there is no SERIAL in the set *MNAME Nameservers*, then terminate these procedures.
+
+10. For each SERIAL in *SERIAL Nameservers* do:
+   1. For each SERIAL in *MNAME Nameservers* do:
```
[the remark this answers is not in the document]
tgreenx: Fixed, and for step 12 as well

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,209 @@
...
+      1. Compare both SERIAL values (using the arithmetic in [RFC1982]).
+         1. If the one from *SERIAL Nameservers* is not higher, then add name server IP address to the *MNAME Is Master* set.
+         2. Else, add name server IP address to the *MNAME Not Master* set.
```
matsduf, line 135: For programming this is fine, but it is quite complex for manual execution. And what IP address is "name server IP address"?

docs/specifications/tests/Zone-TP/zone01.md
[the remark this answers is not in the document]
tgreenx: True, thanks, I got mixed up after the last update. I will make changes accordingly.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,209 @@
...
+
+## Special procedural requirements
+
+If either IPv4 or IPv6 transport is disabled, skip sending queries over that
+transport protocol. A message will be outputted reporting that the transport
+protocol has been skipped.
+
+## Intercase dependencies
+
+There are no intercase dependencies for this Test Case.
```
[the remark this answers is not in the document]
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone01.md
matsduf: Or some other error. That is fine, because your master test depends on SOA SERIAL.

matsduf, line 129: Even if the set has no SOA SERIAL, the comparison of SOA MNAME against the NS list should be done.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,209 @@
...
+7. If the set *MNAME Nameservers* is empty, then terminate these procedures.
+
+8. Else, for each name server name in *MNAME Nameservers* do:
+   1. Do a [DNS Lookup] of name server name (A and AAAA) and add the name server IP address(es) found to the *MNAME Nameservers* set.
+   2. If at least one IP address from the previous step was found, then:
+      1. For each name server IP for the name server name do: 
+         1. Send *SOA Query* to the name server and collect the response.
+            1. If there is a DNS response with [RCODE Name] "NoError", then:
+               1. If the AA flag is not set, then output *[Z01_MNAME_NOT_AUTHORITATIVE]* with name server name and IP address.
+               2. Else, add the SERIAL field to the *MNAME Nameservers* set.
```
matsduf, line 122: In principle the server could have the AA flag set without any SOA record in the response.

docs/specifications/tests/Zone-TP/zone01.md
tgreenx: Updated.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,209 @@
...
+      1. Compare both SERIAL values (using the arithmetic in [RFC1982]).
+         1. If the one from *SERIAL Nameservers* is not higher, then add name server IP address to the *MNAME Is Master* set.
+         2. Else, add name server IP address to the *MNAME Not Master* set.
+      2. Go to next SERIAL.
+
+11. If the set *MNAME Not Master* is non-empty, then output *[Z01_MNAME_NOT_MASTER]*
+   with the name server IP addresses from the set.
+
+12. If the set *MNAME Is Master* is non-empty, then:
+   1. Output *[Z01_MNAME_IS_MASTER]* with the name server IP addresses from the set.
```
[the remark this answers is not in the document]
tgreenx: Fixed

docs/specifications/tests/Zone-TP/zone01.md
tgreenx: Fixed

docs/specifications/tests/Zone-TP/zone01.md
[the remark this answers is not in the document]
tgreenx: Fixed

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,209 @@
...
+The handling of the DNS responses on the DNS queries follow, unless otherwise specified below,
+what is specified for [DNS Response] in the same specification.
+
+1. Create a [DNS Query] with query type SOA and query name *Child Zone* ("SOA Query").
+
+2. Create the following empty sets:
+   1. Name server IP address, name, SERIAL ("MNAME Nameservers")
+   2. Name server IP address, SERIAL ("SERIAL Nameservers")
+   3. Name server IP address ("MNAME Not Master")
+   4. Name server IP address ("MNAME Is Master")
```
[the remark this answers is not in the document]
tgreenx: Fixed

docs/specifications/tests/Zone-TP/zone01.md
tgreenx: I thought this comparison was only intended for the *identified* master name server, so after the SOA SERIAL comparison. From the **Objective** section: `"- the master name server is listed as part of the NS record set of *Child Zone*."`
Here, at this point, those are only master name server *candidates* (i.e. before SOA SERIAL comparison). Should the NS list comparison really be done for all master name server *candidates*?

docs/specifications/tests/Zone-TP/zone01.md
tgreenx: I have updated and reworded this step to make it clearer. Although it may still be a struggle to read, let me know.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,252 @@
...
+There is at least one old and expired Internet-Draft that attempted to standardize
+that behavior, i.e. [draft-jabley-dnsop-missing-mname]. If the SOA MNAME
+is an empty name (".") this Test Case will not try to connect to a server behind
+it since there will never be a server behind that name, as the purpose is most
+definitely to follow that practice. Instead, a special message will be outputted.
+ 
+This Test Case will check that:
+ - the SOA MNAME contains the master name server of *Child Zone*,
+   as best as it can be determined.
+ - the master name server is authoritative of *Child Zone*.
```
matsduf, line 50: It should rather be
```
- the SOA MNAME name server is authoritative of *Child Zone*.
```

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,252 @@
...
+ - the SOA SERIAL of the SOA MNAME is at least equal to the ones found
+   from the name servers in the NS record set of *Child Zone*.
+   This comparison must be done following [RFC1982].
+ - the master name server is listed as part of the NS record set of *Child Zone*.
```
matsduf, line 54: It should rather be
```
- the SOA MNAME name server is listed as part of the NS record set of *Child Zone*.
```

docs/specifications/tests/Zone-TP/zone01.md
matsduf: We could never know what is the master server. If all name servers found, including SOA MNAME, report the same SOA SERIAL, any of the servers could be master server, or some other hidden server. But we can verify that SOA MNAME is in the NS list.

docs/specifications/tests/Zone-TP/zone01.md
matsduf: The localhost IP addresses are special. The only "foreign" servers that Zonemaster connects to are name servers. It does no check on parent or grand-parent name servers when it comes to IP address. The name servers from delegation or the zone of *Child Zone* there is one test case that verifies the IP address and reacts on localhost address (among other non-public or non-global addresses). I do not think that this test case should repeat the entire IP address check, but it could make sense to capture the localhost addresses by having a special message and special code:

Message Tag                   | Level   | Arguments             | Message ID for message tag
:---------------------------- |:--------|:----------------------|:---------------------------------------------------------------------------------------
Z01_MNAME_HAS_LOCALHOST_ADDR      | WARNING   | nsname, ns_ip  | SOA MNAME name server "{nsname}" resolves to a localhost IP address ({ns_ip}).

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,252 @@
...
+
+10. For each SOA SERIAL in *MNAME Nameservers* do:
+    1. For each SOA SERIAL in *SERIAL Nameservers* do:
+       1. Compare both SOA SERIAL values (using the arithmetic in [RFC1982]).
+          1. If the one from *SERIAL Nameservers* is higher, then add SOA MNAME name
+             server name and IP address to the *MNAME Not Master* set, and go to next
+             SOA SERIAL (from the *MNAME Nameservers* set).
+          2. Go to next SOA SERIAL.
+    2. If the set *MNAME Not Master* is empty, then add SOA MNAME name server name and
+       IP address to the *MNAME Is Master* set and exit this loop.
```
matsduf, line 174: If there are two different SOA MNAME, then one could meet your criteria, but the other not. I do not think your steps handle that.

docs/specifications/tests/Zone-TP/zone01.md
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone01.md
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,252 @@
...
+      3. The AA flag is not set in the response.
+      4. There is no SOA record with owner name matching the query.
+   4. From the DNS response, extract the name server name from the SOA MNAME field:
+      1. If the name is "localhost", then add name server IP address to
+         the *MNAME Is Localhost* set.
+      2. Else if the name is ".", then add name server IP address to
+         the *MNAME Is Dot* set.
+      3. Else, add SOA MNAME name server name to the *MNAME Nameservers* set.
+   5. From the DNS response, add the SOA SERIAL field and name server IP address to
+      the *SERIAL Nameservers* set.
```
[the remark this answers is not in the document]
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,252 @@
...
+
+4. For each name server IP address in *Name Server IP* do:
+   1. Send *SOA Query* to the name server and collect the response.
+   2. Go to next name server IP address if at least one of
+      the following criteria is met:
+      1. There is no DNS response.
+      2. [RCODE Name] of the response is not "NoError".
+      3. The AA flag is not set in the response.
+      4. There is no SOA record with owner name matching the query.
+   4. From the DNS response, extract the name server name from the SOA MNAME field:
```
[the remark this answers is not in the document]
tgreenx: Fixed

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,252 @@
...
+9. If there is no SOA SERIAL in the *MNAME Nameservers* set, then
+   terminate these procedures.
+
+10. For each SOA SERIAL in *MNAME Nameservers* do:
+    1. For each SOA SERIAL in *SERIAL Nameservers* do:
+       1. Compare both SOA SERIAL values (using the arithmetic in [RFC1982]).
+          1. If the one from *SERIAL Nameservers* is higher, then add SOA MNAME name
+             server name and IP address to the *MNAME Not Master* set, and go to next
+             SOA SERIAL (from the *MNAME Nameservers* set).
+          2. Go to next SOA SERIAL.
```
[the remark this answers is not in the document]
tgreenx: Yes. I was only being explicit when breaking out from the inner loop to the outer loop (since they both have the same name values ["SOA SERIAL"]).

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,252 @@
...
+    2. If the set *MNAME Not Master* is empty, then add SOA MNAME name server name and
+       IP address to the *MNAME Is Master* set and exit this loop.
+    3. Else, go to next SOA SERIAL.
```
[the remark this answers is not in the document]
tgreenx: Yes, same reason as stated just above.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,252 @@
...
+      for the same entry as the SOA MNAME name server name.
+   2. If at least one IP address from the previous step was found, then:
+      1. For each SOA MNAME name server IP for the SOA MNAME name server name do: 
+         1. Send *SOA Query* to the name server and collect the response.
+            1. If there is a DNS response with [RCODE Name] "NoError", then:
+               1. If the AA flag is not set, then output *[Z01_MNAME_NOT_AUTHORITATIVE]* 
+                  with SOA MNAME name server name and IP address, and name server(s) IP
+                  address(es) from the *SERIAL Nameservers* set.
+               2. Else, add the SOA SERIAL field to the *MNAME Nameservers* set
+                  for the same entry as the SOA MNAME name server name.
```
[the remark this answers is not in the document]
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone01.md
tgreenx: Sure, added.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,252 @@
...
+
+6. If the set *MNAME Is Dot* is non-empty, then output
+   *[Z01_MNAME_IS_DOT]* with name server(s) IP address(es).
+
+7. If the set *MNAME Nameservers* is empty, then terminate these procedures.
+
+8. Else, for each SOA MNAME name server name in *MNAME Nameservers* do:
+   1. Do a [DNS Lookup] of SOA MNAME name server name (A and AAAA) and add the
+      SOA MNAME name server IP address(es) found to the *MNAME Nameservers* set
+      for the same entry as the SOA MNAME name server name.
```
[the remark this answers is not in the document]
tgreenx: Done

docs/specifications/tests/Zone-TP/zone01.md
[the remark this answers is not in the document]
tgreenx: > One SOA MNAME name server name could resolve into more than one IP address. The SOA SERIAL could be different on different name server IP addresses.

By the way, shouldn't this behavior require a check (and a message tag), outputted during the serial check procedure? So when the SOA serial has different values (depending on IPs found) that places this name server as both master and not master.
Something like `Z01_MNAME_INCONSISTENT_SERIAL`.

docs/specifications/tests/Zone-TP/zone01.md
matsduf: I suggested "SOA MNAME name server", not "master name server".

docs/specifications/tests/Zone-TP/zone01.md
matsduf: I suggested "SOA MNAME name server", not "master name server". Do you think that is a problem?

docs/specifications/tests/Zone-TP/zone01.md
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone01.md
tgreenx: Right, I used to think there could only be one. Updated.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,262 @@
...
+
+11. For each SOA SERIAL in *MNAME Nameservers* do:
+    1. For each SOA SERIAL in *SERIAL Nameservers* do:
+       1. Compare both SOA SERIAL values (using the arithmetic in [RFC1982]).
+          1. If the one from *SERIAL Nameservers* is higher, then add SOA MNAME name
+             server name and IP address to the *MNAME Not Master* set, and go to next
+             SOA SERIAL (from the *MNAME Nameservers* set).
+          2. Go to next SOA SERIAL.
+    2. Add SOA MNAME name server name and IP address to the *MNAME Is Master* set.
+    3. Go to next SOA SERIAL.
```
matsduf, line 187: In 11.1.1 the two SOA SERIALs are compared. If the one from *SERIAL Nameservers* is higher then the loop goes to the next SOA SERIAL from *MNAME Nameservers* (11.1.1.1), else the loop goes to the next SOA SERIAL from *SERIAL Nameservers* (11.1.1.2). I do not see how you come to 11.1.2.

I suggest that 11.1.1.2 is changed from "Go to next SOA SERIAL" to "Else, go to next SOA SERIAL (from the *SERIAL Nameservers* set)".

I think 11.1.3 could be removed.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,262 @@
...
+    terminate these procedures.
+
+11. For each SOA SERIAL in *MNAME Nameservers* do:
+    1. For each SOA SERIAL in *SERIAL Nameservers* do:
+       1. Compare both SOA SERIAL values (using the arithmetic in [RFC1982]).
+          1. If the one from *SERIAL Nameservers* is higher, then add SOA MNAME name
+             server name and IP address to the *MNAME Not Master* set, and go to next
+             SOA SERIAL (from the *MNAME Nameservers* set).
+          2. Go to next SOA SERIAL.
+    2. Add SOA MNAME name server name and IP address to the *MNAME Is Master* set.
```
matsduf, line 186: Maybe it should be clarified in 11.1.2 that SOA MNAME here comes from the MNAME Nameservers set, and that it might be multiple name server names and multiple IP addresses per name.

docs/specifications/tests/Zone-TP/zone01.md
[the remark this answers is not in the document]
matsduf: That is covered by [Consistency01], isn't it?

[Consistency01]: https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/Consistency-TP/consistency01.md

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,262 @@
...
+      *Child Nameservers* set, then output *[Z01_MNAME_NOT_IN_NS_LIST]* with
+      SOA MNAME name server name.
+   2. Do a [DNS Lookup] of SOA MNAME name server name (A and AAAA) and add the
+      SOA MNAME name server IP address(es) found to the *MNAME Nameservers* set
+      for the same entry as the SOA MNAME name server name.
+   3. If at least one IP address from the previous step was found, then:
+      1. For each SOA MNAME name server IP for the SOA MNAME name server name do: 
+         1. If the IP address is a localhost address (127.0.0.1 or ::1), then output
+            *[Z01_MNAME_HAS_LOCALHOST_ADDR]* with SOA MNAME name server name and IP.
+         2. Else, send *SOA Query* to the name server and collect the response.
```
matsduf, line 155: Maybe clarify: "to the name server" --> "to the name server IP address"

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,262 @@
...
+            1. If there is a DNS response, with [RCODE Name] "NoError" and
+               an SOA record in the answer section, then:
+               1. If the AA flag is not set, then output *[Z01_MNAME_NOT_AUTHORITATIVE]* 
+                  with SOA MNAME name server name and IP address, and name server(s) IP
+                  address(es) from the *SERIAL Nameservers* set.
+               2. Else, add the SOA SERIAL field to the *MNAME Nameservers* set
```
matsduf, line 161: "SOA SERIAL field" --> "SOA SERIAL value"?

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,262 @@
...
+
+10. If there is no SOA SERIAL in the *MNAME Nameservers* set, then
+    terminate these procedures.
+
+11. For each SOA SERIAL in *MNAME Nameservers* do:
+    1. For each SOA SERIAL in *SERIAL Nameservers* do:
+       1. Compare both SOA SERIAL values (using the arithmetic in [RFC1982]).
+          1. If the one from *SERIAL Nameservers* is higher, then add SOA MNAME name
+             server name and IP address to the *MNAME Not Master* set, and go to next
+             SOA SERIAL (from the *MNAME Nameservers* set).
```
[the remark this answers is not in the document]
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone01.md
tgreenx: I'm not sure it is needed, as the other set in this part of the procedure (SERIAL Nameservers) does not contain any SOA MNAME data anyway. But sure I'll add it.

> 2. Name server IP address, SOA SERIAL ("SERIAL Nameservers")

docs/specifications/tests/Zone-TP/zone01.md
tgreenx: There is no 11.1.2. I assume you talk about 11.2. You go to 11.2 when you keep going to 11.1.1.2, during the inner loop in 11.1. You restart from 11.1 when you reach 11.1.1.1 (I think the confusion comes from that, in my mind this is the logic I use), with the next *MNAME Namservers* SERIAL value. Thus 11.1.3 is also necessary to loop through all values in the *MNAME Nameservers* set (step 11).

Is it unclear this way? Should I try and reword it all?

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,262 @@
...
+Z01_MNAME_HAS_LOCALHOST_ADDR  | WARNING | nsname, ns_ip         | SOA MNAME name server "{nsname}" resolves to a localhost IP address ({ns_ip}).
+Z01_MNAME_IS_DOT              | NOTICE  | ns_ip_list            | SOA MNAME is specified as "." which usually means "no server". Fetched from name servers "{ns_ip_list}".
+Z01_MNAME_IS_LOCALHOST        | WARNING | ns_ip_list            | SOA MNAME name server is "localhost", which is invalid. Fetched from name servers "{ns_ip_list}".
+Z01_MNAME_IS_MASTER           | DEBUG   | ns, ns_ip_list        | SOA MNAME name server "{ns}" does appear to be master. Fetched from name servers "{ns_ip_list}".
+Z01_MNAME_MISSING_SOA_RECORD  | WARNING | ns, ns_ip_list        | SOA MNAME name server "{ns}" reponds to an SOA query with no SOA records in the answer section. Fetched from name servers "{ns_ip_list}".
+Z01_MNAME_NO_RESPONSE         | WARNING | ns, ns_ip_list        | SOA MNAME name server "{ns}" does not respond to an SOA query. Fetched from name servers "{ns_ip_list}".
+Z01_MNAME_NOT_AUTHORITATIVE   | WARNING | ns, ns_ip_list        | SOA MNAME name server "{ns}" is not authoritative for the zone. Fetched from name servers "{ns_ip_list}".
```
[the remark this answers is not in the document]
matsduf: As I read the message the name servers "ns_ip_list" says that "ns" is not authoritative for the zone.

I see what you want to achieve, but I think the message would be too long. If Consistency is updated it could report which name server is reporting which MNAME in case of inconsistency.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,266 @@
...
+:---------------------------- |:--------|:----------------------|:---------------------------------------------------------------------------------------
+Z01_MNAME_HAS_LOCALHOST_ADDR  | WARNING | nsname, ns_ip         | SOA MNAME name server "{nsname}" resolves to a localhost IP address ({ns_ip}).
+Z01_MNAME_IS_DOT              | NOTICE  | ns_ip_list            | SOA MNAME is specified as "." which usually means "no server". Fetched from name servers "{ns_ip_list}".
+Z01_MNAME_IS_LOCALHOST        | WARNING | ns_ip_list            | SOA MNAME name server is "localhost", which is invalid. Fetched from name servers "{ns_ip_list}".
+Z01_MNAME_IS_MASTER           | DEBUG   | ns, ns_ip_list        | SOA MNAME name server "{ns}" appears to be master. Fetched from name servers "{ns_ip_list}".
+Z01_MNAME_MISSING_SOA_RECORD  | WARNING | ns, ns_ip_list        | SOA MNAME name server "{ns}" reponds to an SOA query with no SOA records in the answer section. Fetched from name servers "{ns_ip_list}".
+Z01_MNAME_NO_RESPONSE         | WARNING | ns, ns_ip_list        | SOA MNAME name server "{ns}" does not respond to an SOA query. Fetched from name servers "{ns_ip_list}".
```
matsduf, line 85: `Fetched from name servers "{ns_ip_list}".` should be removed. The query was sent to the IP address of `ns`.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,266 @@
...
+             server name and IP address (of the current SOA SERIAL value from
+             the *MNAME Nameservers* set) to the *MNAME Not Master* set, and go 
+             to next SOA SERIAL (from the *MNAME Nameservers* set).
+          2. Go to next SOA SERIAL.
+    2. Add SOA MNAME name server name and IP address (from the *MNAME Nameservers* set) to the *MNAME Is Master* set.
+    3. Go to next SOA SERIAL.
+
+12. If the set *MNAME Not Master* is non-empty, then output *[Z01_MNAME_NOT_MASTER]*
+    with SOA MNAME name server name(s) and IP address(es) from the set, and name server(s)
+    IP address(es) from the *SERIAL Nameservers* set.
```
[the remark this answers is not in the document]
tgreenx: Updated

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,262 @@
...
+Z01_MNAME_IS_LOCALHOST        | WARNING | ns_ip_list            | SOA MNAME name server is "localhost", which is invalid. Fetched from name servers "{ns_ip_list}".
+Z01_MNAME_IS_MASTER           | DEBUG   | ns, ns_ip_list        | SOA MNAME name server "{ns}" does appear to be master. Fetched from name servers "{ns_ip_list}".
+Z01_MNAME_MISSING_SOA_RECORD  | WARNING | ns, ns_ip_list        | SOA MNAME name server "{ns}" reponds to an SOA query with no SOA records in the answer section. Fetched from name servers "{ns_ip_list}".
+Z01_MNAME_NO_RESPONSE         | WARNING | ns, ns_ip_list        | SOA MNAME name server "{ns}" does not respond to an SOA query. Fetched from name servers "{ns_ip_list}".
+Z01_MNAME_NOT_AUTHORITATIVE   | WARNING | ns, ns_ip_list        | SOA MNAME name server "{ns}" is not authoritative for the zone. Fetched from name servers "{ns_ip_list}".
+Z01_MNAME_NOT_IN_NS_LIST      | INFO    | nsname                | SOA MNAME name server "{nsname}" is not listed as NS record for the zone.
+Z01_MNAME_NOT_MASTER          | WARNING | ns, ns_ip_list        | SOA MNAME name server "{ns}" does not have the highest SOA SERIAL. Fetched from name servers "{ns_ip_list}".
```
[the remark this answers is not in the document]
matsduf: Or `soaserial` and `soaserial_list` to make is shorter and to avoid unnecessary underscore (cf. `mailtarget`). The update of "Argument" could be done in parallel. It does not have to be done before.

Another possibility is to define generic `integer` and `integer_list`, respectively. There are pros and cons and I think both will work.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,264 @@
...
+    1. For each SOA SERIAL in *SERIAL Nameservers* do:
+       1. Compare both SOA SERIAL values (using the arithmetic in [RFC1982]).
+          1. If the one from *SERIAL Nameservers* is higher, then add SOA MNAME name
+             server name and IP address of the current SOA SERIAL value from
+             the *MNAME Nameservers* set to the *MNAME Not Master* set, and go 
+             to next SOA SERIAL (from the *MNAME Nameservers* set).
+          2. Go to next SOA SERIAL (from the *SERIAL Nameservers* set).
+    2. Add SOA MNAME name server name and IP address (from the
+       *MNAME Nameservers* set) to the *MNAME Is Master* set.
+    3. Go to next SOA SERIAL (from the *SERIAL Nameservers* set).
```
matsduf, line 191: Why have a sub-level on line 183? The following would work as well and will give one level less and be easier to read:
```
    1. For each SOA SERIAL in *SERIAL Nameservers* do:
       1. Compare both SOA SERIAL values (using the arithmetic in [RFC1982]).
       2. If the one from *SERIAL Nameservers* is higher, then add SOA MNAME name
          server name and IP address of the current SOA SERIAL value from (...)
```
The `Go to next SOA SERIAL` on lines 188 and 191 are not strictly needed. If they contribute to clarity they should be kept, else not.

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,264 @@
...
+* "Child Zone" - The domain name to be tested.
+
+## Summary
+
+Message Tag                   | Level   | Arguments             | Message ID for message tag
+:---------------------------- |:--------|:----------------------|:---------------------------------------------------------------------------------------
+Z01_MNAME_HAS_LOCALHOST_ADDR  | WARNING | nsname, ns_ip         | SOA MNAME name server "{nsname}" resolves to a localhost IP address ({ns_ip}).
+Z01_MNAME_IS_DOT              | NOTICE  | ns_ip_list            | SOA MNAME is specified as "." which usually means "no server". Fetched from name servers "{ns_ip_list}".
+Z01_MNAME_IS_LOCALHOST        | WARNING | ns_ip_list            | SOA MNAME name server is "localhost", which is invalid. Fetched from name servers "{ns_ip_list}".
+Z01_MNAME_IS_MASTER           | DEBUG   | ns                    | SOA MNAME name server "{ns}" appears to be master.
```
matsduf, line 83: I guess it should be `ns_list` since when it is outputted there could be multiple `ns`:
```
13. If the set *MNAME Is Master* is non-empty, then output *[Z01_MNAME_IS_MASTER]*
    with SOA MNAME name server name(s) and IP address(es) from the set.
```

docs/specifications/tests/Zone-TP/zone01.md
```diff
@@ -3,43 +3,264 @@
...
+Z01_MNAME_MISSING_SOA_RECORD  | WARNING | ns                    | SOA MNAME name server "{ns}" reponds to an SOA query with no SOA records in the answer section.
+Z01_MNAME_NO_RESPONSE         | WARNING | ns                    | SOA MNAME name server "{ns}" does not respond to an SOA query.
+Z01_MNAME_NOT_AUTHORITATIVE   | WARNING | ns                    | SOA MNAME name server "{ns}" is not authoritative for the zone.
+Z01_MNAME_NOT_IN_NS_LIST      | INFO    | nsname                | SOA MNAME name server "{nsname}" is not listed as NS record for the zone.
+Z01_MNAME_NOT_MASTER          | WARNING | ns                    | SOA MNAME name server "{ns}" does not have the highest SOA SERIAL (expected "{soa_serial}" at most but got "{soa_serial_list}")
```
matsduf, line 88: I guess it should be `ns_list` since when it is outputted there could be multiple `ns`:
```
12. If the set *MNAME Not Master* is non-empty, then output *[Z01_MNAME_NOT_MASTER]*
    with SOA MNAME name server name(s) and IP address(es) from the set.
```

There is `soa_serial` and `soa_serial_list` in msgid, but neither in the arguments column nor when the message is outputted.

I am not sure how to interpret `expected "{soa_serial}" at most but got "{soa_serial_list}"`. Could you reword? Is this what it is meant: `expected "{soa_serial}" but got "{soa_serial_list}"`?

docs/specifications/tests/Zone-TP/zone01.md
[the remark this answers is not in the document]
tgreenx: Created PR https://github.com/zonemaster/zonemaster-engine/pull/1138. I went for the first option.

docs/specifications/tests/Zone-TP/zone01.md
tgreenx: You read it correctly. This was here to emphasize on the type of check that is being done in step 11.2.1, but, I think that it can be removed too. Updated.

matsduf (approved):

PR #1033, opened 2022-01-05
First version of Test Case Specification Status

matsduf:
## Purpose

The test case specification format has evolved to more explicit. This PR adds an internal document that will show the status of the specifications and gives an overview of the status.

This must be updated when specifications are updated or added.

### DIFF
--- a/docs/internal-documentation/status/TestCaseSpecificationStatus.md
+++ b/docs/internal-documentation/status/TestCaseSpecificationStatus.md
@@ -1,11 +1,9 @@
-# Test Case Specification Status
-
 The test case specifications have evolved in format. The new or updated
 specifications have explicit and detailed steps to execute the test case, the
 message tags are defined in the specification and the format follows the template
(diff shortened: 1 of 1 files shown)


tgreenx (dismissed):
Thanks Mats, it will be an useful document for future tracking of test specification updates.

tgreenx:
Moreover, I think it would be useful to add another column regarding the use of [MethodsNT](https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/MethodsNT.md).

matsduf:
> Moreover, I think it would be useful to add another column regarding the use of [MethodsNT](https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/MethodsNT.md).

Good suggestion. I have added that in the Tracking Issue that I moved the document to (#1051).

matsduf:
The content of this PR has been moved to an issue of a special type, a Tracking Issue. See #1051.

PR #1041, opened 2022-03-14
Specify the ignoring of some broken DNSKEY RRs

### LINKED ISSUE #1040: Handling DNSKEY RRs with broken public key field
We have some test cases that specify that the key size of DNSKEY RRs should be calculated. When we receive valid DNSKEY RRs this is not a problem, but when we receive DNSKEY RRs with an empty (or otherwise too short) public key field the situation is unspecified. The key size for these broken DNSKEY RRs simply cannot be calculated.

I could think of three possible solutions to this:
1. Specify that Zonemaster should determine the key size of a DNSKEY RR with insufficient data to be 0.
2. Specify that Zonemaster should ignore DNSKEY RRs with insufficient data.
3. Specify that Zonemaster should ignore DNSKEY RRs with insufficient data, but only when calculating the key size.

Test cases DNSSEC03 and DNSSEC14 explicitly depends on the calculation of the key size. The current implementation of DNSSEC05 also has this problem because it outputs a KEY_DETAILS message with a key size field, apparently for diagnostic reasons. When the public key field is empty the size of the key cannot be calculated. Also most of the DNSSEC test cases send DNSKEY queries so the evaluation of those might change as well depending on the choice of solution. We should probably prioritize maximizing correctness over minimizing change in behavior.

mattias-p:
## Purpose

This PR specifies a situation when certain broken DNSKEY RRs should be ignored.

## Context

Fixes #1040.

## Changes

Updates the DNS response handling specification.

## How to test this PR

N/A

### DIFF
--- a/docs/specifications/tests/DNSQueryAndResponseDefaults.md
+++ b/docs/specifications/tests/DNSQueryAndResponseDefaults.md
@@ -120,13 +120,15 @@ response to be considered to be a DNS response.
 * Owner name and record type of a DNS record are compared, by default, against
   *Query Name* and *Query Type* in the question section in the query, not in the
   response.
-  
+
 * When fetching records from the answer section, these are the default criteria:
   * Only records matching *Query Name* and *Query Type* are fetched. Any
     RRSIG records meeting the next criterion are also fetched.
   * RRSIG records matching *Query Name* and covering *Query Type* are
     fetched.
   * CNAME records are ignored unless *Query Type* is CNAME.
+  * DNSKEY records are ignored if their public key field does not contain enough
+    data to calculate the size of the key.
 
 * When the test case specification states that a CNAME chain is to be followed,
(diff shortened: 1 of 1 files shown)


matsduf (approved):

mattias-p:
I'm not so sure about change this anymore. After some more consideration I'm leaning towards the position that responses should either be returned in their entirety or not at all. If we return responses to the test case layer with some RRs kept and some RRs removed, that seems to me like it would muddy the waters.

When it comes to responses containing DNSKEY RRs with insufficient data in their public key field, it's probably better to return those responses in their entirety and deal with the broken field contents at the test case level.

If it turn out we want to prune individual RRs afterall we can reopen this PR.

matsduf:
A new issue, #1055, to create a test case to capture some errors in the DNSKEY.

PR #1073, opened 2022-06-22
Removes special treatment of SHA1 digest

matsduf:
## Purpose

There is no need to have special treatment of SHA1 digests. It is deprecated since years back. The special treatment is removed from the specification.

## How to test this PR

This is documentation only, but this should also be implemented for the test case in Zonemaster-Engine.


tgreenx (approved):

matsduf:
@tgreenx, is this something you could implement? If so, could I merge and create an issue in Engine?

PR #1083, opened 2022-07-15
Updates Template and identifier specification

matsduf:
## Purpose

* Updated the table header in the outcomes section in the template to refer to message tag and message ID, respectively.
* Updated the template to prescribe that message ID (msgid) is defined in the test case specification, not just a description of the message ID.
* Updated the template to prescribe that the arguments are included in the message ID in the table.
* Corrected the template to use all-uppercase identifier.
* Corrected the identifier specification that the identifier in the test case specification should be in upper case.
* Updates the text and reference for "in-bailiwick".
* Adds text and reference for cases when RCODE is referred to.
* Regrouped and reformatted the link section.

The first step in the development of the test case specification was to include the message tags and the default level, i.e. let the specification define the message tags for the test case and the default level for the message. This is the second step to make the specification define the message ID including the arguments. The message ID in the implementation will then be a copy of the defined message ID.

In the third step, not in this PR, the message, level, arguments and message ID will be put into a database. The specification will then be generated to include that information.

## Changes

Test case template and test case identifier specification.

## How to test this PR

Documentation only.

### DIFF
--- a/docs/internal-documentation/templates/specifications/tests/TestCaseIdentifierSpecification.md
+++ b/docs/internal-documentation/templates/specifications/tests/TestCaseIdentifierSpecification.md
@@ -32,6 +32,6 @@ Example of test case identifiers:
 * DNSSEC15
 * ZONE06
 
-One execptional test case identifier has been `BASIC00`, in which the index
+One exceptional test case identifier is `BASIC00`, in which the index
 suffix is out of range of the stated range above. That test case is planned to
 be removed.
--- a/docs/internal-documentation/templates/specifications/tests/TestCaseIdentifierSpecification.md
+++ b/docs/internal-documentation/templates/specifications/tests/TestCaseIdentifierSpecification.md
@@ -13,20 +13,25 @@ test level's name. The following test levels are defined and available:
 * Syntax
 * Zone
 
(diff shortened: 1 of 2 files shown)


docs/internal-documentation/templates/specifications/tests/TestCaseIdentifierSpecification.md
```diff
@@ -13,20 +13,25 @@ test level's name. The following test levels are defined and available:
...
-* Basic04
+* ADDRESS03
+* BASIC04
 * DNSSEC15
-* Zone06
+* ZONE06
 
-One execptional test case identifier is `Basic00`, in which the index suffix is
-out of range of the stated range above.
+One execptional test case identifier has been `BASIC00`, in which the index
```
tgreenx, line 35: execptional -> exceptional

tgreenx (approved):

docs/internal-documentation/templates/specifications/tests/TestCaseIdentifierSpecification.md
matsduf: Fixed.

tgreenx (approved):

Issue #573, opened 2018-03-01
Update outdated design docs


JohanIIS:
Motion to update, move, remove or extract vital information from the following files:

* [/docs/design/Architecture Notes.md](https://github.com/dotse/zonemaster/blob/master/docs/design/Architecture%20Notes.md)
* [/docs/design/Implementation Guidelines.md](https://github.com/dotse/zonemaster/blob/master/docs/design/Implementation%20Guidelines.md)

They contain good information but have been left unfinished since 2013/2014 (one file even ends with "to be continued"). They should either be resurrected and updated, moved to an archive folder, removed entirely or have the information within saved in another document.

sandoche2k:
@JohanIIS Should it be part of the public github?

mattias-p:
I suggest that the files are left where they are, but that a clear note about the documents being out of date is added to the top of the documents.

matsduf:
I disagree. This issue is to do something meaningful with the documents. In the mean time we could put a note that they need update, but this issue should not be closed until we have updated documents or used the relevant information in the documents in other documents and deleted these documents.

sandoche2k:
+1

matsduf:
F2F: update or remove.

matsduf:
#1087 proposes to remove the old documents.

PR #1087, opened 2022-07-25
Removes old design documents

matsduf:
## Purpose

As commented in #573 there are two old design documents that were never completed and never updated. They are not used.

* [/docs/design/Architecture Notes.md](https://github.com/dotse/zonemaster/blob/master/docs/design/Architecture%20Notes.md)
* [/docs/design/Implementation Guidelines.md](https://github.com/dotse/zonemaster/blob/master/docs/design/Implementation%20Guidelines.md)

This PR removes the documents and updates the README file listing them.

## How to test this PR

Documentation only.


mattias-p:
I read through these documents and I think they actually reflect the state of the system really well, even though they haven't been touched for years. They seem fairly timeless to me. I suggest we resurrect them rather than delete them. I can take it upon myself to make the updates.

Easy updates:
* They still refer to Zonemaster even though we renamed the component to Zonemaster Engine.
* The performance non-goal should be upgraded to a proper goal after our last F2F, though it should remain at a lower priority than the three original goals.

Practices we should resurrect:
* The Guideline document speaks of perltidy and perlcritic, and we actually have the config files for these still in the zonemaster-engine root! We should really start using them. Perhaps the config files need some minimal tweaking before we can take them up again, I'm not sure. Then we should integrate checks for them in our CI.

Items for discussion:
* The Architecture document speak of an Information Object and the Guideline document has an "Exposing test metadata" section. They seem to be about different sides of the same coin. This isn't really implemented as an object today, and I'm not sure of its usefulness. This is probably a good topic for a WG meeting.
* The Guideline document has a "Returning information from tests, the practice" section. I'm sure we have things to tell on this subject and it might actually be a healthy thing to actually do. However I can't just spit out an outline for this. Maybe we could brainstorm it at a WG meeting?

What do you guys say? Delete or resurrect? And are there any other noteworthy things in the documents that I didn't list above?

matsduf:
Close this and maybe open a new one.

PR #1089, opened 2022-07-28
Test case Zone09: always output found MX records on success

pnax:
## Purpose

PR #870 does not explicitly ask for a message to be outputted when all nameservers answer with the same RR set of MXs. Such message could have value since it lists all configured MX records. This is about updating the spec so that Zonemaster outputs an INFO message with the list of configured MX on success.

## Context

Follow-up on #870 

## Changes

* Test case Zone09

## How to test this PR

When the test case succeeds and when there are some MX records, a message Z09_MX_DATA of level INFO with the list of found records should be outputted.

### DIFF
--- a/docs/specifications/tests/Zone-TP/zone09.md
+++ b/docs/specifications/tests/Zone-TP/zone09.md
@@ -208,10 +208,11 @@ queries follow, unless otherwise specified below, what is specified for
              of MX records.
           2. If the preference of the [Null MX] is non-zero then output
              *[Z09_NULL_MX_NON_ZERO_PREF]*.
-       2. If *Child Zone* is a [TLD] with a [non-Null MX][Null MX] then
+       2. Else, if *Child Zone* is a [TLD] with a [non-Null MX][Null MX] then
           output *[Z09_TLD_EMAIL_DOMAIN]*.
-       3. If *Child Zone* is the root zone with a [non-Null MX][Null MX] then
+       3. Else, if *Child Zone* is the root zone with a [non-Null MX][Null MX] then
           output *[Z09_ROOT_EMAIL_DOMAIN]*.
+       4. Else, output *[Z09_MX_DATA]*.
 
 11. Else, if the *No MX RRset* set is non-empty then do:
     * Output *[Z09_MISSING_MAIL_TARGET]* unless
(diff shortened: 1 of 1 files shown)


docs/specifications/tests/Zone-TP/zone09.md
```diff
@@ -212,6 +212,7 @@ queries follow, unless otherwise specified below, what is specified for
           output *[Z09_TLD_EMAIL_DOMAIN]*.
        3. If *Child Zone* is the root zone with a [non-Null MX][Null MX] then
           output *[Z09_ROOT_EMAIL_DOMAIN]*.
+       4. If no error are encountered, output *[Z09_MX_DATA]*.
```
matsduf, line 215: It is unclear what "no error" is. I rather suggest:
```
       1. If the mailtarget of any of the MX records in the RRset in *MX RRset*
          is a [Null MX] then do:
          1. If there are more than one record in the MX RRset, then output
             *[Z09_NULL_MX_WITH_OTHER_MX]* with the mail targets from the RDATA
             of MX records.      
          2. If the preference of the [Null MX] is non-zero then output
             *[Z09_NULL_MX_NON_ZERO_PREF]*.
       2. Else, if *Child Zone* is a [TLD] then output *[Z09_TLD_EMAIL_DOMAIN]*.
       3. Else, if *Child Zone* is the root zone then output *[Z09_ROOT_EMAIL_DOMAIN]*.
       4. Else, output *[Z09_MX_DATA]*.
```

matsduf (commented):
I do not see much value of the message in a Zonemaster context, but I will not do any specific harm either besides contributing to the amount of INFO messages that Zonemaster outputs.

I have earlier contributed to too many messages, and now I try to keep the amount down, messages on all levels.

docs/specifications/tests/Zone-TP/zone09.md
```diff
@@ -208,10 +208,11 @@ queries follow, unless otherwise specified below, what is specified for
              of MX records.
           2. If the preference of the [Null MX] is non-zero then output
              *[Z09_NULL_MX_NON_ZERO_PREF]*.
-       2. If *Child Zone* is a [TLD] with a [non-Null MX][Null MX] then
+       2. Else, if *Child Zone* is a [TLD] with a [non-Null MX][Null MX] then
           output *[Z09_TLD_EMAIL_DOMAIN]*.
-       3. If *Child Zone* is the root zone with a [non-Null MX][Null MX] then
+       3. Else, if *Child Zone* is the root zone with a [non-Null MX][Null MX] then
           output *[Z09_ROOT_EMAIL_DOMAIN]*.
+       4. Else, output *[Z09_MX_DATA]*.
```
matsduf, line 215: `Z09_MX_DATA` requires some arguments. I suggest:
```
4. Else, output *[Z09_MX_DATA]* with the mail targets from the RDATA and the associated name server IP addresses in the set.
```

matsduf (changes requested):

tgreenx (approved):

PR #1090, opened 2022-08-02
Update basic04

matsduf:
## Purpose

The purpose of this PR is to add a specific message when IPv4 or IPv6 has been disabled. The default message is too verbose and to be kept on DEBUG level.

Editorial updates have also been done. Below it is specified what changes have any impact on the implementation in Zonemaster::Engine.

* Updated specification - b729a2d96408 - *This update has no impact on implementation.*
  * Updates to match template.
  * No logical changes.
  * No changes that requires updated implementation.
  * msgid has been copied from implementation.
* Editorial update: adjusted link list - 07089023013d7 - *This update has no impact on implementation.*
* Adds reference to "DNS Query and Response Defaults" but no change of logic. - a001b793184b80 - *This update has no impact on implementation.*
* Adds specific messages when IPv4 or IPv6 has been disabled. - b5d62a2a4 - **This update requires an update to the implementation.**

## Context

zonemaster/zonemaster-engine/pull/790 increased the verbosity of the standard messages, which makes it necessary to move them to the DEBUG level also for Basic.

## How to test this PR

This is a document change, and the resulting update in implementation should be tested.

### DIFF
--- a/docs/specifications/tests/Basic-TP/basic04.md
+++ b/docs/specifications/tests/Basic-TP/basic04.md
@@ -3,6 +3,20 @@
 ## Test case identifier
 **BASIC04**
 
+
+## Table of contents
+
+* [Objective](#Objective)
+* [Scope](#Scope)
+* [Inputs](#Inputs)
+* [Summary](#Summary)
+* [Test procedure](#Test-procedure)
+* [Outcome(s)](#Outcomes)
+* [Special procedural requirements](#Special-procedural-requirements)
+* [Intercase dependencies](#Intercase-dependencies)
+* [Terminology](#terminology)
+
+
 ## Objective
 
 Many test cases will query the name servers in the delegation or the
@@ -23,15 +37,44 @@ report problems found in the following areas:
 * Name Server responding with unexpected RCODE (any except "NOERROR")
   on query for SOA or NS for *Child Zone*.
 
+
+## Scope
+
 This test case is expected to run before [Basic02] but neither its
 execution nor outcome are dependent on the order.
+
 
 ## Inputs
 
 * "Child Zone" - The domain name to be tested.
 
 
-## Ordered description of steps to be taken to execute the test case
+## Summary
+
+Message Tag                       | Level   | Arguments            | Message ID for message tag
+:---------------------------------|:--------|:---------------------|:---------------------------------------------------------------------------------
+B04_MISSING_NS_RECORD             | WARNING | ns                   | Nameserver {ns} reponds to a NS query with no NS records in the answer section.
+B04_MISSING_SOA_RECORD            | WARNING | ns                   | Nameserver {ns} reponds to a SOA query with no SOA records in the answer section.
+B04_NO_RESPONSE                   | WARNING | ns                   | Nameserver {ns} does not respond over neither UDP nor TCP.
+B04_NO_RESPONSE_NS_QUERY          | WARNING | ns                   | Nameserver {ns} does not respond to NS queries.
+B04_NO_RESPONSE_SOA_QUERY         | WARNING | ns                   | Nameserver {ns} does not respond to SOA queries.
+B04_NS_RECORD_NOT_AA              | WARNING | ns                   | Nameserver {ns} does not give an authoritative response on an NS query.
+B04_RESPONSE_TCP_NOT_UDP          | WARNING | ns                   | Nameserver {ns} does not respond over UDP.
+B04_SOA_RECORD_NOT_AA             | WARNING | ns                   | Nameserver {ns} does not give an authoritative response on an SOA query.
+B04_UNEXPECTED_RCODE_NS_QUERY     | WARNING | ns                   | Nameserver {ns} responds with an unexpected RCODE ({rcode}) on an NS query.
+B04_UNEXPECTED_RCODE_SOA_QUERY    | WARNING | ns                   | Nameserver {ns} responds with an unexpected RCODE ({rcode}) on an SOA query.
+B04_WRONG_NS_RECORD               | WARNING | ns                   | Nameserver {ns} responds with a wrong owner name ({owner} instead of {name}) on NS queries.
+B04_WRONG_SOA_RECORD              | WARNING | ns                   | Nameserver {ns} responds with a wrong owner name ({owner} instead of {name}) on SOA queries.
+
+The value in the Level column is the default severity level of the message. The
+severity level can be changed in the [Zonemaster-Engine profile]. Also see the
+[Severity Level Definitions] document.
+
+The argument names in the Arguments column lists the arguments used in the
+message. The argument names are defined in the [argument list].
+
+
+## Test procedure
 
 1. Create a SOA query for the *Child Zone* without any OPT record (no EDNS).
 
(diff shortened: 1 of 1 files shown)


docs/specifications/tests/Basic-TP/basic04.md
```diff
@@ -23,46 +37,97 @@ report problems found in the following areas:
 * Name Server responding with unexpected RCODE (any except "NOERROR")
```
tgreenx, line 37: RCODE -> [RCODE Name] (with link to IANA RCODE Name page)
"NOERROR" -> "NoError"

tgreenx: This also applies in the Test Procedure below where I can see two other occurrences

docs/specifications/tests/Basic-TP/basic04.md
```diff
@@ -3,6 +3,20 @@
...
+* [Test procedure](#Test-procedure)
+* [Outcome(s)](#Outcomes)
+* [Special procedural requirements](#Special-procedural-requirements)
+* [Intercase dependencies](#Intercase-dependencies)
+* [Terminology](#terminology)
+
+
 ## Objective
 
 Many test cases will query the name servers in the delegation or the
```
tgreenx, line 22: SOA record or both. Reporting problem is crucial -> SOA record, or both. Reporting problems is crucial

tgreenx (commented):
LGTM

docs/specifications/tests/Basic-TP/basic04.md
matsduf: Fixed three places plus added link in link section.

docs/specifications/tests/Basic-TP/basic04.md
matsduf: Fixed.

docs/specifications/tests/Basic-TP/basic04.md
```diff
@@ -34,7 +34,7 @@ report problems found in the following areas:
   section in the response on an NS query for *Child Zone*.
 * Name Server not setting the AA flag in a response with SOA or NS in
   answer section.
-* Name Server responding with unexpected RCODE (any except "NOERROR")
+* Name Server responding with unexpected [RCODE Name] (any except "NeError")
```
tgreenx, line 37: NeError -> NoError

docs/specifications/tests/Basic-TP/basic04.md
```diff
@@ -21,7 +21,7 @@
 
 Many test cases will query the name servers in the delegation or the
 name servers appointed by the NS records in the zone for the NS or SOA
-record or both. Reporting problem is crucial, but instead of letting
+record, or both. Reporting problem is crucial, but instead of letting
```
tgreenx, line 24: Reporting problem -> Reporting problems

matsduf: I do not like "problems is" so I rephrased it.

docs/specifications/tests/Basic-TP/basic04.md
matsduf: Fixed.

tgreenx (approved):

docs/specifications/tests/Basic-TP/basic04.md
```diff
@@ -20,105 +34,166 @@ report problems found in the following areas:
...
-   2. If there is no DNS response on neither query (UDP), then:
-      1. Send the SOA query over TCP to the name server and collect the
-         response.
-      2. If there is no DNS response on the TCP query, then output
-         *[B04_NO_RESPONSE]* and go to next server.
-      3. Else (there is a DNS response over TCP), then output 
-         *[B04_RESPONSE_TCP_NOT_UDP]* and go to next server.
+4. If IPv4 is disabled then do:
+   1. Extract all name servers with IPv4 address from *Name Server IP*.
+   2. If the set of IPv4 name serververs is non-empty then output
```
pnax, line 100: typo: "serververs" -> "servers"

docs/specifications/tests/Basic-TP/basic04.md
```diff
@@ -20,105 +34,166 @@ report problems found in the following areas:
...
+      *[B04_IPV4_DISABLED]* with the set of IPv4 name servers.
+
+5. If IPv6 is disabled then do:
+   1. Extract all name servers with IPv6 address from *Name Server IP*.
+   2. If the set of IPv6 name serververs is non-empty then output
```
pnax, line 105: typo: "serververs" -> "servers"

docs/specifications/tests/Basic-TP/basic04.md
```diff
@@ -20,105 +34,166 @@ report problems found in the following areas:
...
+      *[B04_IPV6_DISABLED]* with the set of IPv6 name servers.
+
+6. For each name server in *Name Server IP* do:
+
+   1. Send *SOA Query UDP* and *NS Query UDP* to the name server and collect
+      the [DNS Responses][DNS Response].
+   2. If there is no DNS response on neither query), then:
```
pnax, line 112: typo: closing parenthesis

docs/specifications/tests/Basic-TP/basic04.md
```diff
@@ -20,105 +34,166 @@ report problems found in the following areas:
...
+      1. Send *SOA Query TCP* to the name server and collect the [DNS Response].
+      2. If there is no [DNS Response], then output *[B04_NO_RESPONSE]* and go
+         to next server.
+      3. Else, then output *[B04_RESPONSE_TCP_NOT_UDP]* and go to next server.
```
pnax, line 116: Why going to next server and not checking the response over TCP in the rest of the procedure? It feels like the procedure is missing the case where the nameserver responds only over TCP with unexpected responses.

docs/specifications/tests/Basic-TP/basic04.md
matsduf: What a typo! Fixed.

docs/specifications/tests/Basic-TP/basic04.md
matsduf: Well, I guess copy and paste of the previous one. Fixed.

docs/specifications/tests/Basic-TP/basic04.md
matsduf: Removed.

docs/specifications/tests/Basic-TP/basic04.md
matsduf: A name server must respond over UDP, and DNS resolvers will generally send queries over UDP only, and only use TCP of the response was truncated (TC flag set). In the following test cases of Zonemaster all queries will be sent of UDP (one exceptional test case testing queries of TCP). I do not feel that adding tests of TCP will add anything substantial here. The server is unreachable, in a practical sense. The DNS operator must fix that.

After reading my response I felt that the message ID for B04_RESPONSE_TCP_NOT_UDP should probably be more explicit about the issue. Please review my update.

pnax: I understand however I still think we are missing an edge case (that could be easily addressed I think). But I won't bother if we agree this is fine like that.

matsduf: We could let the others express their opinion too. It is of course very possible to do what you suggest.

matsduf: I think it would make more sense to update [Connectivity02] ("TCP connectivity") to only test those name servers that do respond over UDP, but then test them thoroughly over TCP, both SOA and NS query, give correct rcode and have AA flag set.

[Connectivity02]: https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/Connectivity-TP/connectivity02.md

tgreenx: I feel like, since this test case is the base test case in testing a name server functionality, specifications (RFCs) should prevail rather than operational usage. I couldn't find a quote from an RFC (i.e [RFC1035 ](https://www.rfc-editor.org/rfc/rfc1035.html), [RFC7766](https://www.rfc-editor.org/rfc/rfc7766.html)) that clearly indicated that a name server should (or must) respond to UDP if it already responds on TCP. But your point is also valid @matsduf ("DNS resolvers will generally send queries over UDP only"). Thus I think we can proceed with @pnax suggestion and use that TCP response for further testing when UDP is unavailable, while still outputting a comprehensive message tag B04_RESPONSE_TCP_NOT_UDP to warn the user of possible reachability shortcomings.

matsduf: Supporting both UDP and TCP is mandatory. "Section 6.1.3.2 of RFC1123 is updated: All general-purpose DNS implementations MUST support both UDP and TCP transport." (https://www.rfc-editor.org/rfc/rfc7766#section-5)

tgreenx: Right, thanks, somehow I missed that...
From the same section ([Section 5](https://www.rfc-editor.org/rfc/rfc7766#section-5)), it follows with:
```
   Regarding the choice of when to use UDP or TCP, Section 6.1.3.2 of
   RFC 1123 also says:

      ... a DNS resolver or server that is sending a non-zone-transfer
      query MUST send a UDP query first.

   This requirement is hereby relaxed.  Stub resolvers and recursive
   resolvers MAY elect to send either TCP or UDP queries depending on
   local operational reasons.  TCP MAY be used before sending any UDP
   queries.  If the resolver already has an open TCP connection to the
   server, it SHOULD reuse this connection.  In essence, TCP ought to be
   considered a valid alternative transport to UDP, not purely a retry
   option.
```

My understanding is that a DNS response over TCP, independently from UDP, must still be considered valid. So it ought to count for something when it is found. Thus I stand with my previous opinion (i.e., use that TCP response for further testing and warn the user of major issue with lack of UDP support + likelihood of reachability issues that it entails)

matsduf: A DNS response over TCP is valid, if the client sent the query over TCP. The client can choose to send all queries over TCP, but the server must accept both. A DNS resolver must accept queries over both from its clients, but it can choose to send its queries, as client, over TCP.

The background to Basic04 is that many test cases outputted NO_RESPONSE (or similar) to the same name server when failing to connect. It is or was not very helpful to get the same message multiple times. In all these case the message was outputted because connection over **UDP** was failing. I then created Basic04 to give one such message, and since it replaced multiple messages over failing connection over UDP it is natural that Basic04 has a UDP focus. All testcases, with two exceptions (if I remember correctly), only queries over UDP.

I added the extra TCP test when connection over UDP failed, but it can be questioned if that is really relevant for the purpose of the test case. I added that to be helpful in case the name server had UDP blocked by a firewall.

I suggest that we remove the TCP test from Basic04 and then update update  [Connectivity02] ("TCP connectivity") and let it test all name servers:

1. If the name server has no UDP connection only look for TCP as Basic04 does today.
2. If the name server has UDP connection do a thorough test over TCP, both SOA and NS query, give correct RCODE and have AA flag set.

[Connectivity02]: https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/Connectivity-TP/connectivity02.md

matsduf:
Replaced by #1097, #1098 and #1099.

PR #1081, opened 2022-07-12
Add style guide item on autolinks

mattias-p:
## Purpose

Document how to keep our Markdown documentation compatible with tools other than Github.

## Context

Fixes #585.

## Changes

Adds an item to the style guide.

## How to test this PR

N/A

PR #1094, opened 2022-09-01
URL fragments referring to internal headings in lowercase

pnax:
## Purpose

When it comes to referring to internal anchors, we used a mix of upper and lower cases in our Markdown documents. It appears that Github handles this well, but other Markdown to HTML engines suffer to link to the related identifier on the page since they are looking for a perfect match (case sensitively).
This PR is about downcasing all fragments in our Markdown documents.

## Context

n/a

## Changes

Downcase all fragments that refer to internal anchors in our Markdown documents.

## How to test this PR

Documentation only. Links with the updated fragments should work properly once rendered.


tgreenx (approved):
Fine with me. Template01 gets updated too so it should be fine for future specification updates.

matsduf:
Maybe [Style Guide] should be updated too? (There is an open RP on that #1081.)

[Style Guide]: https://github.com/zonemaster/zonemaster/blob/develop/docs/internal-documentation/maintenance/StyleGuide.md

matsduf (approved):

pnax:
> Maybe [Style Guide](https://github.com/zonemaster/zonemaster/blob/develop/docs/internal-documentation/maintenance/StyleGuide.md) should be updated too? (There is an open RP on that https://github.com/zonemaster/zonemaster/pull/1081.)

Good point, I've provided an update of the document in a separate PR, see #1095.

PR #1099, opened 2022-09-14
Replace basic04 with connectivity01

matsduf:
## Purpose

This PR should be merged when #1097 has been merged.

This PR replaces all references to Basic04 with references to Connectivity01, and removes Basic04.

The test case template is also updated.

## How to test this PR

Testing is done through review.

PR #1098, opened 2022-09-14
Rewrite of Connectivity02 (TCP connectivity)

matsduf:
## Purpose

This PR is a complete rewrite of the test case. The main body of the the specification is copied from the rewritten Connectivity01 (#1097). 

## How to test this PR

This documentation only. Testing is done by reviewing and manual running the test case.

PR #1100, opened 2022-09-28
Add specifications for testing SPF and DMARC policies

### LINKED ISSUE #995: Add SPF and possibly other mail related test cases [Feature Request]
Add, into, perhaps ZONE (or a new MX section) a test for SPF (using the SPF RR  as well as the TXT RR) - including if it is formatted correctly according to RFC 7208 (which replaced 4408)

If after consideration this is accepted as a to-do :

Recommended results:  
if RR  not exist, return as Warning.
if RR exists, then formatting test determines, if its ok return INFO or a failure return ERROR, considering -
-if policy is !all return ERROR   (useless might as well not have the record)
-if policy is ~all return NOTICE (soft fail, a notice so they are aware it wont block)

What you think ?

marc-vanderwal:
## Purpose

This PR introduces three test cases for testing SPF and DMARC policies for the domain under test.

## Context

Fixes #995.

## Changes

Added three test cases:

- checking SPF for compliance with respect to RFC 7208;
- checking for non-use of the deprecated SPF resource record type;
- checking DMARC for compliance with respect to RFC 7489.

## How to test this PR

This is a documentation update. However #1099 should be merged first.

### DIFF
--- a/docs/specifications/tests/Zone-TP/zone11.md
+++ b/docs/specifications/tests/Zone-TP/zone11.md
@@ -1,0 +1,278 @@
+# ZONE11: SPF policy validation
+
+## Test case identifier
+**Zone11**
+
+## Table of contents
+
+* [Objective](#objective)
+* [Caveat](#caveat)
+* [Scope](#scope)
+* [Inputs](#inputs)
+* [Summary](#summary)
+* [Test procedure](#test-procedure)
+* [Outcome(s)](#outcomes)
+* [Special procedural requirements](#special-procedural-requirements)
+* [Intercase dependencies](#intercase-dependencies)
+* [Terminology](#terminology)
+
+## Objective
+
+Sender Policy Framework (SPF), described in [RFC 7208][RFC7208], is a mechanism
+allowing domain name owners to specify which hosts are allowed to send mail
+claiming to be from that domain. It is implemented by means of TXT records in a
+structured format.
+
+This test case looks up SPF records in the domain to be tested. It checks that
+there is at most one published SPF version 1 policy and, if present, also checks
+the following:
+
+ * the policy has correct syntax;
+ * the policy uses no deprecated [terms][SPF term] nor deprecated [macros][macro].
+
+## Caveat
+
+In order to avoid any denial-of-service attacks, SPF imposes limits on the
+number of recursive DNS lookups an SPF validator may carry out when evaluating a
+policy. This test case will generate an error if it can be certain that a policy
+would generate more than 10 recursive DNS queries by a validator. A policy may
+also use `include` mechanisms or `redirect` modifiers for indirection, but in
+order to keep the test case simple, Zonemaster will not follow these terms.
+
+For that reason, if the published SPF policy contains `include` mechanisms or
+a `redirect` modifier, Zonemaster not raising an error about a too complex
+policy does not mean that the policy will never exceed DNS processing limits
+specified in [RFC 7208, section 4.6.4][RFC7208#4.6.4]. Due to the macro
+expansion facilities of SPF, such problems may only be restricted to certain
+sender domains.
+
+A true SPF validator would need, as inputs, the IPv4 or IPv6 address of the
+sending mail host along with the domain to test. Consequently, Zonemaster will
+not attempt to evaluate the published SPF policy.
+
+## Scope
+
+It is assumed that *Child Zone* has been tested and reported by
+[Connectivity01]. This test case will just ignore non-responsive name servers
+or name servers not giving a correct DNS response for an authoritative name
+server.
+
+## Inputs
+
+* "Child Zone" - The domain name to be tested.
+
+## Summary
+
+Message Tag                     | Level   | Arguments                | Message ID for message tag
+:-------------------------------|:--------|:-------------------------|:--------------------------------------------
+Z11_INCONSISTENT_SPF_POLICIES   | WARNING |                          | The *Child Zone* publishes different SPF policies on different name servers.
+Z11_NO_SPF_FOUND                | DEBUG   |                          | The *Child Zone* does not publish an SPF policy.
+Z11_SPF1_DUPLICATE_MODIFIER     | ERROR   | ns_ip_list, spf_modifier | The *Child Zone*’s SPF version 1 policy contains more than one {spf_modifier} modifier. Policy retrieved from the following nameservers: {ns_ip_list}.
+Z11_SPF1_HAS_P_MACRO            | WARNING | ns_ip_list               | The *Child Zone*’s SPF version 1 policy contains a %{p} macro, which should no longer be used. Policy retrieved from the following nameservers: {ns_ip_list}.
+Z11_SPF1_HAS_PTR                | WARNING | ns_ip_list               | The *Child Zone*’s SPF version 1 policy contains a ptr mechanism, which should no longer be used. Policy retrieved from the following nameservers: {ns_ip_list}.
+Z11_SPF1_MULTIPLE_RECORDS       | ERROR   | ns_ip_list               | The *Child Zone* publishes more than one SPF version 1 policy. Policies retrieved from the following nameservers: {ns_ip_list}.
+Z11_SPF1_RECURSIVE              | NOTICE  | ns_ip_list               | The *Child Zone*’s SPF version 1 policy uses "include" or "redirect". Policy retrieved from the following nameservers: {ns_ip_list}.
+Z11_SPF1_REDIRECT_AND_ALL       | WARNING | ns_ip_list               | The *Child Zone*’s SPF version 1 policy contains both the "all" keyword and a redirect modifier, which will cause "all" to be ignored. Policy retrieved from the following nameservers: {ns_ip_list}.
+Z11_SPF1_REDIRECT_NOT_AT_END    | WARNING | ns_ip_list               | The *Child Zone*’s SPF version 1 policy contains a redirect modifier, but it does not appear at the end of the string. Policy retrieved from the following nameservers: {ns_ip_list}.
+Z11_SPF1_SYNTAX_ERROR           | ERROR   | ns_ip_list               | The *Child Zone*’s SPF version 1 policy has a syntax error. Policy retrieved from the following nameservers: {ns_ip_list}.
+Z11_SPF1_SYNTAX_OK              | INFO    |                          | The *Child Zone*’s SPF version 1 policy has correct syntax.
+Z11_SPF1_TOO_COMPLEX            | ERROR   | ns_ip_list, count        | Evaluating the *Child Zone*’s SPF version 1 policy would require {count} recursive DNS lookups by an SPF validator, exceeding the limit of 10. Policy retrieved from the following nameservers: {ns_ip_list}.
+Z11_UNABLE_TO_CHECK_FOR_SPF     | ERROR   |                          | Unable to check whether *Child Zone* publishes an SPF policy.
+
+The value in the Level column is the default severity level of the message. The
+severity level can be changed in the [Zonemaster-Engine profile]. Also see the
+[Severity Level Definitions] document.
+
+The argument names in the Arguments column lists the arguments used in the
+message. The argument names are defined in the [argument list].
+
+## Test procedure
+
+In this section and unless otherwise specified below, the term "[DNS Query]"
+follows the specification for DNS queries as specified in [DNS Query and
+Response Defaults]. The handling of the DNS responses on the DNS queries follow,
+unless otherwise specified below, what is specified for [DNS Response] in the
+same specification.
+
+1. Create a [DNS Query] with query type TXT and query name *Child Zone* ("TXT
+   query").
+
+2. Create an empty set of pairs of IP addresses and strings, "SPF-Policies".
+
+3. Obtain the set of name server IP addresses using [Method4] and [Method5]
+   ("Name Server IP").
+
+4. For each name server in *Name Server IP* do:
+
+   1. Send *TXT Query* to the name server and collect the [DNS Response].
+
+   2. Go to the next name server IP address if at least one of the following
+      criteria is met:
+
+      1. There is no DNS response.
+      2. [RCODE Name] of the response is not "NoError".
+      3. The AA flag is not set in the response.
+
+   3. If the name server responds with no TXT record, then add the pair
+      consisting of the *Name Server IP* and the empty string to the
+      *SPF-Policies* set.
+
+   4. If the name server responds with at least one TXT record and none is an
+      [SPF TXT record], then add the pair consisting of the *Name Server IP*
+      and the empty string to the *SPF-Policies* set.
+
+   5. If the name server responds with at least one TXT record that is an [SPF
+      TXT record], then, for each [SPF TXT record], add the pair consisting of
+      the *Name Server IP* and the lowercased [concatenation][concatenate] of
+      all strings in the TXT record’s data to the *SPF-Policies* set.
+
+   6. Go to the next name server.
+
+5. If the *SPF-Policies* set is empty, then output
+   *[Z11_UNABLE_TO_CHECK_FOR_SPF]* and terminate the test.
+
+6. If all the pairs in the *SPF-Policies* set contain empty strings, then
+   output *[Z11_NO_SPF_FOUND]* and terminate the test.
+
+7. Compare the set of *SPF-Policies* retrieved from all name servers. If at
+   least two different name servers have returned different sets of SPF
+   policies, then output *[Z11_INCONSISTENT_SPF_POLICIES]*.
+
+8. If the *SPF-Policies* set contains at least two pairs with the same IP
+   address, then output *[Z11_SPF1_MULTIPLE_RECORDS]* with the list of
+   nameservers that returned more than one SPF policy and terminate the test.
+
+9. The following steps assume that all pairs in the *SPF-Policies* set have
+   the same string ("SPF policy").
+
+10. If the *SPF Policy* does not [pass the syntax check][passing the syntax
+    check] for SPF version 1 records, then output *[Z11_SPF1_SYNTAX_ERROR]* and
+    terminate the test.
+
+11. If the *SPF Policy* contains more than one `redirect` modifier, or more than
+    one `exp` modifier, then output *[Z11_SPF1_DUPLICATE_MODIFIER]*.
+
+12. If the *SPF Policy* contains exactly one `redirect` modifier anywhere except
+    at the end of the policy text, then output *[Z11_SPF1_REDIRECT_NOT_AT_END]*.
+
+13. If the *SPF Policy* contains both a `redirect` modifier and an `all`
+    mechanism, then output *[Z11_SPF1_REDIRECT_AND_ALL]*.
+
+14. If the *SPF Policy* contains at least one `ptr` mechanism, then output
+    *[Z11_SPF1_HAS_PTR]*.
+
+15. Count the total number of terms in the *SPF Policy* that is either a
+    mechanism of type `include`, `a`, `mx`, `ptr`, `exists` or a modifier of
+    type `redirect` ("DNS Lookup count").
+
+16. If the `%{p}` [macro] appears as a domain-spec for an [SPF term] in the
+    *SPF Policy*, then:
+
+    1. Output *[Z11_SPF1_HAS_P_MACRO]*.
+
+    2. If the same policy contains no `ptr` mechanism, then increment the *DNS
+    Lookup count*.
+
+17. If the *DNS Lookup count* is greater than 10, then output
+    *[Z11_SPF1_TOO_COMPLEX]*.
+
+18. If the *DNS Lookup count* is less than or equal to 10 and the *SPF Policy*
+    contains a `redirect` modifier or at least one `include` mechanism, then
+    output *[Z11_SPF1_RECURSIVE]*.
+
+19. If no other message was outputted by this test case, then output
+    *[Z11_SPF1_SYNTAX_OK]*.
+
+## Outcome(s)
+
+The outcome of this Test Case is "fail" if there is at least one message
+with the severity level *[ERROR]* or *[CRITICAL]*.
+
+The outcome of this Test Case is "warning" if there is at least one message
+with the severity level *[WARNING]*, but no message with severity level
+*ERROR* or *CRITICAL*.
+
+In other cases, no message or only messages with severity level
+*[INFO]* or *[NOTICE]*, the outcome of this Test Case is "pass".
+
+## Special procedural requirements
+
+If either IPv4 or IPv6 transport is disabled, skip sending queries over that
+transport protocol. A message will be outputted reporting that the transport
+protocol has been skipped.
+
+## Intercase dependencies
+
+None.
+
+## Terminology
+
+* "SPF TXT record" - The term is used to refer to a TXT resource record which,
+  after [concatenating][concatenate] all strings within that resource record
+  into one string, yields a string either equal to `v=spf1` or starting with
+  `v=spf1` followed by a space, irrespective of character case.
+
+* "concatenate" - The term is used to refer to the conversion of a TXT
+  resource record’s data to a single contiguous string, as specified in [RFC
+  7208, section 3.3][RFC7208#3.3].
+
+* "macro" - The term is used to refer to the macros permissible in SPF
+  domain-specs, as specified in [RFC 7208, section 7][RFC7208#7].
+
+* "passing the syntax check" - The term is used in this document to refer to
+  text that is valid according to the ABNF grammar published in [RFC
+  7208][RFC7208] starting from [section 4.5][RFC7208#4.5].
+
+* "using Method" - The term is used when data is fetched using the defined
+  [Method][Methods].
+
+* "SPF term" - The term is used to refer to either an SPF mechanism as defined
+  in [RFC 7208, section 5][RFC7208#5] or an SPF modifier as defined in [RFC
+  7208, section 6][RFC7208#6].
+
+[Argument list]:                        https://github.com/zonemaster/zonemaster-engine/blob/master/docs/logentry_args.md
+[argument]:                             #terminology
+[concatenate]:                          #terminology
+[Connectivity01]:                       ../Connectivity-TP/connectivity01.md
+[CRITICAL]:                             ../SeverityLevelDefinitions.md#critical
+[DNS Query and Response Defaults]:      ../DNSQueryAndResponseDefaults.md
+[DNS Query]:                            ../DNSQueryAndResponseDefaults.md#default-setting-in-dns-query
+[DNS Response]:                         ../DNSQueryAndResponseDefaults.md#default-handling-of-a-dns-response
+[ERROR]:                                ../SeverityLevelDefinitions.md#error
+[INFO]:                                 ../SeverityLevelDefinitions.md#info
+[macro]:                                #terminology
+[Message Tag Specification]:            MessageTagSpecification.md
+[Method4]:                              ../Methods.md#method-4-obtain-glue-address-records-from-parent
+[Method5]:                              ../Methods.md#method-5-obtain-the-name-server-address-records-from-child
+[Methods]:                              ../Methods.md
+[NOTICE]:                               ../SeverityLevelDefinitions.md#notice
+[passing the syntax check]:             #terminology
+[RCODE Name]:                           https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-6
+[RFC7208#3.3]:                          https://www.rfc-editor.org/rfc/rfc7208#section-3.3
+[RFC7208#4.5]:                          https://www.rfc-editor.org/rfc/rfc7208#section-4.5
+[RFC7208#4.6.4]:                        https://www.rfc-editor.org/rfc/rfc7208#section-4.6.4
+[RFC7208#4]:                            https://www.rfc-editor.org/rfc/rfc7208#section-4
+[RFC7208#5]:                            https://www.rfc-editor.org/rfc/rfc7208#section-5
+[RFC7208#6]:                            https://www.rfc-editor.org/rfc/rfc7208#section-6
+[RFC7208#7]:                            https://www.rfc-editor.org/rfc/rfc7208#section-7
+[RFC7208]:                              https://www.rfc-editor.org/rfc/rfc7208
+[Severity Level Definitions]:           ../SeverityLevelDefinitions.md
+[SPF term]:                             #terminology
+[SPF TXT record]:                       #terminology
+[Test Case Identifier Specification]:   TestCaseIdentifierSpecification.md
+[Undelegated test]:                     ../../test-types/undelegated-test.md
+[WARNING]:                              ../SeverityLevelDefinitions.md#warning
+[Z11_INCONSISTENT_SPF_POLICIES]:        #summary
+[Z11_NO_SPF_FOUND]:                     #summary
+[Z11_SPF1_DUPLICATE_MODIFIER]:          #summary
+[Z11_SPF1_HAS_P_MACRO]:                 #summary
+[Z11_SPF1_HAS_PTR]:                     #summary
+[Z11_SPF1_MULTIPLE_RECORDS]:            #summary
+[Z11_SPF1_RECURSIVE]:                   #summary
+[Z11_SPF1_REDIRECT_AND_ALL]:            #summary
+[Z11_SPF1_REDIRECT_NOT_AT_END]:         #summary
+[Z11_SPF1_SYNTAX_ERROR]:                #summary
+[Z11_SPF1_SYNTAX_OK]:                   #summary
+[Z11_SPF1_TOO_COMPLEX]:                 #summary
+[Z11_UNABLE_TO_CHECK_FOR_SPF]:          #summary
+[Zonemaster-Engine profile]:            https://github.com/zonemaster/zonemaster-engine/blob/master/docs/Profiles.md
--- a/docs/specifications/tests/Zone-TP/zone12.md
+++ b/docs/specifications/tests/Zone-TP/zone12.md
@@ -1,0 +1,156 @@
+# ZONE12: No use of the deprecated SPF resource record type
+
+## Test case identifier
+**Zone12**
+
+## Table of contents
+
+* [Objective](#objective)
+* [Scope](#scope)
+* [Inputs](#inputs)
+* [Summary](#summary)
+* [Test procedure](#test-procedure)
+* [Outcome(s)](#outcomes)
+* [Special procedural requirements](#special-procedural-requirements)
+* [Intercase dependencies](#intercase-dependencies)
+* [Terminology](#terminology)
+
+## Objective
+
+Sender Policy Framework (SPF) was first specified by [RFC 4408][RFC4408],
+which was then obsoleted by [RFC 7208][RFC7208].
+
+[RFC 4408, section 3.1.1][RFC4408#3.1.1] proposed a dedicated resource record
+type named “SPF”, to be used for publishing SPF policies. The same RFC
+recommended to publish the same SPF policy twice, with one copy using the SPF
+resource record type and the other using the more generic TXT type.
+
+However, [RFC 7208, section 3.1][RFC7208#3.1] deprecates this dual publication
+practice, stating that SPF policies must be published in TXT resource records
+only. SPF resource records should therefore no longer be used.
+
+This test case checks for the presence of any SPF resource records at the apex
+of the domain being tested and emits a warning if any such resource record is
+found. No further validation of any kind is performed on the data contained in
+any SPF resource records that are found.
+
+## Scope
+
+It is assumed that *Child Zone* has been tested and reported by
+[Connectivity01]. This test case will just ignore non-responsive name servers or
+name servers not giving a correct DNS response for an authoritative name server.
+
+## Inputs
+
+* "Child Zone" - The domain name to be tested.
+
+## Summary
+
+Message Tag                     | Level   | Arguments            | Message ID for message tag
+:-------------------------------|:--------|:---------------------|:--------------------------------------------
+Z12_SPF_RECORD_FOUND            | WARNING | ns_ip_list           | The *Child Zone* publishes an SPF policy with SPF resource records, which are deprecated in favor of TXT. Policy retrieved from the following nameservers: {ns_ip_list}.
+Z12_NO_SPF_RECORDS              | INFO    |                      | The *Child Zone* does not publish any SPF policies using the deprecated SPF resource record type.
+Z12_UNABLE_TO_CHECK             | ERROR   |                      | Unable to check whether *Child Zone* uses resource records of type SPF.
+
+The value in the Level column is the default severity level of the message. The
+severity level can be changed in the [Zonemaster-Engine profile]. Also see the
+[Severity Level Definitions] document.
+
+The argument names in the Arguments column lists the arguments used in the
+message. The argument names are defined in the [argument list].
+
+## Test procedure
+
+In this section and unless otherwise specified below, the term "[DNS Query]"
+follows the specification for DNS queries as specified in [DNS Query and
+Response Defaults]. The handling of the DNS responses on the DNS queries follow,
+unless otherwise specified below, what is specified for [DNS Response] in the
+same specification.
+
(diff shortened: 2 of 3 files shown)


matsduf (changes requested):
The proposals do not follow the specification [template]. Could you rewrite them? You could also check the active PRs for specification updates, e.g. #1098 and #1032.

[template]: https://github.com/zonemaster/zonemaster/blob/develop/docs/internal-documentation/templates/specifications/tests/Template01.md

mvw-afnic:
Sure, I will.

I was also unsure about where to add these new test cases. I’ve added them in `Zone` because that’s where the tests involving MX records are, but they could be in a new category (maybe `Email`).

matsduf:
> I was also unsure about where to add these new test cases. I’ve added them in `Zone` because that’s where the tests involving MX records are, but they could be in a new category (maybe `Email`).

I think `Zone` is fine. I am not sure that adding a new test level would make it better. Before adding a new level I think we should have a broader discussion about which levels to have and what they should contain.

mvw-afnic:
Okay, sticking with `Zone`, then.

mvw-afnic:
> The proposals do not follow the specification [template](https://github.com/zonemaster/zonemaster/blob/develop/docs/internal-documentation/templates/specifications/tests/Template01.md). Could you rewrite them? You could also check the active PRs for specification updates, e.g. #1098 and #1032.

Done.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,153 @@
...
+
+4. If at least one TXT records starts with the string `v=spf`, but none
+   specify an SPF version 1 policy, then output *[SPF_WRONG_VERSION]*.
+
+5. Select all TXT records either equal to the string `v=spf1` or starting with
+   `v=spf1` followed by a space. If there are two or more such records, then
+   output *[SPF1_MULTIPLE_RECORDS]* and terminate the test. The next steps
+   assume there is exactly one such TXT record.
+
+6. If the text does not pass the syntax check for SPF version 1 records,
```
matsduf, line 80: Please make "pass the syntax check" a link to the terminology section. See how "Null MX" is handled in https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/Zone-TP/zone09.md?plain=1

matsduf (changes requested):

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,153 @@
...
+
+This test looks up SPF records in the domain to be tested. It ensures that
+there is at most one such record and, if present, ensures it is syntactically
+valid and that its evaluation does not require more than 10 DNS lookups.
+
+## Scope
+
+It is assumed that *Child Zone* has been tested and reported by [Basic04]. This
+test case will just ignore non-responsive name servers or name servers not
+giving a correct DNS response for an authoritative name server.
```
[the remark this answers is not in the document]
mvw-afnic: Fixed. But this means that #1099 should be merged first before merging this PR.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,153 @@
...
+SPF1_MORE_THAN_ONE_SAME_MODIFIER | ERROR  | modifier             | The published SPF version 1 policy contains the same modifier (redirect or exp) more than once.
+SPF1_MULTIPLE_RECORDS           | ERROR   |                      | The domain publishes more than one SPF version 1 policy.
+SPF1_REDIRECT_AND_ALL           | WARNING |                      | The published SPF version 1 policy contains both the "all" keyword and a redirect modifier.
+SPF1_REDIRECT_NOT_AT_END        | WARNING |                      | The published SPF version 1 policy contains a redirect modifier, but it does not appear at the end of the string.
+SPF1_SYNTAX_ERROR               | ERROR   |                      | The published SPF version 1 policy has a syntax error.
+SPF_NOT_CONFIGURED              | INFO    |                      | No SPF policy was found.
+SPF_UNEXPECTED_ERROR            | ERROR   | rcode                | The server responded with an unexpected RCODE when looking up SPF records.
+SPF_WRONG_VERSION               | WARNING | version              | The domain publishes an SPF policy, but its version number is invalid.
```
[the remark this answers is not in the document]
mvw-afnic: Fixed. Changing to `spf_version` in order to avoid any confusion.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,153 @@
...
+The argument names in the Arguments column lists the arguments used in the
+message. The argument names are defined in the [argument list].
+
+## Test procedure
+
+1. Create a TXT query for the apex of the *Child Zone* and send it over UDP to
+   the name server.
+
+2. If the name server responds with an RCODE different from NOERROR, then
+   output *[SPF_UNEXPECTED_ERROR]* and terminate the test.
```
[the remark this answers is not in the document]
mvw-afnic: Okay, so this makes a hypothetical `Z11_SPF_UNEXPECTED_ERROR` (and variants of these in the DMARC test) redundant.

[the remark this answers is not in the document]
matsduf: No, not necessarily. In [Nameserver02] the name server is tested how it handles EDNS queries. First an SOA query without EDNS is sent, and if the response is normal, a query with EDNS is sent and the behavior if this is anlyzed. Here a SOA query could be sent and you would expect the same RCODE if the owner name is the same.

[Nameserver02]: https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/Nameserver-TP/nameserver02.md

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,153 @@
...
+
+* "Child Zone" - The domain name to be tested.
+
+## Summary
+
+Message Tag outputted           | Level   | Arguments            | Description of when message tag is outputted
+:-------------------------------|:--------|:---------------------|:--------------------------------------------
+SPF1_FOUND_AND_VALID            | INFO    |                      | The domain publishes a valid SPF version 1 policy.
+SPF1_LOOKUP_LIMIT_EXCEEDED      | ERROR   |                      | Evaluating the published SPF version 1 policy would require more than 10 DNS lookups.
+SPF1_MORE_THAN_ONE_SAME_MODIFIER | ERROR  | modifier             | The published SPF version 1 policy contains the same modifier (redirect or exp) more than once.
```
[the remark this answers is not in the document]
matsduf: I think one PR is enough unless there are multiple updates of the argument list, then I guess an issue would be fine so that we do not forget it.

I guess that the second question we should ask if a new argument name is needed, or if a generic name could be used. If a specific one should be used, should the argument name be `spf_modifier` or `spf_mod` (to keep it shorter).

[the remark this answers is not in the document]
mvw-afnic: Oh sorry, I haven’t pushed a new version of the PR yet. I was also going to change the argument to `spf_modifier`.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,199 @@
...
+follows the specification for DNS queries as specified in [DNS Query and
+Response Defaults]. The handling of the DNS responses on the DNS queries follow,
+unless otherwise specified below, what is specified for [DNS Response] in the
+same specification.
+
+
+1. Create a [DNS Query] with query type TXT query and query name *Child Zone*
+   ("TXT query").
+
+2. Create three empty sets, "TXT-Data", "SPF-Policies" and "SPFv1-Policies".
```
matsduf, line 72: I suggest that the type of data to be stored in each set is specified.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,199 @@
...
+
+3. Obtain the set of name server IP addresses using [Method4] and [Method5]
+   ("Name Server IP").
+
+4. For each name server in *Name Server IP* do:
+
+   1. Send *TXT Query* to the name server and collect the [DNS Responses][DNS
+      Response].
```
matsduf, line 80: If you send one query, there will only be one response, right?

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,199 @@
...
+      criteria is met:
+
+      1. There is no DNS response.
+      2. [RCODE Name] of the response is not "NoError".
+      3. The AA flag is not set in the response.
+      4. There is no TXT resource record in the response.
+
+   3. If the name server responds with at least one TXT record, then, for each
+      TXT record, concatenate all strings within the TXT record’s data together
+      and add the resulting string to the *TXT-Data* set.
```
matsduf, line 92: Concatenate sounds like you take the RDATA from all TXT records and turn them into one string. Don't you keep the TXT records apart? Wouldn't it be simpler to already here sort out SPF related TXT records and ignore the rest?

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,199 @@
...
+
+   4. Go to the next name server.
+
+5. For each string in the *TXT-Data* set, if the string starts with `v=spf`,
+   then add it to the *SPF-Policies* set.
+
+6. If the *SPF-Policies* set is empty, then output *[Z11_SPF_NOT_CONFIGURED]*
+   and terminate the test.
```
matsduf, line 100: Do you assume that all nameservers give the same response?

mvw-afnic: How about introducing an extra test case in the “Consistency” group which ensures all nameservers publish the same SPF policies?

docs/specifications/tests/Zone-TP/zone11.md
mvw-afnic: I can try to adjust the wording slightly to remove the ambiguity, and add “concatenate” to the Terminology section, linking to [RFC 7208 § 3.3](https://www.rfc-editor.org/rfc/rfc7208#section-3.3). This passage states:

> As defined in [[RFC1035](https://www.rfc-editor.org/rfc/rfc1035)], Sections [3.3](https://www.rfc-editor.org/rfc/rfc7208#section-3.3) and [3.3.14](https://www.rfc-editor.org/rfc/rfc7208#section-3.3.14), a single text DNS record can be composed of more than one string.  If a published record contains multiple character-strings, then the record MUST be treated as if those strings are concatenated together without adding spaces.

docs/specifications/tests/Zone-TP/zone11.md
mvw-afnic: Yes, my bad.

mvw-afnic:
Overhauled Zone11. Rewrites of Zone12 and Zone13 will come later.

docs/specifications/tests/Zone-TP/zone11.md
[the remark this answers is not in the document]
mvw-afnic: You are right: suppose all nameservers of a zone return SERVFAIL when we ask them for TXT records. Then the test would wrongfully emit a “no SPF configured” message. That would be misleading, because in fact we were unable to check for an SPF policy. That’s an issue I still need to take care of.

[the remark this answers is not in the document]
matsduf: I would say that it is unlikely that a server would that the server would correctly return an SOA record but not a TXT record.

docs/specifications/tests/Zone-TP/zone11.md
matsduf: That would be OK. We could also consider saving test case and coding by including it in this test case.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,262 @@
...
+When the term "passing the syntax check" is used, it means that the text is
+valid according to the ABNF grammar published in [RFC 7208][RFC7208] starting
+from [§ 4.5][RFC7208#4.5].
+
+When the term "using Method" is used, names and IP addresses are fetched
+using the defined [Methods].
+
+When the term "SPF term" is used, it includes both SPF mechanisms as defined in
+[RFC 7208, § 5][RFC7208#5] and modifiers as defined in [RFC 7208, §
+6][RFC7208#6].
```
matsduf, line 220: For this test case specification there are five terms. This is the format used in other specifications, but I wonder if it would be more readable if the term is topicalized and using a format like the following:

* "concatenate" - When the term is used, it refers to the conversion of a TXT
resource record’s data to a single contiguous string, as specified in [RFC 7208,
§ 3.3][RFC7208#3.3].
* "macro" - When the term is used, it refers to the macros permissible in SPF
domain-specs, as specified in [RFC 7208, § 7][RFC7208#7].

matsduf (changes requested):
I have only reviewed Zone11.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,262 @@
...
+
+      1. [Concatenate][concatenate] all strings within the TXT record’s data
+         together.
+      2. If the resulting string starts with `v=spf`, then add it to the
+         *SPF-Policies* set.
+
+   4. Go to the next name server.
+
+5. If the *SPF-Policies* set is empty, then output *[Z11_SPF_NOT_CONFIGURED]*
+   and terminate the test.
```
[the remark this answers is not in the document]
mvw-afnic: I disagree. If a domain name owner does not plan on using their domain name for receiving or sending email, it would make perfect sense for them to specify `v=spf1 -all`.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,262 @@
...
+   starts with `v=spf1` followed by a space, then add it to the *SPFv1-Policies*
+   set.
+
+7. If the *SPFv1-Policies* set is empty, then for each string in the
+   *SPF-Policies* set, output *[Z11_SPF_WRONG_VERSION]* with the version number
+   extracted from the string.
+
+8. If the *SPFv1-Policies* set contains more than one element, then output
+   *[Z11_SPF1_MULTIPLE_RECORDS]* and terminate the test. The next steps assume
+   there is exactly one string in the *SPFv1-Policies* set ("SPF policy").
```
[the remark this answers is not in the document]
mvw-afnic: Yes, this is case 1. The next version I’m pushing will include an explicit check for case 2. I’ll add `ns_ip_list` to the arguments.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,262 @@
...
+
+9. If the *SPF Policy* does not [pass the syntax check][passing the syntax
+   check] for SPF version 1 records, then output *[Z11_SPF1_SYNTAX_ERROR]* and
+   terminate the test.
```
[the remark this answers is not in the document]
mvw-afnic: If there are multiple records, the test should have been terminated (see comment at end of step 8). [RFC 7208, § 4.5](https://www.rfc-editor.org/rfc/rfc7208#section-4.5) explicitly states that evaluation must be aborted if there are multiple records.

SPF policies can sometimes exceed 255 characters, that’s true, but I don’t think most policies will be that complex. We should be fine with including the full SPF text in this message.

docs/specifications/tests/Zone-TP/zone11.md
mvw-afnic: Works for me. In that case we should consider updating the template and the specifications that already follow the new template.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,262 @@
...
+   2. If the same policy contains no `ptr` mechanism, then increment the *DNS
+   Lookup count*.
+
+16. If the *DNS Lookup count* is greater than 10, then output
+    *[Z11_SPF1_TOO_COMPLEX]*.
+
+17. If the *DNS Lookup count* is less than or equal to 10 and the *SPF Policy*
+    contains a `redirect` modifier or at least one `include` mechanism, then
+    output *[Z11_SPF1_RECURSIVE]*.
+
```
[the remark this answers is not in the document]
mvw-afnic: I’ll add `spf_policy` as an argument to all of these messages.

docs/specifications/tests/Zone-TP/zone11.md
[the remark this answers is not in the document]
matsduf: There is a difference if multiple SPF records are returned from the same name server or if different name servers return diffrent SPF records, but just one each. Your specification needs to be updated to handle the situation when the name servers return different SPF records. Compare how it is done in [Zone09].

If the policy is included, that could be truncated down to, say, 80 characters to keep the length of the message down.

[Zone09]: https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/Zone-TP/zone09.md

docs/specifications/tests/Zone-TP/zone11.md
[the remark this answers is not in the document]
matsduf: A second question, do you really think it is relevant to output the message for root zone, TLD zones, reverse zones and zones with NULL MX records? See [Zone09].

[Zone09]: https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/Zone-TP/zone09.md

docs/specifications/tests/Zone-TP/zone11.md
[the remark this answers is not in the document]
matsduf: You have renumbered. My comment was about returning the name server where the SPF record with problem was found.

docs/specifications/tests/Zone-TP/zone11.md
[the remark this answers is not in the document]
mvw-afnic: I think it’s useful to point it out either way:

- if a domain owner intends their domain to be used for outgoing e-mail, SPF will protect them and potentially improve their deliverability, so they have something to gain by publishing an SPF policy;
- conversely, if a domain owner does not intend to use their domain name for sending e-mail, they can limit the potential for their domain to be abused by spammers if they publish an SPF policy stating that nobody shall send email from that domain.

In the second case, I’ve read in several places the advice to publish `-all` as an SPF policy, and `p=reject sp=reject` as a DMARC policy, on domains that should not emit mail, as a hardening practice. In both cases there’s something to be gained for a domain owner by implementing SPF.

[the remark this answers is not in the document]
mvw-afnic: > A second question, do you really think it is relevant to output the message for root zone, TLD zones, reverse zones and zones with NULL MX records? See [Zone09](https://github.com/zonemaster/zonemaster/blob/develop/docs/specifications/tests/Zone-TP/zone09.md).

You are right, we shouldn’t run the test on the root zone, TLD zones (but do we handle edge cases of second-level domains that have the same semantics as TLD zones, such as `co.uk`?).

As for DMARC, its test case is skipped if what [RFC 7489](https://www.rfc-editor.org/rfc/rfc7489) calls the "organizational domain" cannot be deduced from the domain under test.

And we should also exclude from the test case any domain below some of the following special-use domains:

- .arpa;
- .test;
- .localhost;
- .example;
- .invalid.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,262 @@
...
+      3. The AA flag is not set in the response.
+      4. There is no TXT resource record in the response.
+
+   3. If the name server responds with at least one TXT record, then, for each
+      TXT record:
+
+      1. [Concatenate][concatenate] all strings within the TXT record’s data
+         together.
+      2. If the resulting string starts with `v=spf`, then add it to the
+         *SPF-Policies* set.
```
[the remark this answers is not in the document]
mvw-afnic: I’ve updated the spec, but I didn’t feel comfortable leaving out the definition of what an SPF TXT record is. I’ve therefore added an item in the Terminology section.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,153 @@
...
+SPF1_FOUND_AND_VALID            | INFO    |                      | The domain publishes a valid SPF version 1 policy.
+SPF1_LOOKUP_LIMIT_EXCEEDED      | ERROR   |                      | Evaluating the published SPF version 1 policy would require more than 10 DNS lookups.
+SPF1_MORE_THAN_ONE_SAME_MODIFIER | ERROR  | modifier             | The published SPF version 1 policy contains the same modifier (redirect or exp) more than once.
+SPF1_MULTIPLE_RECORDS           | ERROR   |                      | The domain publishes more than one SPF version 1 policy.
+SPF1_REDIRECT_AND_ALL           | WARNING |                      | The published SPF version 1 policy contains both the "all" keyword and a redirect modifier.
+SPF1_REDIRECT_NOT_AT_END        | WARNING |                      | The published SPF version 1 policy contains a redirect modifier, but it does not appear at the end of the string.
+SPF1_SYNTAX_ERROR               | ERROR   |                      | The published SPF version 1 policy has a syntax error.
+SPF_NOT_CONFIGURED              | INFO    |                      | No SPF policy was found.
```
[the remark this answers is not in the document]
mvw-afnic: I eventually renamed the message tag to `Z11_NO_SPF_FOUND`. As for the choice of level or choosing whether or not to output this message if a null MX is encountered, we’re having the same conversation below, so I’ll mark this bit as resolved.

docs/specifications/tests/Zone-TP/zone11.md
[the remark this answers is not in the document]
mvw-afnic: I am not the only one to disagree. There is at least one other testing tool that checks for appropriate SPF and DMARC policies for domains that have null MX records. Which makes sense, because if you clearly state a domain is not to be used to receive email, usually you don’t intend to send e-mail with it either and the right SPF and DMARC policies can help you state that fact.

Here’s an example: https://internet.nl/mail/chargemove.fr/799026/

docs/specifications/tests/Zone-TP/zone11.md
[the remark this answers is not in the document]
mvw-afnic: Okay, that makes sense too. I think it would be useful to have both the erroneous SPF policy and the nameserver that returns it.

As for the length of the SPF policies, let’s just start with including them in their entirety in the messages. We can evaluate later whether it turns out to be impractical. I’ll need to research the average length of an SPF policy.

docs/specifications/tests/Zone-TP/zone11.md
mvw-afnic: This PR is merged, so I’ll rebase the next iteration on the new `develop` branch and rewrite the test cases so that they follow the updated template.

docs/specifications/tests/Zone-TP/zone11.md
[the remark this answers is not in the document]
mvw-afnic: I’ll add a step to check just that in the next iteration.

As for the length of the SPF policies, let’s just start with including them in their entirety in the messages. We can evaluate later whether it turns out to be impractical. I’ll need to research the average length of an SPF policy but I think they’ll rarely exceed 80 octets.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,284 @@
...
+      TXT record:
+
+      1. If the record is an [SPF TXT record], then add the lowercased
+         [concatenation][concatenate] of all strings in the TXT record’s data to
+         the *SPF-Policies* set.
+
+   4. Go to the next name server.
+
+5. Compare the set of *SPF-Policies* retrieved from all name servers. If at
+   least one name server has responded with a set of SPF policies and a second
```
matsduf, line 126: I think "with a set" --> "with one set" is more idiomatic.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,284 @@
...
+   name server has responded with a different set of SPF policies, then output
+   *[Z11_INCONSISTENT_SPF_POLICIES]*.
+
+6. If no name server in the *Name Server IP* set responded with an authoritative
+   "NoError" response, then output *[Z11_UNABLE_TO_CHECK_SPF]* and terminate the
+   test.
```
matsduf, line 132: You have not collected that data, have you? *SPF-Policies* will be empty then, but that will also be the case if all servers responds authoritatively with not TXT record.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,284 @@
...
+
+7. If the *SPF-Policies* set is empty, then output *[Z11_NO_SPF_FOUND]*
+   and terminate the test.
+
+8. For each string in *SPF-Policies* set, if the string is equal to `v=spf1` or
+   starts with `v=spf1` followed by a space, then add it to the *SPFv1-Policies*
+   set.
```
matsduf, line 139: Plus the name server IP?

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,284 @@
...
+
+9. If the *SPFv1-Policies* set is empty, then for each string in the
+   *SPF-Policies* set, output *[Z11_SPF_WRONG_VERSION]* with the version number
+   extracted from the string and the nameservers that returned the string.
+
+10. If the *SPFv1-Policies* set contains more than one element, then output
+    *[Z11_SPF1_MULTIPLE_RECORDS]* with the list of nameservers that returned
+    more than one SPF policy and terminate the test. The next steps assume
+    there is exactly one string in the *SPFv1-Policies* set ("SPF policy").
```
matsduf, line 148: I think it will be clearer if the last sentence is move to a bullet of its own in between 10 and 11.

"The next steps" --> "The following steps"?

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,284 @@
...
+
+15. If the *SPF Policy* contains at least one `ptr` mechanism, then output
+    *[Z11_SPF1_HAS_PTR]*.
+
+16. Count the number of terms in the *SPF Policy* that are either ("DNS Lookup
+    count"):
+
+    1. A mechanism of type `a`, `mx`, `ptr`, `exists`;
+
+    2. Or a modifier of type `redirect`.
```
matsduf, line 171: Is that the same as:
```
Count the total number of terms in the *SPF Policy* that is either a mechanism of type `a`, `mx`, `ptr`, `exists` or a modifier of type `redirect` ("DNS Lookup count").
```
If so, I think that would be easier to read. Else, please clarify.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,284 @@
...
+## Test procedure
+
+In this section and unless otherwise specified below, the term "[DNS Query]"
+follows the specification for DNS queries as specified in [DNS Query and
+Response Defaults]. The handling of the DNS responses on the DNS queries follow,
+unless otherwise specified below, what is specified for [DNS Response] in the
+same specification.
+
+1. Create a [DNS Query] with query type TXT and query name *Child Zone* ("TXT
+   query").
```
matsduf, line 97: What about SPF records? Are they just ignored?

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,284 @@
...
+
+## Inputs
+
+* "Child Zone" - The domain name to be tested.
+
+## Summary
+
+Message Tag                     | Level   | Arguments                     | Message ID for message tag
+:-------------------------------|:--------|:------------------------------|:--------------------------------------------
+Z11_INCONSISTENT_SPF_POLICIES   | WARNING |                               | The *Child Zone* publishes different SPF policies on two different name servers.
```
matsduf, line 66: Exactly two? "different SPF policies on two different name servers" --> "different SPF policies on different name servers"?

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,284 @@
...
+Z11_SPF1_SYNTAX_ERROR           | ERROR   | ns_ip_list, spf_policy        | The following SPF version 1 policy has a syntax error: {spf_policy}. Policy retrieved from the following nameservers: {ns_ip_list}.
+Z11_SPF1_SYNTAX_OK              | INFO    |                               | The *Child Zone*’s SPF version 1 policy has correct syntax.
+Z11_SPF1_TOO_COMPLEX            | ERROR   | ns_ip_list, spf_policy, count | Evaluating the following SPF version 1 policy would require {count} recursive DNS lookups by an SPF validator, exceeding the limit of 10: {spf_policy}. Policy retrieved from the following nameservers: {ns_ip_list}.
+Z11_SPF_WRONG_VERSION           | WARNING | ns_ip_list, spf_version       | The *Child Zone* publishes an SPF policy with invalid version number {spf_version}. Policy retrieved from the following name servers: {ns_ip_list}.
+Z11_UNABLE_TO_CHECK_SPF         | ERROR   |                               | Unable to check whether *Child Zone* publishes an SPF policy.
```
matsduf, line 79: "Z11_UNABLE_TO_CHECK_SPF" sounds more like "unable to validate SPF". Maybe "Z11_UNABLE_TO_CHECK_FOR_SPF"? On the other hand it seems less likely that a name sever responds authoritatively on a SOA record query but not an TXT record query.

matsduf:
Validation of an SPF policy according to the RFC is quite complex!

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,284 @@
...
+
+* "Child Zone" - The domain name to be tested.
+
+## Summary
+
+Message Tag                     | Level   | Arguments                     | Message ID for message tag
+:-------------------------------|:--------|:------------------------------|:--------------------------------------------
+Z11_INCONSISTENT_SPF_POLICIES   | WARNING |                               | The *Child Zone* publishes different SPF policies on two different name servers.
+Z11_NO_SPF_FOUND                | DEBUG   |                               | The *Child Zone* does not publish an SPF policy.
+Z11_SPF1_HAS_P_MACRO            | WARNING | ns_ip_list, spf_policy        | The following SPF version 1 policy contains a %{p} macro, which should no longer be used: {spf_policy}. Policy retrieved from the following nameservers: {ns_ip_list}.
```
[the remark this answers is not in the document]
marc-vanderwal: Based on what I see on the `.fr` zone, 92 % of SPF rules are 80 octets long or less. The 95th percentile is at 93 octets. However, the outliers can be massive, with the largest being 2 132 octets.

While most SPF rules could be expressed literally, maybe the sheer fact that such outliers exist does sort of weigh against including the whole policy in the message. Maybe just leave it out, after all?

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,284 @@
...
+In this section and unless otherwise specified below, the term "[DNS Query]"
+follows the specification for DNS queries as specified in [DNS Query and
+Response Defaults]. The handling of the DNS responses on the DNS queries follow,
+unless otherwise specified below, what is specified for [DNS Response] in the
+same specification.
+
+1. Create a [DNS Query] with query type TXT and query name *Child Zone* ("TXT
+   query").
+
+2. Create two empty sets of strings, "SPF-Policies" and "SPFv1-Policies".
```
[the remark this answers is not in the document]
marc-vanderwal: Both sets should then be pairs of IP addresses and strings. Fixing.

docs/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: I’m going to rephrase this sentence anyway because it’s awkward. I’m considering:

> If at least two different name servers have returned different sets of SPF policies, then output *[Z11_INCONSISTENT_SPF_POLICIES]*.

docs/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: I’ll create a set named “Name Servers returning NOERROR” and will test for emptiness of the set.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,284 @@
...
+      2. [RCODE Name] of the response is not "NoError".
+      3. The AA flag is not set in the response.
+      4. There is no TXT resource record in the response.
+
+   3. If the name server responds with at least one TXT record, then, for each
+      TXT record:
+
+      1. If the record is an [SPF TXT record], then add the lowercased
+         [concatenation][concatenate] of all strings in the TXT record’s data to
+         the *SPF-Policies* set.
```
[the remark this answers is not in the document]
marc-vanderwal: Yes. Fixing.

docs/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: Yes indeed. Fixing.

docs/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: Fixing.

docs/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: The SPF resource record type (type 99) is deprecated and should no longer be used to publish SPF policies. The Zone12 test case looks for any data using this specific type and will emit a warning.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,284 @@
...
+## Terminology
+
+* "SPF TXT record" - The term is used to refer to a TXT resource record which,
+  after [concatenating][concatenate] all strings within that resource record
+  into one string, yields a string starting with `v=spf`, irrespective of
+  character case.
+
+* "concatenate" - The term is used to refer to the conversion of a TXT
+  resource record’s data to a single contiguous string, as specified in [RFC
+  7208, § 3.3][RFC7208#3.3].
```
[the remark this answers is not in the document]
marc-vanderwal: My bad, that’s an old habit of mine. I use the [BÉPO keyboard layout](https://bepo.fr), a keyboard layout optimized for French (while still working pretty well for code), which makes typing symbols like § a breeze. I’ll replace all occurrences of "§" with "section" in all the new files in this PR.

docs/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: At least two, but yes, I’ll go with your suggestion.

docs/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: Okay, renaming the message tag.

marc-vanderwal:
Yes, validating an SPF policy is indeed a chore (and a tool like Zonemaster cannot necessarily check everything). But I’m doing some research on the use of SPF in the `.fr` zone and as far as I can see, I’m pretty confident that this test case will catch most errors people make when publishing SPF policies.

docs/specifications/tests/Zone-TP/zone11.md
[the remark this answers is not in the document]
matsduf: I suggest to leave it out. Even with a 40 characters policy it will be hard to read the message with first some text and then the policy and then some more text and then a list of IP addresses. I always put the list of IP addresses at the end of the message to make sure it does not prevent the reader from finding the text. (And I have created issue https://github.com/zonemaster/zonemaster-gui/issues/315 for improved readability.)

docs/specifications/tests/Zone-TP/zone11.md
matsduf: Fine.

docs/specifications/tests/Zone-TP/zone11.md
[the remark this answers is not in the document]
marc-vanderwal: Okay, good point. I’ve removed the placeholders for the SPF policy in my latest iteration.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,276 @@
...
+   3. If the name server responds with no TXT record, then add the pair
+      consisting of the *Name Server IP* and the empty string to the
+      *SPF-Policies* set.
+
+   4. If the name server responds with at least one TXT record, then, for each
+      TXT record:
+
+      1. If the record is an [SPF TXT record], then add the pair consisting of
+         the *Name Server IP* and the lowercased [concatenation][concatenate]
+         of all strings in the TXT record’s data to the *SPF-Policies* set.
```
matsduf, line 125: In 4 you should probably add `*Name Server IP* and the empty string to the *SPF-Policies* set` if none of the TXT records were SPF TXT records. No TXT records at all and none of the TXT records being SPF records should be the same thing.

matsduf:
> People very rarely make mistakes in the SPF version number. Less than 0,1% of domains in the `.fr` zone, which our crawler has classified as “not parked”, have TXT records starting with `v=spf` followed by something else than `1`. So such mistakes are not worth testing for.

If I understand it correctly, Zone11 will consider such TXT records as non-SPF records. It is hard to say how much errors to look for. There is always a limit.

matsduf:
This PR is a draft. Are Zone12 and Zone13 ready for review?

marc-vanderwal:
> If I understand it correctly, Zone11 will consider such TXT records as non-SPF records. It is hard to say how much errors to look for. There is always a limit.

That’s right, Zone11 will output Z11_NO_SPF_FOUND.

However it is interesting, I think, to look for TXT records that do contain `v=spf1` somewhere in their text. I’ve seen instances of mistakes such as:

- including quotation marks around the SPF policy (which `dig` would display as `… IN TXT "\"v=spf1 …\""`);
- copy-pasting the entire resource record in the data field (e.g `example. 3600 IN TXT "example. 3600 IN TXT \"v=spf1 …\""`).

I haven’t quantified this phenomenon yet but maybe it’s something to do in a separate test case.

> This PR is a draft. Are Zone12 and Zone13 ready for review?

I’ve rewritten Zone12 and Zone13 as well based on your review of Zone11, so yes, they are. I even think that this specification is stable/mature enough for me to mark this PR as ready for review.

docs/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: Yes indeed. That’s fixed now.

matsduf:
> However it is interesting, I think, to look for TXT records that do contain `v=spf1` somewhere in their text. I’ve seen instances of mistakes such as:
> 
>     * including quotation marks around the SPF policy (which `dig` would display as `… IN TXT "\"v=spf1 …\""`);
> 
>     * copy-pasting the entire resource record in the data field (e.g `example. 3600 IN TXT "example. 3600 IN TXT \"v=spf1 …\""`).
> 
> I haven’t quantified this phenomenon yet but maybe it’s something to do in a separate test case.

I think it fits into this test case instead of a separate one. The cases are maybe rare, but Zonemaster should also look for rare cases such as responding to queries for SOA but not MX. At least I think it should be simple to create test zones with the odd data in a TXT record. I think that looking for a TXT record that contains `v=spf` but not being valid SPF would be reasonable. It might not being ERROR, but NOTICE or WARNING.

> I’ve rewritten Zone12 and Zone13 as well based on your review of Zone11, so yes, they are. I even think that this specification is stable/mature enough for me to mark this PR as ready for review.

OK, I will dig into Zone12 and Zone13 then.

marc-vanderwal:
After discussion, I’m going to close this PR in order to split it up into smaller ones.

I also finished conducting some preliminary research on the use of SPF, DKIM and DMARC on the `.fr` zone and published some results on (Afnic’s blog)[https://www.afnic.fr/en/observatory-and-resources/expert-papers/spf-dkim-and-dmarc-whats-the-situation-with-the-fr-tld/]. Erroneous SPF and DMARC policies are very rare, and I don’t think it’s worth giving detailed feedback on possible errors in SPF and DMARC records unless there is sufficient demand for it, which seems unlikely to me.

Based on that research, I’m therefore going to create new specifications that just focus on checking for SPF, DKIM and DMARC on the zone under test and maybe just conduct a basic syntax check on SPF and DMARC policies (I’ve actually used Mail::SPF and Mail::DMARC to do just that during my research, so that’s going to be an easy thing to implement), raising warnings if there’s an obvious error. All of these will go in separate PRs.

PR #1105, opened 2022-10-12
System message - a report for discussion

matsduf:
## Purpose

This PR is not to be merged.

This is a report of all message tags that are outputted with SYSTEM module (level). The report shows some inconsistencies and errors in the messages, and those should be corrected.

* Should all such messages with a msgid be maintained together with the messages from the test cases, i.e. in the Zonemaster/Zonemaster component in a file or database?
* Should SYSTEM messages without msgid be handled in the same way or direct in the Zonemaster-Engine code as today?

### DIFF
--- a/docs/specifications/tests/System-Messages.md
+++ b/docs/specifications/tests/System-Messages.md
@@ -1,0 +1,121 @@
+# SYSTEM messages
+
+## Current status
+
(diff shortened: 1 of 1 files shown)


matsduf:
Not to be done.

PR #1106, opened 2022-10-14
Update GUI build environment instructions

matsduf:
## Purpose

This PR updates the instructions to create a GUI build environment. The version of Node has increased.

## How to test this PR

Documentation only, review and try.

### DIFF
--- a/docs/internal-documentation/distrib-testing/Ubuntu-Node.js-build-environment.md
+++ b/docs/internal-documentation/distrib-testing/Ubuntu-Node.js-build-environment.md
@@ -27,14 +27,14 @@ here.
    source ~/.bashrc
    ```
 
-5. Install the supported Node.js version
+5. Install the Node.js version that supports Zonemaster GUI
    ```sh
-   nvm install 18.11.0
+   nvm install 16.18.0
    ```
 
 6. Switch to the previously installed version
    ```sh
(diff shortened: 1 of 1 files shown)


blacksponge:
This could impact dependencies not supporting the new node version. From the issues in GUI it seems that there are issue building against this version. I would advice to not upgrade before we figure out why. Also node 16 is still supported (maintenance mode) for a year.

matsduf:
@blacksponge, please re-review.

Is it correct to change from **v0.38.0** to **v0.39.2** in "https://raw.githubusercontent.com/nvm-sh/nvm/XXXX/install.sh"?

docs/internal-documentation/distrib-testing/Ubuntu-Node.js-build-environment.md
```diff
@@ -27,14 +27,14 @@ here.
    source ~/.bashrc
    ```
 
-5. Install the supported Node.js version
+5. Install the Node.js version that supports Zonemaster GUI
    ```sh
-   nvm install 16
+   nvm install 16.18.0
```
blacksponge, line 32: We have not strong requirement on the minor version and so I think we should not mention it (and definitely not mention the patch version, in case of security upgrade). The latest minor should work well.

PR #1109, opened 2022-10-19
Updates DNSSEC02: Better message for non-matching DS

matsduf:
## Purpose

The domain (zone) dnssec-failed.org fails DNSSEC on purpose. The DS record for the zone (there is only one) does not match any DNSKEY record in the zone. This is an issue that should be captured by DNSSEC02, and it does report the problem with the DS, but it does not report that there is no "working" DS, i.e. the zone is bogus.
```
# zonemaster-cli --show-testcase --test dnssec/dnssec02 dnssec-failed.org
Seconds Level     Testcase       Message
======= ========= ============== =======
   8.52 WARNING   DNSSEC02       The DNSKEY record with tag 106 that the DS refers to does not exist in the DNSKEY RRset. Fetched from the nameservers with IP "2001:558:1004:7:68:87:85:132;2001:558:100a:5:68:87:68:244;2001:558:100e:5:68:87:72:244;2001:558:1014:c:68:87:76:228;2001:558:fe23:8:69:252:250:103;68.87.68.244;68.87.72.244;68.87.76.228;68.87.85.132;69.252.250.103".
```

This PR adds a message that will be outputted if there is no DS matching any DNSKEY, and another message if there is such a DNSKEY, but that DNSKEY has not signed the DNSKEY RRset.

## Changes

* Adds message when no DS matches a valid DNSKEY.
* Adds message when no DS matches a DNSKEY that has correctly signed the DNSKEY RRset
* Includes msgid for all message tags. For the unchanged tags the msgid is just copied from the implementation and requires no change of implementation.
* Modifies structure go make it easier to follow the flow, but does not change any logic besides the logic that outputs the new messages.

## How to test this PR

This PR requires reviewing.


docs/specifications/tests/DNSSEC-TP/dnssec02.md
```diff
@@ -67,15 +67,17 @@ inconsistent.
...
+Message Tag outputted              | Level   | Arguments                              | Description of when message tag is outputted
+:----------------------------------|:--------|:---------------------------------------|:-----------------------------------------------------------------------
+DS02_ALGO_NOT_SUPPORTED_BY_ZM      | NOTICE  |ns_ip_list, algo_mnemo, algo_num, keytag| DNSKEY with tag {keytag} uses unsupported algorithm {algo_num} ({algo_mnemo}) by this installation of Zonemaster. Fetched from the nameservers with IP addresses "{ns_ip_list}".
+DS02_DNSKEY_NOT_FOR_ZONE_SIGNING   | ERROR   | ns_ip_list, keytag                     | Flags field of DNSKEY record with tag {keytag} has not ZONE bit set although DS with same tag is present in parent. Fetched from the nameservers with IP addresses "{ns_ip_list}".
+DS02_DNSKEY_NOT_SEP                | NOTICE  | ns_ip_list, keytag                     | Flags field of DNSKEY record with tag {keytag} has not SEP bit set although DS with same tag is present in parent. Fetched from the nameservers with IP addresses "{ns_ip_list}".
```
tgreenx, line 74: "has not" -> "does not have"

docs/specifications/tests/DNSSEC-TP/dnssec02.md
```diff
@@ -147,11 +152,13 @@ queries follow, unless otherwise specified below, what is specified for
...
           section of the [DNSSEC Response].
-    3. Retrieve the DNSKEY RRset (store as "DNSKEY RRs") from the
+    3. Add the name server IP address to the *Responding Child Name Servers* set.
+    4. Retrieve the DNSKEY RRset (store as "DNSKEY RRs") from the
        [DNSSEC Response].
-    4. Retrieve the RRSIG records covering the DNSKEY RRset, possibly
+    5. Retrieve the RRSIG records covering the DNSKEY RRset, possibly
        none (store as "DNSKEY RRSIG") from the [DNSSEC Response].
-    5. For each DS in *DS Records*, do:
+    6. Empty the *DNSKEY Match DS* set.
```
tgreenx, line 160: This isn't needed as the set is never used before.

docs/specifications/tests/DNSSEC-TP/dnssec02.md
```diff
@@ -162,52 +169,78 @@ queries follow, unless otherwise specified below, what is specified for
...
           DNSKEY record key tag and name server IP to the *DNSKEY not SEP*
           set.
-       6. Find the equivalent RRSIG in *DNSKEY RRSIG* by key ID (key tag). If
-          there is more than one such RRSIG record, select the correct one.
-       7. If matching RRSIG is not found, add DNSKEY record key tag and name
-          server IP to the *No matching DNSKEY RRSIG* set.
-       8. Else, if the Zonemaster installation does not have support for the
+       6. Add the DNSKEY record, key tag, name server IP to the *DNSKEY Match DS*
+          set.
+    7. Empty the *Has DNSKEY RRSIG Match DS* set.
```
tgreenx, line 182: This isn't needed as the set is never used before.

tgreenx (commented):
Just a few comments otherwise LGTM.

docs/specifications/tests/DNSSEC-TP/dnssec02.md
matsduf: It must be empty each time a new name server is inspected. It is defined as a "global variable" but it is used as a "local variable".

docs/specifications/tests/DNSSEC-TP/dnssec02.md
matsduf: It must be empty each time a new name server is inspected. It is defined as a "global variable" but it is used as a "local variable".

The two "empty" statements could be put just after the "for each child named server" if you think that is better.

docs/specifications/tests/DNSSEC-TP/dnssec02.md
tgreenx: I think that is problematic. Because this set is also used in step 14, which is supposed to contain data for all name servers. Right now it will contain data for only one name server, namely the last in the *Child NS IP* set.

docs/specifications/tests/DNSSEC-TP/dnssec02.md
tgreenx: Same comment as above. *Has DNSKEY RRSIG Match DS* set is also used in step 16.

docs/specifications/tests/DNSSEC-TP/dnssec02.md
matsduf: I will investigate.

docs/specifications/tests/DNSSEC-TP/dnssec02.md
matsduf: I will investigate.

matsduf:
@tgreenx, please re-review.

tgreenx (approved):

matsduf:
Resolved by https://github.com/zonemaster/zonemaster-engine/pull/1158

Issue #1134, opened 2023-02-01
Thoughts and suggestion on our git workflow


pnax:
From my understanding, our current git workflow is based on successive refinements performed across the years. Based on the discussions and questions I've asked and delving into the git history since I'm working on this project, this is what I've understood.

### Stage 0: everyone worked into master

At the beginning of the project, few people were involved. It was quite easy and common to directly work into _master_.

### Stage 1: let's use the merge request feature

To clarify and track changes, pull requests were used. This introduced the usage of external branches to conduct the work. Everyone still pushed its merge into _master_. Sometimes commits were made directly into _master_ without request.

### Stage 2: develop in "develop", release in "master"

At some point (in 2017) it was decided to conduct the work in a separate branch _develop_ that will be merged into _master_ at release time.

This allowed to always have a branch with a working state of the component that is pointing to its latest released version.

### Stage 3: how can we handle a revert?

> edit: this stage was not exactly the one encountered. In fact the problem encountered was that some merge requests were merged in _master_ instead of _develop_ and then reverted in _master_. See https://github.com/zonemaster/zonemaster/pull/1137#pullrequestreview-1279862975, https://github.com/zonemaster/zonemaster/pull/1137#issuecomment-1413433216 for a discussion and https://github.com/zonemaster/zonemaster/pull/1137#issuecomment-1422441811 for an explanation.

This was great until a release was done that was broken. A rollback occurred. This translates into a _revert_ in _master_. Then the work was fixed in _develop_ and merged into _master_. This is were something bad happened. Reverting a change in git can be tricky since it does remove the data but does not remove the associated commits from the history [[1],[2],[3],[4],[5]].

To avoid future issue when dealing with potential rollback (and revert), new steps were introduced during the release process. The main idea is to use an intermediate branch between _develop_ and _master_ at release time that glue those branch and history together. Hence reverting the merge between _merge-develop-into-master_ (the glue branch name) won't be a trouble when fixing the error and doing a new release.

As of [v2022.2][merge develop into master], these are the steps to perform this task:

```                                                        
git checkout -b merge-develop-into-master origin/develop
git reset --soft origin/master
git commit -m 'Update master to state of develop'
git merge origin/develop

# the next commands are the equivalent to the merge request
git checkout master
git merge --no-ff merge-develop-into-master
```

One drawback of this method is that the git history seen from the _master_ branch is hidden in one commit. Commands such as `git blame` will show their utility from the _develop_ branch.

[...]

PR #1137, opened 2023-02-01
Release documentation: update git workflow

pnax:
## Purpose

Ease the release process and keep a light and clean git history.

## Context

Addresses https://github.com/zonemaster/zonemaster/issues/1134

## Changes

Internal document ReleaseProcess-release.md

## How to test this PR

Check that the steps are explicit enough.

### DIFF
--- a/docs/internal-documentation/maintenance/ReleaseProcess-release.md
+++ b/docs/internal-documentation/maintenance/ReleaseProcess-release.md
@@ -407,6 +407,7 @@ parent number is `1`.
 
 <!-- Zonemaster links point on purpose on the develop branch. -->
 [Appendix A]:                                    #appendix-a-on-version-number-in-makefilepl
+[Appendix B]:                                    #appendix-b-on-reverting-commits
 [Backend.pm]:                                    https://github.com/zonemaster/zonemaster-backend/blob/develop/lib/Zonemaster/Backend.pm
 [Build Environment Preparation]:                 ../distrib-testing/BuildEnvironmentPreparation.md
 [Build environment for Node.js]:                 ../distrib-testing/Ubuntu-Node.js-build-environment.md
--- a/docs/internal-documentation/maintenance/ReleaseProcess-release.md
+++ b/docs/internal-documentation/maintenance/ReleaseProcess-release.md
@@ -22,7 +22,6 @@ Release process - Release
 * [17. Announce the release](#17-announce-the-release)
 * [18. Merge master into develop](#18-merge-master-into-develop)
 * [Appendix A on version number in Makefile.PL](#appendix-a-on-version-number-in-makefilepl)
-* [Appendix B on how to verify merge develop branch into master branch](#appendix-b-on-how-to-verify-merge-develop-branch-into-master-branch)
 
 ## 1. Overview
 
@@ -359,42 +358,8 @@ the "v", which may affect the version comparison.
 
 [(Top)](#table-of-contents)
 
-## Appendix B on how to verify merge develop branch into master branch
-
-> Note: appendix to remove
-
-Below are two ways of verifying that the branch created for merging
-develop branch into master branch is correct. The first is to use the
-local clone where the `merge-develop-into-master` branch was created.
-
-Run the command
-
-```
-git diff origin/develop merge-develop-into-master
-```
-The result should be no difference, i.e. the contents of the new branch
-is the same as the develop branch. Then run the following two commands
-```
-git merge-base origin/develop merge-develop-into-master
-git show-ref --hash origin/develop
-```
-The two commands should report the same commit.
-
-The second way of verifying is to use Githubs tool for creating pull
-requests. Start to create a pull request from
-`merge-develop-into-master` develop `master` (but do not really create
-it) and inspect what files that Github says will be updated, added or
-deleted.
-
-If Github says that no files are updated, added or created, then
-branch `merge-develop-into-master` is correct, and a pull request into
-`master` branch (not `develop` branch) can be created.
-
-[(Top)](#table-of-contents)
-
 <!-- Zonemaster links point on purpose on the develop branch. -->
 [Appendix A]:                                    #appendix-a-on-version-number-in-makefilepl
-[Appendix B]:                                    #appendix-b-on-how-to-verify-merge-develop-branch-into-master-branch
 [Backend.pm]:                                    https://github.com/zonemaster/zonemaster-backend/blob/develop/lib/Zonemaster/Backend.pm
 [Build Environment Preparation]:                 ../distrib-testing/BuildEnvironmentPreparation.md
 [Build environment for Node.js]:                 ../distrib-testing/Ubuntu-Node.js-build-environment.md
--- a/docs/internal-documentation/maintenance/ReleaseProcess-release.md
+++ b/docs/internal-documentation/maintenance/ReleaseProcess-release.md
@@ -267,25 +267,21 @@ Make sure you're up to date and your working directory is completely clean:
 > **Note:** To throw away any and all changes to tracked and untracked files you
 > can run `git clean -dfx ; git reset --hard`.
 
-Create a new branch named `merge-develop-into-master` with both *master* and
-*develop* as ancestors and with the exact contents from *develop*:
+Verify that there are no commits belonging both to `master` and `develop` (this
+could occur when pull requests are merged in the wrong branch):
 
-    # We want the contents of the *develop* branch
-    git checkout -b merge-develop-into-master origin/develop
+    TAG=$( git tag --list --sort=-creatordate | head -n 1 )
+    cat <( git rev-list $TAG^2..develop ) <( git rev-list $TAG..master ) | sort | uniq -d
 
-    # But on top of the history of the *master* branch
-    git reset --soft origin/master
+If the previous command returns such commits, they should have been reverted in
+`master`. Look for such reverts and revert them in another branch
+`prepare-release`, that can be merged into `master` without review or approval:
 
-    # With all differences squashed into a single commit
-    git commit -m 'Update master to state of develop'
+    git checkout -b prepare-release origin/master
+    git revert <COMMIT...>
 
-    # And we want the history of *develop* merged too
-    git merge origin/develop
-
-Create a pull request from `merge-develop-into-master` into `master` and merged
-it. No reviewer or approval is required for this update.
-
-In [Appendix B] it is shown how `merge-develop-into-master` can be verified.
+Finally create a pull request from `develop` into `master` and merge it. No
+reviewer or approval is required for this update.
 
 [(Top)](#table-of-contents)
 
@@ -364,6 +360,8 @@ the "v", which may affect the version comparison.
 [(Top)](#table-of-contents)
 
 ## Appendix B on how to verify merge develop branch into master branch
+
+> Note: appendix to remove
 
 Below are two ways of verifying that the branch created for merging
 develop branch into master branch is correct. The first is to use the
--- a/docs/internal-documentation/maintenance/ReleaseProcess-release.md
+++ b/docs/internal-documentation/maintenance/ReleaseProcess-release.md
@@ -270,19 +270,12 @@ Make sure you're up to date and your working directory is completely clean:
 Verify if there are commits on `master` since last release:
 
     TAG=$( git tag --list --sort=-creatordate | head -n 1 )
-    git rev-list --count $TAG..master
+    # count the number of commits in master since last release and
+    # if the count is not 0, look for commits belonging to master and develop
+    [ $(git rev-list --count $TAG..master) -ne 0 ] && cat <( git rev-list $TAG^2..develop ) <( git rev-list $TAG..master ) | sort | uniq -d
 
-If the previous command returns a count greater than 0, check for commits
-belonging both to `master` and `develop` branches since last release:
-
-    cat <( git rev-list $TAG^2..develop ) <( git rev-list $TAG..master ) | sort | uniq -d
-
-If the previous command returns such commits, it's assumed they have been
-reverted in `master`. Such reverts need to be reverted (yes revert of the
-revert) in `master` before merging `develop`. [Appendix B] gives some clues on
-how to perform such revert.
-
-Otherwise, everything looks good for the merge.
+If commits are returned by this previous command, some additional work is
+needed. See [Appendix B] for that.
 
 Finally once `master` branch is ready, create a pull request from `develop`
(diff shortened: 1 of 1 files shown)


docs/internal-documentation/maintenance/ReleaseProcess-release.md
```diff
@@ -267,25 +268,9 @@ Make sure you're up to date and your working directory is completely clean:
 > **Note:** To throw away any and all changes to tracked and untracked files you
 > can run `git clean -dfx ; git reset --hard`.
 
-Create a new branch named `merge-develop-into-master` with both *master* and
-*develop* as ancestors and with the exact contents from *develop*:
+Create a pull request from `develop` into `master` and merge it.
```
matsduf, line 271: Add "No reviewer or approval is required for this update" as today.

docs/internal-documentation/maintenance/ReleaseProcess-release.md
```diff
@@ -267,25 +268,9 @@ Make sure you're up to date and your working directory is completely clean:
...
-    git commit -m 'Update master to state of develop'
-
-    # And we want the history of *develop* merged too
-    git merge origin/develop
-
-Create a pull request from `merge-develop-into-master` into `master` and merged
-it. No reviewer or approval is required for this update.
-
-In [Appendix B] it is shown how `merge-develop-into-master` can be verified.
+In [Appendix C] it is shown how to rollback if needed.
```
matsduf, line 273: Can you explain when this step will be taken in the release process?

docs/internal-documentation/maintenance/ReleaseProcess-release.md
```diff
@@ -394,6 +381,36 @@ branch `merge-develop-into-master` is correct, and a pull request into
...
+3] for more about reverting a merge.
+
+Once the problem is fixed in _develop_, prepare the _master_ branch for the
+[merge of _develop_][merge develop into master] by re-reverting the revert
+commit introduced:
+
+```
+git revert <revert commit>
+git push origin master
+```
```
matsduf, line 410: I do not see when this will be used in the release process.

pnax:
I had a different use case in mind while working on this. However the one you're referring to can be addressed the same way.

For the record, the use case I had in mind: after a release (R2), it appears the release contains severe broken code. A new release (R3) reverting (R2) should be published without notice, with R3=R1 (where R1 was the release before R2).

The use case you're sharing: a merge happens on _master_ instead of _develop_ and should be reverted and merged into _develop_ instead.

Both use cases are about reverting a merge commit. I'll update the documentation to reflect that 2nd use case that fits more into the release process.

pnax:
> Instead of resetting and removing the commits we have reverted those commits. After that the PR was merged to the develop branch. When, at release we merged develop into master there was some issues.

Rethinking about this part, I tried to test this case and could not produce any issue. I think there might be issues if the code is not reverted from _master_. Could that be what happened?

matsduf:
> For the record, the use case I had in mind: after a release (R2), it appears the release contains severe broken code. A new release (R3) reverting (R2) should be published without notice, with R3=R1 (where R1 was the release before R2).

It has not happened, but it might happen so it is good to have some steps for that. Reverting means going back in version number, and the question is if that is what we want.

In any case, that is not part of step 14. Such a problem will be discovered later, most likely after the release i completed. I suggest that a step 19 with the headline "Post-release considerations" or something is created, and there the reference to the new appendix with some more text, can be added.

> The use case you're sharing: a merge happens on _master_ instead of _develop_ and should be reverted and merged into _develop_ instead.

That is also what we have done in the past, reverting the commits on master branch and merging into develop branch, but that has created problems in some cases, and that is the background to the current way of merging the content into master.

You did not comment my suggestion on creating the tag in develop branch before merging it into master branch. Then we could probably skip the step of merging master into develop. There will then be a difference of merge commits, but is that a problem?

matsduf:
> Rethinking about this part, I tried to test this case and could not produce any issue. I think there might be issues if the code is not reverted from _master_. Could that be what happened?

I am not what you mean with "not reverted from master", but as far as I remember the revert was done in the master branch. I do not see why we would done it otherwise.

pnax:
> I am not what you mean with "not reverted from master"

I meant, maybe the commits were kept in master but not reverted. Which could indeed resolving conflicts at release time.

> You did not comment my suggestion on creating the tag in develop branch before merging it into master branch.

See https://github.com/zonemaster/zonemaster/issues/1134#issuecomment-1412455618

> Reverting means going back in version number, and the question is if that is what we want.

Would this be problematic? In my opinion this revert scenario is only necessary in the case we can't wait for a fix release.

matsduf:
> I meant, maybe the commits were kept in master but not reverted. Which could indeed entail in resolving conflicts at release time.

The commits were reverted.

> > You did not comment my suggestion on creating the tag in develop branch before merging it into master branch.
> 
> See [#1134 (comment)](https://github.com/zonemaster/zonemaster/issues/1134#issuecomment-1412455618)

I do not understand why it would be a problem to set the tag in develop branch in the same way as we set the program version in develop branch.

> > Reverting means going back in version number, and the question is if that is what we want.
> 
> Would this be problematic? In my opinion this revert scenario is only necessary in the case we can't wait for a fix release.

If program version goes backwards then it will be hard to update from CPAN. Maybe reverting could be a good idea in some cases, but that should be followed by a new version with higher version number than the faulty version so that updates work.

matsduf:
There is a reason why the complex merging of develop into master was created. I do not think we should remove that without fully understand it. It less problematic to remove the merging from master to develop. The only real thing that is lost are the tags.

pnax:
> Maybe reverting could be a good idea in some cases, but that should be followed by a new version with higher version number than the faulty version so that updates work

Ok, this will be left to the release process to bump the version after reverting.

> There is a reason why the complex merging of develop into master was introduced. I do not think we should remove that without fully understand it. It is less problematic to remove the merging from master to develop. The only real thing that is lost are the tags.

I found the explanation of why sometimes there is a conflict and sometimes not when some commits were reverted from _master_ and ported to _develop_.

It appears that sometimes the exact same commit (call it `c0`) appears both in _master_ and _develop_. This is ok. This has just the down effet that in _master_ this commit `c0` is reverted. And in _develop_ changes might appear after `c0` that update similar part of the code. Therefore when merging, git sees some conflict.

On the other hand, sometimes, this commit `c0` is copied (say using `cherry-pick`) to _develop_, say as commit `c1`. This will result in `c0` in _master_ and `c1` in _develop_. In this particular case, there won't be any conflict when merging _develop_ into _master_.

The followinng commands will print commits that belongs both to _develop_ and _master_ since the latest release:

```
# retrieve the latest tag release
TAG=$( git tag --list --sort=-creatordate | head -n 1 )
# print out commit belonging both to develop and master
cat <( git rev-list $TAG^2..develop ) <( git rev-list $TAG..master ) | sort | uniq -d
```

Explanations:

 * `git rev-list $TAG^2..develop`: get all commits between commit `$TAG^2` and commit referenced by `develop`. `$TAG^2` commit is the latest commit on _develop_ branch before the previous release referenced by $TAG.

 * `git rev-list $TAG..master`: get all commit on _master_ since latest release referenced by $TAG.

 * `cat <(...) <(...) | sort | uniq -d`: concatenate and sort all the found commits, then keep only the one found both in _develop_ and _master_

For instance if you look at the work done on Zonemaster-Engine between release v3.1.2 and v4.0.0, you can see that commit `bcfb180` belongs to both branch. To try this out, you could play with the following commands (this will recreate the context just befor release v4.0.0):

```
git clone https://github.com/zonemaster/zonemaster-engine.git
cd zonemaster-engine
git branch test-master e76be44   # set master tipping point
git branch test-develop 2561ae8  # set develop tipping poing
TAG=v3.1.2
cat <( git rev-list $TAG^2..test-develop ) <( git rev-list $TAG..test-master ) | sort | uniq -d
```

This should print commit `bcfb180`.

Now try to merge _test-develop_ into _test-master_ (this should result in a conflict, that you can abort with `git merge --abort`):

```
git checkout test-master
git merge --no-ff test-develop
```

Let's go back to a clean state:

```
git merge --abort
```

Now there should be a commit on _master_ that reverts the faulty commits, here
commit `6d50f13` (and associated PR #747, commit `b8145d2`) reverts commit
`bcfb180`. This revert should be reverted:

```
git checkout test-master
git revert -m 1 b8145d2
```

Now properly merge _test-develop_ into _test-master_:

```
git checkout test-master
git merge --no-ff test-develop
```

No conflict here. You can verify that this merge results in the same content as
the official merge tagged with `v4.0.0`, the command should not output any
diff:

```
git diff test-master v4.0.0
```

matsduf:
> It appears that sometimes the exact same commit (call it `c0`) appears both in _master_ and _develop_. This is ok. This has just the down effet that in _master_ this commit `c0` is reverted. And in _develop_ changes might appear after `c0` that update similar part of the code. Therefore when merging, git sees some conflict.

Yes, and that is why we have the complex merging of develop into master.

> For instance if you look at the work done on Zonemaster-Engine between release v3.1.2 and v4.0.0, you can see that commit `bcfb180` belongs to both branch. To try this out, you could play with the following commands (this will recreate the context just befor release v4.0.0):
> 
(...)
> 
> This should print commit `bcfb180`.

Couldn't there be common commits between the develop and master even if none has been reverted?

> Now there should be a commit on _master_ that reverts the faulty commits, here commit `6d50f13` (and associated PR #747, commit `b8145d2`) reverts commit `bcfb180`. This revert should be reverted:

How did you find the commits reverting `bcfb180`? Why did you choose to revert `b8145d2` instead of `6d50f13`?

> ```
> git checkout test-master
> git revert -m 1 b8145d2
> ```

How did you know that it should be first parent number ("-m 1")?

The method above seems to solve the issue, and maybe in a better way, but we have to have documentation for the release process that makes reasonable to execute the release.

1. How would the release officer know that reverting is needed?
2. How would the release officer find the commit to revert?
3. How would the release officer know parent number to revert?

docs/internal-documentation/maintenance/ReleaseProcess-release.md
```diff
@@ -267,25 +266,21 @@ Make sure you're up to date and your working directory is completely clean:
...
 
-Create a new branch named `merge-develop-into-master` with both *master* and
-*develop* as ancestors and with the exact contents from *develop*:
+Verify that there are no commits belonging both to `master` and `develop` (this
+could occur when pull requests are merged in the wrong branch):
 
-    # We want the contents of the *develop* branch
-    git checkout -b merge-develop-into-master origin/develop
+    TAG=$( git tag --list --sort=-creatordate | head -n 1 )
+    cat <( git rev-list $TAG^2..develop ) <( git rev-list $TAG..master ) | sort | uniq -d
```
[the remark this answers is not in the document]
pnax: > Should this list all commits from the previous release up today?

yes

> Maybe add "since last release".

will do

docs/internal-documentation/maintenance/ReleaseProcess-release.md
```diff
@@ -267,25 +266,21 @@ Make sure you're up to date and your working directory is completely clean:
...
 
-    # But on top of the history of the *master* branch
-    git reset --soft origin/master
+If the previous command returns such commits, they should have been reverted in
+`master`. Look for such reverts and revert them in another branch
+`prepare-release`, that can be merged into `master` without review or approval:
```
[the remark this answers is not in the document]
pnax: I'll address those.

docs/internal-documentation/maintenance/ReleaseProcess-release.md
pnax: This appendix has been removed for now.

docs/internal-documentation/maintenance/ReleaseProcess-release.md
pnax: Removed for now as this is not exactly addressing the problem

docs/internal-documentation/maintenance/ReleaseProcess-release.md
pnax: done

pnax:
> Couldn't there be common commits between the develop and master even if none has been reverted?

The command I provided will list all common commits without knowledge if they were reverted or not.

> How did you find the commits reverting bcfb180? Why did you choose to revert b8145d2 instead of 6d50f13?

I found it using `git log --graph --oneline test-v3.1.2..test-master` I choose the merge commit because there might be several commits in the merge request.

> How did you know that it should be first parent number ("-m 1")?

It was merged into _master_ which is the first parent of the merge.

matsduf:
In general I think this change is fine. I dislike to loose the commit history. My concern is that the release process is well-defined. We know that we have had reverting in master branch before, and we will probably have it again. We should be able to handle the issue.

pnax:
> 1. How would the release officer know that reverting is needed?
> 2. How would the release officer find the commit to revert?
> 3. How would the release officer know parent number to revert?

I tried to come up with answers to those questions that could be included in the documentation if it fulfills our needs

#### How would the release officer know that reverting is needed?

If there are commits in _master_ since latest release, then check for shared commits between _develop_ and _master_.

```
# retrieve the latest tag release
TAG=$( git tag --list --sort=-creatordate | head -n 1 )

# count the number of commits since
git rev-list --count $TAG..master

# if the count is not 0
# print out commits belonging both to develop and master
cat <( git rev-list $TAG^2..develop ) <( git rev-list $TAG..master ) | sort | uniq -d
```

If this latest command returns commits, then their associated reverts need to be reverted.

#### How would the release officer find the commit to revert?

Usually a revert commits contains a predefined title (or subject) looking like: `Revert "subject of the reverted commit"`.

She can looks for commits with this string using a command like:

```
for c in $(cat <( git rev-list $TAG^2..develop ) <( git rev-list $TAG..master ) | sort | uniq -d); \
    do git log --oneline --grep="Revert \"$(git show -s --pretty=%s $c)\"" $c~..master ^$TAG ; \
    done
```

Or look for them visually with:

```
git log --oneline --graph $c~..master ^$TAG
```

Once found, either revert each commits (`git revert <commit> ...`) or the merge commit.

#### How would the release officer know parent number to revert?

In case she reverts the merge commit, since it was merged into master, the parent number is `1`.

```
git revert -m 1 <merge commit>
```

matsduf:
Thanks, that looks fine. Please work that into the document with enough details. Since this will hopefully be rare, the check for commits on top of master branch should be part of the process, but the special steps in case there are could preferably be put into an appendix to keep the process shorter.

pnax:
I've updated the document based on our discussion. Please re-review.

matsduf (approved):
This looks fine.

PR #1145, opened 2023-03-08
Updates test case BASIC02

matsduf:
## Purpose

This PR adjusts BASIC02 in two respects:

1. One message tag is raised from DEBUG to INFO. The reason it was DEBUG was that GUI presented all messages in a long list before and the INFO messages drowned other messages. With the new (current) presentation per test case it makes sense to include the message as INFO.
2. The order in which AA flag and RCODE were checked meant that configurations with unexpected RCODE were reported as non-AA, which is true, but less informative.

## Changes

The test case specification is updated. The update must also be implemented.

## How to test this PR

This is specification only. Review the update.


matsduf:
@tgreenx, can I merge this?

matsduf:
@marc-vanderwal, please add your review.

marc-vanderwal (approved):
Looks good to me.

PR #1146, opened 2023-04-06
Add specification for testing SPF

### LINKED ISSUE #995: Add SPF and possibly other mail related test cases [Feature Request]

marc-vanderwal:
## Purpose

This PR adds a simple test case for SPF on a domain under test.

## Context

Following up on, and simplifying, PR #1100.

## Changes

Adds one test case which tests SPF.

## How to test this PR

This is a documentation update.

### DIFF
--- a/docs/specifications/tests/Zone-TP/zone11.md
+++ b/docs/specifications/tests/Zone-TP/zone11.md
@@ -40,14 +40,15 @@ server.
 
 ## Summary
 
-Message Tag                     | Level   | Arguments    | Message ID for message tag
-:-------------------------------|:--------|:-------------|:--------------------------------------------
-Z11_INCONSISTENT_SPF_POLICIES   | WARNING |              | The *Child Zone* publishes different SPF policies on different name servers.
-Z11_NO_SPF_FOUND                | DEBUG   |              | The *Child Zone* does not publish an SPF policy.
-Z11_SPF1_MULTIPLE_RECORDS       | ERROR   | ns_ip_list   | The *Child Zone* publishes more than one SPF version 1 policy. Policies retrieved from the following nameservers: {ns_ip_list}.
-Z11_SPF1_SYNTAX_ERROR           | ERROR   | ns_ip_list   | The *Child Zone*’s SPF version 1 policy has a syntax error. Policy retrieved from the following nameservers: {ns_ip_list}.
-Z11_SPF1_SYNTAX_OK              | INFO    |              | The *Child Zone*’s SPF version 1 policy has correct syntax.
-Z11_UNABLE_TO_CHECK_FOR_SPF     | ERROR   |              | None of the name servers for *Child Zone* responded with an authoritative response to queries for SPF policies.
+Message Tag                      | Level   | Arguments    | Message ID for message tag
+:--------------------------------|:--------|:-------------|:--------------------------------------------
+Z11_INCONSISTENT_SPF_POLICIES    | WARNING |              | The *Child Zone* publishes different SPF policies on different name servers.
+Z11_DIFFERENT_SPF_POLICIES_FOUND | NOTICE  | ns_ip_list   | The following name servers returned the same SPF version 1 policy, but other name servers returned a different policy. Name servers: {ns_ip_list}.
+Z11_NO_SPF_FOUND                 | DEBUG   |              | The *Child Zone* does not publish an SPF policy.
+Z11_SPF1_MULTIPLE_RECORDS        | ERROR   | ns_ip_list   | The following name servers returned more than one SPF version 1 policy for the *Child Zone*. Name servers: {ns_ip_list}.
+Z11_SPF1_SYNTAX_ERROR            | ERROR   | ns_ip_list   | The *Child Zone*’s SPF version 1 policy has a syntax error. Policy retrieved from the following nameservers: {ns_ip_list}.
+Z11_SPF1_SYNTAX_OK               | INFO    |              | The *Child Zone*’s SPF version 1 policy has correct syntax.
+Z11_UNABLE_TO_CHECK_FOR_SPF      | ERROR   |              | None of the name servers for *Child Zone* responded with an authoritative response to queries for SPF policies.
 
 The value in the Level column is the default severity level of the message. The
 severity level can be changed in the [Zonemaster-Engine profile]. Also see the
@@ -109,8 +110,14 @@ same specification.
 
 7. Compare the set of *SPF-Policies* retrieved from all name servers. If at
    least two different name servers have returned different sets of SPF
-   policies, then output *[Z11_INCONSISTENT_SPF_POLICIES]* and terminate the
-   test.
+   policies, then:
+
+   1. Output *[Z11_INCONSISTENT_SPF_POLICIES]*.
+   2. Group *SPF-Policies* by equal sets of SPF policies, such that a set of
+      SPF policies is mapped to the list of *Name Server IPs* that returned it.
+   3. For each such group of name servers, output
+      *[Z11_DIFFERENT_SPF_POLICIES_FOUND]*.
+   4. Terminate the test.
 
 8. If the *SPF-Policies* set contains at least two pairs with the same IP
    address, then output *[Z11_SPF1_MULTIPLE_RECORDS]* with the list of
@@ -191,6 +198,7 @@ None.
 [Test Case Identifier Specification]:   TestCaseIdentifierSpecification.md
 [Undelegated test]:                     ../../test-types/undelegated-test.md
 [WARNING]:                              ../SeverityLevelDefinitions.md#warning
+[Z11_DIFFERENT_SPF_POLICIES_FOUND]:     #summary
 [Z11_INCONSISTENT_SPF_POLICIES]:        #summary
 [Z11_NO_SPF_FOUND]:                     #summary
 [Z11_SPF1_MULTIPLE_RECORDS]:            #summary
--- a/docs/specifications/tests/Zone-TP/zone11.md
+++ b/docs/specifications/tests/Zone-TP/zone11.md
@@ -18,14 +18,14 @@
 
 ## Objective
 
-Sender Policy Framework (SPF), described in [RFC 7208][RFC7208], is a mechanism
+Sender Policy Framework (SPF), described in [RFC 7208], is a mechanism
 allowing domain name owners to specify which hosts are allowed to send mail
 claiming to be from that domain. It is implemented by means of TXT records in a
 structured format.
 
-This test case looks up SPF records in the apex of the domain to be tested. It
-checks that there is at most one published SPF version 1 policy and, if
-present, also checks its syntax.
+This test case looks up SPF records in the apex of *Child Zone*. It checks
+that there is at most one published SPF version 1 policy and, if present, also
+checks its syntax.
 
 ## Scope
 
@@ -47,7 +47,7 @@ Z11_NO_SPF_FOUND                | DEBUG   |              | The *Child Zone* does
 Z11_SPF1_MULTIPLE_RECORDS       | ERROR   | ns_ip_list   | The *Child Zone* publishes more than one SPF version 1 policy. Policies retrieved from the following nameservers: {ns_ip_list}.
 Z11_SPF1_SYNTAX_ERROR           | ERROR   | ns_ip_list   | The *Child Zone*’s SPF version 1 policy has a syntax error. Policy retrieved from the following nameservers: {ns_ip_list}.
 Z11_SPF1_SYNTAX_OK              | INFO    |              | The *Child Zone*’s SPF version 1 policy has correct syntax.
-Z11_UNABLE_TO_CHECK_FOR_SPF     | ERROR   |              | None of the name servers for *Child Zone* responded with an authoritative NOERROR response to queries for SPF policies.
+Z11_UNABLE_TO_CHECK_FOR_SPF     | ERROR   |              | None of the name servers for *Child Zone* responded with an authoritative response to queries for SPF policies.
 
 The value in the Level column is the default severity level of the message. The
 severity level can be changed in the [Zonemaster-Engine profile]. Also see the
@@ -92,9 +92,12 @@ same specification.
       and the empty string to the *SPF-Policies* set.
 
    5. If the name server responds with at least one TXT record that is an [SPF
-      TXT record], then, for each [SPF TXT record], add the pair consisting of
-      the *Name Server IP* and the lowercased [concatenation][concatenate] of
-      all strings in the TXT record’s data to the *SPF-Policies* set.
+      TXT record], then, for each [SPF TXT record] do:
+
+       1. [Concatenate] all strings in the RDATA field.
+       2. Lowercase the resulting string.
+       3. Add a pair consisting of the *Name Server IP* and the lowercase
+          string thus derived from the RDATA field to the *SPF-Policies* set.
 
    6. Go to the next name server.
 
@@ -154,11 +157,11 @@ None.
 
 * "concatenate" - The term is used to refer to the conversion of a TXT
   resource record’s data to a single contiguous string, as specified in [RFC
-  7208, section 3.3][RFC7208#3.3].
+  7208, section 3.3][RFC 7208#3.3].
 
 * "passing the syntax check" - The term is used in this document to refer to
-  text that is valid according to the ABNF grammar published in [RFC
-  7208][RFC7208] starting from [section 4.5][RFC7208#4.5].
+  text that is valid according to the ABNF grammar published in [RFC 7208]
+  starting from [section 4.5][RFC 7208#4.5].
 
 * "using Method" - The term is used when data is fetched using the defined
   [Method][Methods].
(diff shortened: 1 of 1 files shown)


docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,201 @@
+# ZONE11: SPF policy validation
+
+## Test case identifier
+**Zone11**
```
matsduf, line 4: To be consistent with the other specifications, `**ZONE11**`

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,201 @@
...
+* [Terminology](#terminology)
+
+## Objective
+
+Sender Policy Framework (SPF), described in [RFC 7208][RFC7208], is a mechanism
+allowing domain name owners to specify which hosts are allowed to send mail
+claiming to be from that domain. It is implemented by means of TXT records in a
+structured format.
+
+This test case looks up SPF records in the domain to be tested. It checks that
```
matsduf, line 26: In the apex of the domain to be tested?

The the zone is name.xa do you put the SPF records in "name.xa. TXT" to apply for "xxx@name.xa"? Do they also apply for "xxx@subname.name.xa"?

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,201 @@
...
+## Summary
+
+Message Tag                     | Level   | Arguments    | Message ID for message tag
+:-------------------------------|:--------|:-------------|:--------------------------------------------
+Z11_INCONSISTENT_SPF_POLICIES   | WARNING |              | The *Child Zone* publishes different SPF policies on different name servers.
+Z11_NO_SPF_FOUND                | DEBUG   |              | The *Child Zone* does not publish an SPF policy.
+Z11_SPF1_MULTIPLE_RECORDS       | ERROR   | ns_ip_list   | The *Child Zone* publishes more than one SPF version 1 policy. Policies retrieved from the following nameservers: {ns_ip_list}.
+Z11_SPF1_SYNTAX_ERROR           | ERROR   | ns_ip_list   | The *Child Zone*’s SPF version 1 policy has a syntax error. Policy retrieved from the following nameservers: {ns_ip_list}.
+Z11_SPF1_SYNTAX_OK              | INFO    |              | The *Child Zone*’s SPF version 1 policy has correct syntax.
+Z11_UNABLE_TO_CHECK_FOR_SPF     | ERROR   |              | Unable to check whether *Child Zone* publishes an SPF policy.
```
matsduf, line 50: This is a case when none of the servers responds in a correct way. Could the message ID be clearer? "None of the name servers could be checked for SPF policy for *Child Zone*" or something?

matsduf (changes requested):
It looks good but I have some minor comments.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,201 @@
...
+* [Terminology](#terminology)
+
+## Objective
+
+Sender Policy Framework (SPF), described in [RFC 7208][RFC7208], is a mechanism
+allowing domain name owners to specify which hosts are allowed to send mail
+claiming to be from that domain. It is implemented by means of TXT records in a
+structured format.
+
+This test case looks up SPF records in the apex of the domain to be tested. It
```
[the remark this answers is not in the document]
marc-vanderwal: Works for me. Fixing.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,201 @@
...
+## Summary
+
+Message Tag                     | Level   | Arguments    | Message ID for message tag
+:-------------------------------|:--------|:-------------|:--------------------------------------------
+Z11_INCONSISTENT_SPF_POLICIES   | WARNING |              | The *Child Zone* publishes different SPF policies on different name servers.
+Z11_NO_SPF_FOUND                | DEBUG   |              | The *Child Zone* does not publish an SPF policy.
+Z11_SPF1_MULTIPLE_RECORDS       | ERROR   | ns_ip_list   | The *Child Zone* publishes more than one SPF version 1 policy. Policies retrieved from the following nameservers: {ns_ip_list}.
+Z11_SPF1_SYNTAX_ERROR           | ERROR   | ns_ip_list   | The *Child Zone*’s SPF version 1 policy has a syntax error. Policy retrieved from the following nameservers: {ns_ip_list}.
+Z11_SPF1_SYNTAX_OK              | INFO    |              | The *Child Zone*’s SPF version 1 policy has correct syntax.
+Z11_UNABLE_TO_CHECK_FOR_SPF     | ERROR   |              | None of the name servers for *Child Zone* responded with an authoritative NOERROR response to queries for SPF policies.
```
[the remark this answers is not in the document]
marc-vanderwal: I’m fine with that. Fixing.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,201 @@
...
+      *SPF-Policies* set.
+
+   4. If the name server responds with at least one TXT record and none is an
+      [SPF TXT record], then add the pair consisting of the *Name Server IP*
+      and the empty string to the *SPF-Policies* set.
+
+   5. If the name server responds with at least one TXT record that is an [SPF
+      TXT record], then, for each [SPF TXT record], add the pair consisting of
+      the *Name Server IP* and the lowercased [concatenation][concatenate] of
+      all strings in the TXT record’s data to the *SPF-Policies* set.
```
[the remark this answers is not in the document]
marc-vanderwal: Works for me. I’ll go with your suggestion, with minor modifications.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,201 @@
...
+
+* "Child Zone" - The domain name to be tested.
+
+## Summary
+
+Message Tag                     | Level   | Arguments    | Message ID for message tag
+:-------------------------------|:--------|:-------------|:--------------------------------------------
+Z11_INCONSISTENT_SPF_POLICIES   | WARNING |              | The *Child Zone* publishes different SPF policies on different name servers.
+Z11_NO_SPF_FOUND                | DEBUG   |              | The *Child Zone* does not publish an SPF policy.
+Z11_SPF1_MULTIPLE_RECORDS       | ERROR   | ns_ip_list   | The *Child Zone* publishes more than one SPF version 1 policy. Policies retrieved from the following nameservers: {ns_ip_list}.
```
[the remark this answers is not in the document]
marc-vanderwal: I don’t know about that. I chose this writing in order to stay consistent with how other error messages from Zonemaster are written.

How about:
`The following nameservers returned more than one SPF version 1 policy for the *Child Zone*: {ns_ip_list}`
or something like that (maybe `for the *Child Zone*` could be left out)?

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,201 @@
...
+5. If the *SPF-Policies* set is empty, then output
+   *[Z11_UNABLE_TO_CHECK_FOR_SPF]* and terminate the test.
+
+6. If all the pairs in the *SPF-Policies* set contain empty strings, then
+   output *[Z11_NO_SPF_FOUND]* and terminate the test.
+
+7. Compare the set of *SPF-Policies* retrieved from all name servers. If at
+   least two different name servers have returned different sets of SPF
+   policies, then output *[Z11_INCONSISTENT_SPF_POLICIES]* and terminate the
+   test.
```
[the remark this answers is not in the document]
marc-vanderwal: It could, but there are many edge cases where it’s hard to decide which is the “good” SPF policy and which one should be fixed, or where the message could be misleading.

Between two nameservers, it’s impossible to decide. We can just say that nameserver A returned something and nameserver B returned something else.

Between three nameservers, this message would only make sense if two out of three had the same SPF policy and only one has a different string. If all three are inconsistent, we can’t output this message.

More generally speaking, if there are N nameservers (N > 2) and among them, M (where 2 < M < N) agree on the same data and P other servers (where 1 ⩽ P < M) disagree, it’s easy to determine which is the odd one out. Otherwise we can only something along the lines of “no nameserver agrees on an SPF policy”.

This isn’t the only area where we might run into this problem; ideally we should find a generic solution.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,201 @@
...
+  into one string, yields a string either equal to `v=spf1` or starting with
+  `v=spf1` followed by a space, irrespective of character case.
+
+* "concatenate" - The term is used to refer to the conversion of a TXT
+  resource record’s data to a single contiguous string, as specified in [RFC
+  7208, section 3.3][RFC7208#3.3].
+
+* "passing the syntax check" - The term is used in this document to refer to
+  text that is valid according to the ABNF grammar published in [RFC
+  7208][RFC7208] starting from [section 4.5][RFC7208#4.5].
```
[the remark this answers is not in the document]
marc-vanderwal: Fixing

docs/specifications/tests/Zone-TP/zone11.md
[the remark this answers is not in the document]
marc-vanderwal: After further discussion, I realized that I misunderstood what you meant. I’ll implement the suggestion.

docs/specifications/tests/Zone-TP/zone11.md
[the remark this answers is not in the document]
matsduf: I think `for the *Child Zone*` can be left out. Your suggestion looks fine.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,208 @@
...
+
+* "Child Zone" - The domain name to be tested.
+
+## Summary
+
+Message Tag                      | Level   | Arguments    | Message ID for message tag
+:--------------------------------|:--------|:-------------|:--------------------------------------------
+Z11_INCONSISTENT_SPF_POLICIES    | WARNING |              | The *Child Zone* publishes different SPF policies on different name servers.
+Z11_DIFFERENT_SPF_POLICIES_FOUND | NOTICE  | ns_ip_list   | The following name servers returned the same SPF version 1 policy, but other name servers returned a different policy. Name servers: {ns_ip_list}.
+Z11_NO_SPF_FOUND                 | DEBUG   |              | The *Child Zone* does not publish an SPF policy.
```
matsduf, line 47: With the new GUI layout (messages grouped by test case) I think it makes sense to raise this to INFO. I think it makes sense to always have some message from each test case.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,208 @@
...
+
+   1. Output *[Z11_INCONSISTENT_SPF_POLICIES]*.
+   2. Group *SPF-Policies* by equal sets of SPF policies, such that a set of
+      SPF policies is mapped to the list of *Name Server IPs* that returned it.
+   3. For each such group of name servers, output
+      *[Z11_DIFFERENT_SPF_POLICIES_FOUND]*.
+   4. Terminate the test.
+
+8. If the *SPF-Policies* set contains at least two pairs with the same IP
+   address, then output *[Z11_SPF1_MULTIPLE_RECORDS]* with the list of
```
matsduf, line 123: I think it would be easier to read if it says "... two entries with the same IP address ...".

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,208 @@
...
+  into one string, yields a string either equal to `v=spf1` or starting with
+  `v=spf1` followed by a space, irrespective of character case.
+
+* "concatenate" - The term is used to refer to the conversion of a TXT
+  resource record’s data to a single contiguous string, as specified in [RFC
+  7208, section 3.3][RFC 7208#3.3].
+
+* "passing the syntax check" - The term is used in this document to refer to
+  text that is valid according to the ABNF grammar published in [RFC 7208]
+  starting from [section 4.5][RFC 7208#4.5].
```
matsduf, line 171: It is not easy to many verify this unless you are familiar with the syntax. Are there any on-line SPF validation tools that we could refer to, not as an normative reference, but something that the reader could use to check with?

(I realize that the same comment could be put on the DNSSEC test case specification where validation is done.)

matsduf (changes requested):
It looks fine, but I have a few comments. Then I can approve.

docs/specifications/tests/Zone-TP/zone11.md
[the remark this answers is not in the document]
marc-vanderwal: Good. Correcting.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,208 @@
+# ZONE11: SPF policy validation
+
+## Test case identifier
+**ZONE11**
+
+## Table of contents
+
+* [Objective](#objective)
+* [Caveat](#caveat)
```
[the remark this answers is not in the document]
marc-vanderwal: Indeed; fixing.

docs/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: Yes, I agree; I believe I originally wanted this message to be INFO level. Fixing.

docs/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: Fixing.

docs/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: There is a rather nice SPF syntax checker here: <https://vamsoft.com/support/tools/spf-syntax-validator>. It’s one of the few where you can actually input an SPF policy to check, instead of a domain name.

In my draft implementation for which I’ve yet to create a pull request, I am using the syntax check from the Mail::SPF module. Based on a sample of 50 SPF policies with Mail::SPF deems to be in error, the aforementioned syntax checker says that 40 have a syntax error, 9 are “OK with warnings” (for example `v=spf1 ip4:192.0.2.25 +a +mx +include:spf.jabatus.fr ~allgoogle-site-verification=yjZfTeZ8FHjaH4tY8UZvrEc_enSwtJFRtyJtbNy7RPg`, with the actual IP address replaced with one in a documentation range) and only 1 was “OK”.

The only discrepancy I was with an SPF policy containing a newline (U+000A) character. In that case, Mail::SPF will raise an error, but the online tool said it was OK. However, the ABNF grammar for SPF only allows spaces (U+0020) as whitespace, so Mail::SPF is correct here.

docs/specifications/tests/Zone-TP/zone11.md
```diff
@@ -0,0 +1,210 @@
...
+
+* "Child Zone" - The domain name to be tested.
+
+## Summary
+
+Message Tag                      | Level   | Arguments    | Message ID for message tag
+:--------------------------------|:--------|:-------------|:--------------------------------------------
+Z11_INCONSISTENT_SPF_POLICIES    | WARNING |              | The *Child Zone* publishes different SPF policies on different name servers.
+Z11_DIFFERENT_SPF_POLICIES_FOUND | NOTICE  | ns_ip_list   | The following name servers returned the same SPF version 1 policy, but other name servers returned a different policy. Name servers: {ns_ip_list}.
+Z11_NO_SPF_FOUND                 | INFO    |              | The *Child Zone* does not publish an SPF policy.
```
matsduf, line 46: Looking at this again, is it a problem that SPF policy is missing? Is that against recommendations? Should we raise the level to NOTICE when there is no policy?

Do you see a connection between the existence of MX and existence of SPF for a domain?

matsduf (dismissed):
I have one comment, but that could be considered later. It would not prevent the implementation.

docs/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: IMHO, not publishing an SPF policy is similar to not signing a zone with DNSSEC: not an error in itself, but maybe something a domain name owner should consider implementing.

According to my data, on the `.fr` zone, 62.6% of zones have both MX records and an SPF policy, and 33.2% of zones have no MX at all but an SPF policy. I assume most of these would be “null SPF” policies (i.e. `v=spf -all`).

The same can be said for DKIM (unless the SPF policy is a so-called “null SPF”, such as `v=spf1 -all`) and DMARC.

marc-vanderwal: In order to be consistent with the level of the message Zonemaster outputs if there is no DNSSEC signature, I think we can raise the message level for `Z11_NO_SPF_FOUND` to NOTICE.

Regardless of whether the domain originates mail, it is a good practice to implement SPF:
* if a domain originates mail, specifying the mail servers authorized to send mail on the domain’s behalf is a basic defense against spoofing and even though SPF alone won’t stop all forms of spoofing, it’s required by DMARC;
* if a domain does not originate mail (for example, when it’s a parked domain), then it’s a [good practice](https://www.m3aawg.org/sites/default/files/m3aawg_parked_domains_bcp-2022-06.pdf) to publish a “null SPF” policy.

matsduf: >     * if a domain does not originate mail (for example, when it’s a parked domain), then it’s a [good practice](https://www.m3aawg.org/sites/default/files/m3aawg_parked_domains_bcp-2022-06.pdf) to publish a “null SPF” policy.

In ZONE09 the test case detects Null MX. Is it easy to detect a Null SPF? Should this test case report Null SPF and update ZONE09 to report Null MX?

matsduf (approved):
This is fine for implementation. My comment can be handled later.

marc-vanderwal:
Either way, let’s merge this!

matsduf:
@marc-vanderwal, you should move the specification to the new path.

marc-vanderwal:
Sure, I’ve just created a pull request for that.

PR #1150, opened 2023-04-19
Create Test Case "Checking for revealed software version": Nameserver15

### LINKED ISSUE #1126: Create test case checking for software version
Following discussions during F2F, it was decided that it would be useful to have a new test case that check if a nameserver responds to CHAOS TXT queries regarding `version.bind`, and if so report that version at the NOTICE level.

```
$ dig @ns2.nic.fr version.bind chaos txt +short
"9.18.9"
```

tgreenx:
## Purpose

This PR proposes a specification for a new Test Case titled "Checking for revealed software version". The proposed Test Case Identifier is Nameserver15.

## Context

Fixes #1126 

## Changes

- New specification (docs/public/specifications/tests/Nameserver-TP/nameserver15.md)

## How to test this PR

N/A. Documentation to review only.

### DIFF
--- a/docs/public/specifications/tests/Nameserver-TP/nameserver15.md
+++ b/docs/public/specifications/tests/Nameserver-TP/nameserver15.md
@@ -36,8 +36,8 @@ giving a correct DNS response for an authoritative name server.
 
 Message Tag                 | Level   | Arguments                      | Message ID for message tag
 :---------------------------|:--------|:-------------------------------|:----------------------------------------------------------------------------------------------------------------------------
-N15_SOFTWARE_VERSION        | INFO    | ns_ip_list, query_name, string | The following name server(s) respond to version query "{query_name}" with string "{string}". Returned from name servers: "{ns_ip_list}"
-N15_NO_VERSION              | INFO    | ns_ip_list                     | The following name server(s) do not respond to version queries. Returned from name servers: "{ns_ip_list}"
+N15_SOFTWARE_VERSION        | INFO    | ns_ip_list, query_name, string | The following name server(s) respond to software version query "{query_name}" with string "{string}". Returned from name servers: "{ns_ip_list}"
+N15_NO_VERSION              | INFO    | ns_ip_list                     | The following name server(s) do not respond to software version queries. Returned from name servers: "{ns_ip_list}"
 
 The value in the Level column is the default severity level of the message. The
 severity level can be changed in the [Zonemaster-Engine Profile]. Also see the
@@ -78,8 +78,8 @@ message. The argument names are defined in the [Argument List].
       add name server to the *No Version* set.
 
 6. If the *TXT Data* set is non-empty, then, for each unique
-   TXT string in the set, output *[N15_SOFTWARE_VERSION]* with
-   name server IP list, query name and TXT string.
+   string and query name pair in the set, output *[N15_SOFTWARE_VERSION]*
+   with name server IP list, query name and string.
 
 7. If the *No Version* set is non-empty, then output
    *[N15_NO_VERSION]* with name server IP list.


docs/public/specifications/tests/Nameserver-TP/nameserver15.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Inputs
+
+* "Child Zone" - The domain name to be tested.
+
+## Summary
+
+Message Tag                 | Level   | Arguments                     | Message ID for message tag
+:---------------------------|:--------|:------------------------------|:---------------------------------------------------------------------------------------
+N15_BIND_VERSION            | INFO    | ns_ip_list, string            | Name server(s) "{ns_ip_list}" respond(s) to CHAOS TXT query "version.bind" with value "{string}".
```
matsduf, line 40: The "ns_ip_list" can be a long list. I always put that at the end or else reading may break.

Since the value can be anything, isn't "with string" better?

"string" is not a defined argument name. Should it be "string" or "bind_version"? Please update the argument list.

Is "CHAOS TXT" relevant here? Or should it rather be "name server version query"?

docs/public/specifications/tests/Nameserver-TP/nameserver15.md
```diff
@@ -0,0 +1,119 @@
+# NAMESERVER15: Checking for software version
+
+## Test case identifier
+
+**NAMESERVER15**
```
matsduf, line 5: I think there should be no empty line here. I might not matter, but we have scripts reading the files.

docs/public/specifications/tests/Nameserver-TP/nameserver15.md
```diff
@@ -0,0 +1,119 @@
+# NAMESERVER15: Checking for software version
```
matsduf, line 1: Usually it says "Test for". Is "checking" better?

docs/public/specifications/tests/README.md
```diff
@@ -132,6 +132,7 @@ project. All the details are in the [Master Test Plan].
 |[NAMESERVER12](Nameserver-TP/nameserver12.md)|Test for unknown EDNS flags|
 |[NAMESERVER13](Nameserver-TP/nameserver13.md)|Test for truncated response on EDNS query|
 |[NAMESERVER14](Nameserver-TP/nameserver14.md)|Test for unknown version with unknown OPTION-CODE|
+|[NAMESERVER15](Nameserver-TP/nameserver15.md)|Checking for software version|
```
matsduf, line 135: This file is to be updated at release time so it is not needed to include the change now.

docs/public/SUMMARY.md
```diff
@@ -99,6 +99,7 @@
             - [Nameserver12](specifications/tests/Nameserver-TP/nameserver12.md)
             - [Nameserver13](specifications/tests/Nameserver-TP/nameserver13.md)
             - [Nameserver14](specifications/tests/Nameserver-TP/nameserver14.md)
+            - [Nameserver15](specifications/tests/Nameserver-TP/nameserver15.md)
```
matsduf, line 102: This file is to be updated at release time so it is not needed to include the change now.

matsduf (changes requested):
With the new GUI design I think it is desirable to have some output with level INFO-ERROR from each test case. If no name servers supports "version.bind" then nothing is outputted. Maybe a message (INFO) "No name server outputs anyting on name server version query" if none of the nameservers gave any output?

docs/public/specifications/tests/Nameserver-TP/nameserver15.md
marc-vanderwal: How about naming this test “NAMESERVER15: Test for name servers revealing their software's version numbers”?

docs/public/specifications/tests/Nameserver-TP/nameserver15.md
marc-vanderwal: I’d move the `{ns_ip_list}` at the end, but apart from that, the message ID looks fine to me.

docs/public/specifications/tests/Nameserver-TP/nameserver15.md
```diff
@@ -0,0 +1,119 @@
...
+severity level can be changed in the [Zonemaster-Engine Profile]. Also see the
+[Severity Level Definitions] document.
+
+The argument names in the Arguments column lists the arguments used in the
+message. The argument names are defined in the [Argument List].
+
+## Test procedure
+
+1. Create the following empty sets:
+   1. Name server IP and string ("TXT Data")
```
marc-vanderwal, line 52: No need for a list if there is only one item in it.

docs/public/specifications/tests/Nameserver-TP/nameserver15.md
```diff
@@ -0,0 +1,119 @@
...
+3. Obtain the set of name server IP addresses using [Method4] and [Method5]
+   ("Name Server IP").
+
+4. For each name server in *Name Server IP* do:
+   1. [Send] *TXT Query* to the name server and collect the response.
+   2. If at least one of the following criteria is met, go to next name server:
+      1. There is no DNS response.
+      2. The [DNS Response] does not have the [RCODE Name] NoError.
+      3. The [DNS Response] does not have any TXT record in the
+         answer section with owner name "version.bind".
```
marc-vanderwal, line 66: What if there are two or more TXT resource records? In this case I suggest going to the next name server.

docs/public/specifications/tests/Nameserver-TP/nameserver15.md
```diff
@@ -0,0 +1,119 @@
...
+   3. Extract the string from the RDATA of the TXT record in the
+      answer section of the [DNS Response].
```
marc-vanderwal, line 68: There should be a definition of what “extracting the string” means, because a TXT resource record can contain one or more strings. I suggest that each string be concatenated into one large string without addition of whitespace, as if by `join("", @strings)`, like what is done for SPF and DMARC.

marc-vanderwal (changes requested):
Looks already pretty good to me, but there are small issues that should be addressed.

docs/public/specifications/tests/Nameserver-TP/nameserver15.md
matsduf: It could be different on different servers.

docs/public/specifications/tests/Nameserver-TP/nameserver15.md
matsduf: I rather suggest that both are added to the *TXT Data* set.

docs/public/specifications/tests/Nameserver-TP/nameserver15.md
matsduf: Good point. Should that process be added to "Default handling of a DNS Response" in [DNS Query and Response Defaults] so that we do not have to repeat that every time a TXT record is processed?

[DNS Query and Response Defaults]: https://github.com/zonemaster/zonemaster/blob/develop/docs/public/specifications/tests/DNSQueryAndResponseDefaults.md

docs/public/specifications/tests/Nameserver-TP/nameserver15.md
matsduf: There is also "version.server" to query for, and then a more generic msgid is reasonable.

docs/public/specifications/tests/Nameserver-TP/nameserver15.md
```diff
@@ -0,0 +1,138 @@
...
+## Objective
+
+This Test Case verifies if a name server responds to TXT queries in the CHAOS class, specifically
+about its software version as it may sometimes be desirable not to reveal that information.
+
+A description of DNS classes can be found in [RFC2929], section 3.2.
+
+## Scope
+
+It is assumed that *Child Zone* is also tested and reported by [Connectivity01].
```
matsduf, line 27: The link `[Connectivity01]` is not defined.

docs/public/specifications/tests/Nameserver-TP/nameserver15.md
```diff
@@ -0,0 +1,138 @@
...
+   1. Name server IP, query name and string ("TXT Data")
+   2. Name server IP ("No Version")
+
+2. Create [DNS Query] with query type TXT and query class CHAOS ("TXT Query").
+
+3. Create the set of query names with values "version.bind"
+   and "version.server" ("Query Names").
+
+4. Obtain the set of name server IP addresses using [Method4] and
+   [Method5] ("Name Server IP").
```
matsduf, line 61: The links `[Method4]` and `[Method5]` are not defined.

docs/public/specifications/tests/Nameserver-TP/nameserver15.md
```diff
@@ -0,0 +1,138 @@
...
+   1. For each query name in *Query Names* do:
+      1. [Send] *TXT Query* with query name to
+         the name server and collect the response.
+      2. If at least one of the following criteria is met, go to next query name:
+         1. There is no DNS response.
+         2. The [DNS Response] does not have the [RCODE Name] NoError.
+         3. The [DNS Response] does not have any TXT record in the
+            answer section with query name as owner name.
+      3. For each TXT record in the answer section of the [DNS Response]
+         with owner name query name, extract and [concatenate] the string(s)
```
matsduf, line 73: The link `[concatenate]` is not defined.

docs/public/specifications/tests/Nameserver-TP/nameserver15.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Intercase dependencies
+
+None
+
+## Terminology
+
+* "Concatenate" - The term is used to refer to the conversion of a TXT
+  resource record’s data to a single contiguous string, as specified in [RFC
+  7208, section 3.3[RFC7208#3.3].
```
matsduf, line 112: Incorrect link:
```
[RFC 7208, section 3.3[RFC7208#3.3]
```

matsduf (commented):
It looks fine, besides some errors in links.

docs/public/specifications/tests/Nameserver-TP/nameserver15.md
```diff
@@ -0,0 +1,142 @@
...
+         with owner name query name, extract and [concatenate] the string(s)
+         from the RDATA of the record.
+      4. If the extracted string is non-empty, add *Name Server IP*, query name
+         and the string to the *TXT Data* set.
+   2. If nothing was added to the *TXT Data* set in the previous steps,
+      add name server to the *No Version* set.
+
+6. If the *TXT Data* set is non-empty, then, for each unique
+   TXT string in the set, output *[N15_SOFTWARE_VERSION]* with
+   name server IP list, query name and TXT string.
```
matsduf, line 82: "...for each unique combination of string and query name in the set, output..."

In the definition of the set the field is "string" not "TXT string".

marc-vanderwal (approved):
Looks good to me.

matsduf:
This TC does not specify that CNAME should be followed, which Zonemaster by default does not do. I do not expect any CNAME records for these names, but just to make sure we do not disregard it by mistake.

matsduf (approved):

matsduf:
@tgreenx, do you want me to set test zones up in the #1151 framework?

PR #1152, opened 2023-04-22
Fix typos and inconsistencies in DNSSEC16 and DNSSEC17 specifications

matsduf:
## Purpose

Typos in the text was found in the DNSSEC16 specification. Related inconsistencies between some messages and the logic were found in DNSSEC16 and DNSSEC17 specifications. The messages are updated to match the logic.

No change of logic are needed in the implementation of the two test cases, but the messages must be updated in the implementation when this PR has been merged.

## How to test this PR

Review the messages and the logic of the test cases.


matsduf:
No update of the messages in the implementation is needed. Those are good enough, but not identical. In the specifications the arguments are missing in the messages.

PR #1157, opened 2023-05-31
Update installation for Rocky Linux 8.8 and 9.2

mattias-p:
## Purpose

This PR updates the installation instruction for Rocky Linux 8.8 and 9.2.

## Context

This was done as part of installation testing for 2023.1.

## Changes

* Rocky Linux 9 has replaced PowerTools with CRB (both needed for EPEL) so we need to enable the right one. Fortunately there's a uniform way to do that.
* Rocky Linux 9 disables SHA-1 by default so we need to explicitly enable it.
* Rocky Linux 9 doesn't install locales by default so we need to install them explicitly.
* There's been some changes to available packages in Rocky Linux, so we adapt to that.

## How to test this PR

N/A


docs/public/installation/zonemaster-backend.md
```diff
@@ -111,9 +111,9 @@ cd `perl -MFile::ShareDir=dist_dir -E 'say dist_dir("Zonemaster-Backend")'`
 sudo install -v -m 755 -d /etc/zonemaster
 sudo install -v -m 640 -g zonemaster ./backend_config.ini /etc/zonemaster/
 sudo install -v -m 775 -g zonemaster -d /var/log/zonemaster
-sudo install -v -m 755 ./zm-rpcapi.lsb /etc/init.d/zm-rpcapi
-sudo install -v -m 755 ./zm-testagent.lsb /etc/init.d/zm-testagent
 sudo install -v -m 755 ./tmpfiles.conf /usr/lib/tmpfiles.d/zonemaster.conf
+sudo install -v -m 664 ./zm-rpcapi.service /etc/systemd/system/
```
marc-vanderwal, line 115: Wouldn’t 644 be a better file mode than 664, in order to mirror the 755 mode formerly used for the init scripts?

marc-vanderwal (commented):
Looks good to me, save for one remark.

docs/public/installation/zonemaster-backend.md
mattias-p: I just followed the [documentation](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/assembly_working-with-systemd-unit-files_configuring-basic-system-settings#proc_creating-custom-unit-files_assembly_working-with-systemd-unit-files). It looked funny to me too but I thought maybe that there's a reason for doing things that way. But it's probably a typo in the documentation. I discussed this with a couple of ops people and they agreed with the typo hypothesis.

pnax (approved):

PR #1178, opened 2023-07-13
Removes limitation on multiple batches

matsduf:
## Purpose

The purpose is described in https://github.com/zonemaster/zonemaster-backend/pull/1118 where the change is implemented. The statement that a user only can have one active batch running is removed.

## Changes

The specification of the RPC-API.

## How to test this PR

Review the document and compare with https://github.com/zonemaster/zonemaster-backend/pull/1118

### DIFF
--- a/docs/public/using/backend/rpcapi-reference.md
+++ b/docs/public/using/backend/rpcapi-reference.md
@@ -1316,8 +1316,8 @@ A [*batch id*][Batch id].
 
 #### `"error"`
 
-* If the given `profile` is not among the [available profiles][Profile sections],
-  a user error is returned, see the [profile name section][profile name].
+If the given `profile` is not among the [available profiles][Profile sections], a
+user error is returned, see the [profile name section][profile name].
 
 Trying to add a batch when wrong [*username*][Username] or [*api key*][Api key] is used:
 ```json
--- a/docs/public/using/backend/rpcapi-reference.md
+++ b/docs/public/using/backend/rpcapi-reference.md
@@ -1252,6 +1252,10 @@ A [*username*][Username] and its [*api key*][Api key] can be added with the
 
 *Tests* enqueud using this method are assigned a [*priority*][Priority] of 5.
 
+> In previous versions of Zonemaster-Backend a new batch could not be created by
+> the same [*username*][Username] if that *username* had created a batch that was
+> not yet finished. That restriction has been removed.
+
 
 Example request:
 ```json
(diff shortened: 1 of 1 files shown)


pnax:
There is a whole other part to remove from the `add_batch_job` documentation too.
https://github.com/zonemaster/zonemaster/blob/27243a55dc3d39306fce40ce3ba1724d0b71ec0c/docs/public/using/backend/rpcapi-reference.md#error-10

matsduf:
> There is a whole other part to remove from the `add_batch_job` documentation too.

Fixed.

matsduf:
@pnax, please re-review.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1248,11 +1248,14 @@ This method is not available if [`RPCAPI.enable_add_batch_job`][RPCAPI.enable_ad
...
 A [*username*][Username] and its [*api key*][Api key] can be added with the
-[`add_api_user`][API add_api_user] method. A [*username*][Username] can only have
-one un-finished *batch* at a time.
+[`add_api_user`][API add_api_user] method.
 
 *Tests* enqueud using this method are assigned a [*priority*][Priority] of 5.
 
+> In previous versions of Zonemaster-Backend a new batch could not be created by
+> the same [*username*][Username] if that *username* had created a batch that was
+> not yet finished. That restriction has been removed.
```
mattias-p, line 1257: I think notes on how this version differs from the previous version belong in the release notes, not in the reference documentation for this version. Don't you agree?

matsduf: There should of course be a comment in the release notes, but sometimes I think such note can be reasonable in the documentation, not forever, but for a while. That is why I added it.

marc-vanderwal: I agree with @matsduf here: it isn’t unusual to state in the documentation that some feature X is available from version Y onwards. A comment in the release notes will most likely be read by administrators of existing Zonemaster installations who are upgrading to a later versions; however, administrators of brand new installations might not think about reading release notes.

I just suggest adding the version number (e.g. “That restriction has been removed in version 2023.2”.)

matsduf: Updated as suggested with version number.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1313,29 +1316,9 @@ A [*batch id*][Batch id].
 
 #### `"error"`
 
-* You cannot create a new batch job if a *batch* with unfinished *tests* already
-  exists for this [*username*][Username].
 * If the given `profile` is not among the [available profiles][Profile sections],
   a user error is returned, see the [profile name section][profile name].
```
marc-vanderwal, line 1320: The bullet point list has only one item left, so how about turning it into a regular paragraph?
```suggestion
If the given `profile` is not among the [available profiles][Profile sections], a
user error is returned, see the [profile name section][profile name].
```

marc-vanderwal (approved):

mattias-p (approved):

PR #1183, opened 2023-07-25
Updates algortithm table in DNSSEC05 specification

matsduf:
## Purpose

Zonemaster needs DNSKEY algorithm mnemonic for all possible algorithms. The IANA table has no mnemonic for resevered or unassigned algorithms. The purpose of this PR is to let this test case defined such mnemonics for the use in Zonemaster implementation of any test case needing it.

This PR does not update any other parts of the specification even though it does not fully match the current test case specification template. That is for another PR.

## Context

See https://github.com/zonemaster/zonemaster/pull/1179#discussion_r1273247849 and the responses on that.

This PR makes this specification authoritative of the table above, which means that [Argument list] must be updated (to-do).

## Changes

This PR does not change any logic and requires not updates to the implementation of the test case.

## How to test this PR

Review the changes.

[Argument list]:  https://github.com/zonemaster/zonemaster-engine/blob/master/docs/logentry_args.md

### DIFF
--- a/docs/public/specifications/tests/DNSSEC-TP/dnssec05.md
+++ b/docs/public/specifications/tests/DNSSEC-TP/dnssec05.md
@@ -16,7 +16,7 @@ reference to the RFC.
 
 The table of algorithms below is copied from [IANA registry]. Only the first
 three columns are copied. The complete table is available at [IANA registry].
-In the table below, however, mnemonic is defined when empty in the IANA table.
+In the table below, however, mnemonic is defined when undefined in the IANA table.
 
 Algorithm no | Algorithm (or description)    | Mnemonic          | Note
 :------------|:------------------------------|:------------------|:----
--- a/docs/public/specifications/tests/DNSSEC-TP/dnssec05.md
+++ b/docs/public/specifications/tests/DNSSEC-TP/dnssec05.md
@@ -24,27 +24,27 @@ Algorithm no | Algorithm (or description)    | Mnemonic          | Note
 1            | RSA/MD5                       | RSAMD5            |
 2            | Diffie-Hellman                | DH                |
 3            | DSA/SHA1                      | DSA               |
-4            | Reserved                      | reserved          | (1)
+4            | Reserved                      | RESERVED          | (1)
 5            | RSA/SHA-1                     | RSASHA1           |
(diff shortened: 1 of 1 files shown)


marc-vanderwal (approved):

matsduf:
@tgreenx and @marc-vanderwal, the Zonemaster defined mnemonics in this PR are lower case, while the IANA defined are in upper case. Should I change to upper case?

marc-vanderwal:
> @tgreenx and @marc-vanderwal, the Zonemaster defined mnemonics in this PR are lower case, while the IANA defined are in upper case. Should I change to upper case?

If “reserved” and the other mnemonics are kept as lowercase, DS10_ALGO_NOT_SUPPORTED_BY_ZM would be rendered as:
>DNSKEY with tag 51966 uses unsupported algorithm 123 (reserved) by this installation of Zonemaster. Fetched from the nameservers with IP addresses "ns.example/2001:db8:8::53".

If switching to uppercase, that would be:
>DNSKEY with tag 51966 uses unsupported algorithm 123 (Reserved) by this installation of Zonemaster. Fetched from the nameservers with IP addresses "ns.example/2001:db8:8::53".

I think the lowercase form looks best.

matsduf:
@marc-vanderwal, what I meant was all uppercase, e.g.

>DNSKEY with tag 51966 uses unsupported algorithm 123 (RESERVED) by this installation of Zonemaster. Fetched from the nameservers with IP addresses "ns.example/2001:db8:8::53".

The IANA defined mnemonics are in all uppercase. Let us say that Gost is not supported, then the same message will be today:

>DNSKEY with tag 51966 uses unsupported algorithm 12 (ECC-GOST) by this installation of Zonemaster. Fetched from the nameservers with IP addresses "ns.example/2001:db8:8::53".

I think that the Zonemaster defined mnemonics should follow the same pattern as the IANA defined mnemonics, i.e. be in all uppercase.

marc-vanderwal:
You’re right, the all-uppercase versions actually look better. I’m fine with this change.

marc-vanderwal (approved):

matsduf:
@marc-vanderwal and @tgreenx, please re-review.

PR #1194, opened 2023-08-07
Set "dash version" on Docker image when Zonemaster-CLI version is unchanged

matsduf:
## Purpose

The current release process requires the Docker image version (for Docker Hub) to be the same as Zonemaster-CLI. That is fine when Zonemaster-CLI itself has been updated, but does not work well if a new image is created because some dependency of Zonemaster-CLI (e.g. Zonemaster-Engine) has been updated.

This updates creates a method to set a "dash version" on the Docker image instead of having to do an "empty" update of Zonemaster-CLI to force a version update.

~~If this PR is approved before v2023.1.3 is released, then https://github.com/zonemaster/zonemaster-cli/pull/348/files can be ignored.~~

## How to test this PR

Review the updated steps and make sure they work.

### DIFF
--- a/docs/internal/maintenance/ReleaseProcess-create-docker-image.md
+++ b/docs/internal/maintenance/ReleaseProcess-create-docker-image.md
@@ -43,7 +43,7 @@ commands to clean up before proceeding (see section "[Handy Docker commands]"):
 
 ## 3. Create Docker images
 
-Clone the three repositories:
+### Clone the three repositories
 
 ```sh
 git clone https://github.com/zonemaster/zonemaster-ldns
@@ -51,7 +51,7 @@ git clone https://github.com/zonemaster/zonemaster-engine
 git clone https://github.com/zonemaster/zonemaster-cli
 ```
 
-Check out right branch depending on the use case.
+### Check out right branch depending on the use case
 
 * Check out `develop` branch when creating an image for release testing:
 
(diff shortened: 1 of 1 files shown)


pnax:
> That is fine when Zonemaster-CLI itself has been updated, but does not work well if a new image is created

I'm not sure I understand the underlying issue here. Could you elaborate on the error?

matsduf:
> I'm not sure I understand the underlying issue here. Could you elaborate on the error?

The highest version of Zonenamaster-CLI on Docker Hub is v6.0.1. Now we have to create a new Docker image with higher version. The current release process sets the same version on the image as the Zonemaster-CLI version. This means that the Zonemaster-CLI version must be increase even though is has not been changed, e.g. to v6.0.2 (see https://github.com/zonemaster/zonemaster-cli/pull/348). We probably do not want to be forced to update the Zonemaster-CLI version just because the Docker image has to be updated due to an update of a dependency to Zonemaster-CLI.

The PR defines a method to set a higher version on the Docker image without increasing the version of Zonemaster-CLI. And the suggestion is to use it for v2023.1.3, i.e. keep the Zonemaster-CLI version to v6.0.1 and set the version of the new Docker image to v6.0.1-1.

matsduf:
@pnax and @tgreenx, please review.

matsduf:
@pnax and @mattias-p, please review. Is this fine to be merged?

marc-vanderwal (approved):

PR #1200, opened 2023-08-30
change letter case for test case identifier

blacksponge:
## Purpose

Improve test case identifier and module readability by preferring capitalized words to upper case.

## Context

https://github.com/zonemaster/zonemaster-gui/issues/437

## Changes

* prefer capitalized words to full uppercase

## How to test this PR

\-

### DIFF
--- a/docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
+++ b/docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
@@ -15,7 +15,7 @@ that test level’s name. The following test levels are defined and available:
 
 The test level name is not case sensitive, but the forms defined above
 must be used when referring to the test levels, i.e. only the first letter
-uppercase, expect for acronyms, in the case all uppercase is used.
+uppercase, expect for acronyms for wich all uppercase is used.
 For exemple "Address" and neither "ADDRESS" nor "address".
 
 The test case identifier in the test case specification (both in the main
--- a/docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
+++ b/docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
@@ -1,6 +1,6 @@
 # Test Case Identifier Specification
 
-All test cases belong to one module, and it gets its name from that module's
+All test cases belong to one module and their names are based on that module’s
 name. The following modules are defined and available:
 
 * Address
--- a/docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
+++ b/docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
@@ -1,7 +1,7 @@
 # Test Case Identifier Specification
 
-All test cases belong to one specific test level, and it gets its name from that
-test level's name. The following test levels are defined and available:
+All test cases belong to one module, and it gets its name from that module's
+name. The following modules are defined and available:
 
 * Address
 * Basic
@@ -13,25 +13,26 @@ test level's name. The following test levels are defined and available:
 * Syntax
 * Zone
 
-The test level name is not case sensitive.
+The module name is not case sensitive, but camel case, as defined above,
+should be prefered.
 
 The test case identifier in the test case specification (both in the main
-headline and in the "Test case identifier" section) uses the test level name
-in all uppercase and has the format: `{Test Level name} + {Index}`
+headline and in the "Test case identifier" section) uses the module name,
+as defined above, and has the format: `{Module name} + {Index}`
 
-When referencing to a test case the test level name may be used in other letter
-case than all uppercase.
+When referencing to a test case the module name may be used in other letter
+case, but for readability, camel case sould be prefered.
 
 The `{Index}` is a two-digit suffix 01-99, and should be selected so that the test
 case identifier is unique. Normally the first free index is selected.
 
 Example of test case identifiers:
 
-* ADDRESS03
-* BASIC04
+* Address03
+* Basic04
 * DNSSEC15
-* ZONE06
+* Zone06
 
-One exceptional test case identifier is `BASIC00`, in which the index
+One exceptional test case identifier is `Basic00`, in which the index
 suffix is out of range of the stated range above. That test case is planned to
 be removed.
--- a/docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
+++ b/docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
@@ -13,15 +13,16 @@ name. The following modules are defined and available:
 * Syntax
 * Zone
 
-The module name is not case sensitive, but camel case, as defined above,
-should be prefered.
+The module name is not case sensitive, but the forms defined above,
+should be used when referring to the modules, i.e. "Address" and neither
+"ADDRESS" nor "address".
 
 The test case identifier in the test case specification (both in the main
 headline and in the "Test case identifier" section) uses the module name,
 as defined above, and has the format: `{Module name} + {Index}`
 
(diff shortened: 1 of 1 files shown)


docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
```diff
@@ -1,7 +1,7 @@
 # Test Case Identifier Specification
 
-All test cases belong to one specific test level, and it gets its name from that
-test level's name. The following test levels are defined and available:
+All test cases belong to one module, and it gets its name from that module's
+name. The following modules are defined and available:
```
marc-vanderwal, line 4:
```suggestion
All test cases belong to one module and their names are based on that module’s name.
The following modules are defined and available:
```

marc-vanderwal (commented):
I only have one little suggestion for one paragraph.

docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
blacksponge: I changed the wording to match your proposal

marc-vanderwal (dismissed):

matsduf (changes requested):
Instead of recommendation I think we should have a "should".

docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
```diff
@@ -13,25 +13,26 @@ test level's name. The following test levels are defined and available:
...
 The test case identifier in the test case specification (both in the main
-headline and in the "Test case identifier" section) uses the test level name
-in all uppercase and has the format: `{Test Level name} + {Index}`
+headline and in the "Test case identifier" section) uses the module name,
+as defined above, and has the format: `{Module name} + {Index}`
 
-When referencing to a test case the test level name may be used in other letter
-case than all uppercase.
+When referencing to a test case the module name may be used in other letter
+case, but for readability, camel case sould be prefered.
```
matsduf, line 24: I do not think these are examples of "camelCase", https://en.wikipedia.org/wiki/Camel_case

"sould" --> "should"

docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
```diff
@@ -13,25 +13,26 @@ test level's name. The following test levels are defined and available:
 * Syntax
 * Zone
 
-The test level name is not case sensitive.
+The module name is not case sensitive, but camel case, as defined above,
+should be prefered.
```
matsduf, line 17:
```
The module name is not case sensitive, but the forms defined above,
should be used when referring to the modules, i.e. "Address" and neither
"ADDRESS" nor "address".
```

docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
```diff
@@ -13,25 +13,26 @@ test level's name. The following test levels are defined and available:
...
-* ADDRESS03
-* BASIC04
+* Address03
+* Basic04
 * DNSSEC15
-* ZONE06
+* Zone06
 
-One exceptional test case identifier is `BASIC00`, in which the index
+One exceptional test case identifier is `Basic00`, in which the index
```
matsduf, line 36: Basic00 has been removed, hasn't it?

docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
blacksponge: done.

docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
blacksponge: Not yet

docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
```diff
@@ -1,7 +1,7 @@
 # Test Case Identifier Specification
 
-All test cases belong to one specific test level, and it gets its name from that
-test level's name. The following test levels are defined and available:
+All test cases belong to one module and their names are based on that module’s
+name. The following modules are defined and available:
```
matsduf, line 4: This document now refer "modules" but previously it referred to "test levels". We should use the same terminology in all places. The reason why this document referred to "test level" is because it is called so in other documents. We should be consistent.

See e.g. https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/tests/MasterTestPlan.md#definition-of-test-levels

matsduf (changes requested):
This document defines the case of the test level identifiers, and I think that is good, but I would like to see stronger wording.

It also tries to rename "test level" to "module", and that might be good, but it requires more changes and it should not start here. Rather in the master test plan.

In the description it is referred to "Camel case", but none of the suggestions are Camel case. "iPhone" is example of Camel case, but "Address" is not.

docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
matsduf: No, "Address" is not PascalCase.

docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
blacksponge: I think the current wording across component is Module,

* Test zone data : https://github.com/zonemaster/zonemaster/blob/397c1281da08d0f3323fae6eae1ab54ec0779838/test-zone-data/README.md?plain=1#L94C1-L96C62
* Backend RPCAPI document : https://github.com/zonemaster/zonemaster/blob/397c1281da08d0f3323fae6eae1ab54ec0779838/docs/public/using/backend/rpcapi-reference.md?plain=1#L328
* CLI usage : https://github.com/zonemaster/zonemaster-cli/blob/master/lib/Zonemaster/CLI.pm#L119-L127

and so on.

blacksponge:
> It also tries to rename "test level" to "module", and that might be good, but it requires more changes and it should not start here. Rather in the master test plan.

It does not. The wording "module" is used quite a lot, see, Engine, CLI and Backend implementation.

> In the description it is referred to "Camel case", but none of the suggestions are Camel case. "iPhone" is example of Camel case, but "Address" is not.

It refers to PascalCase not amelCase. "Address" is PascalCase, with one word it would just be equivalent to the capitalized word, but covers the case of something with several words without having to change the document in the future.

docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md
blacksponge: See my other comment.

matsduf:
It is obvious that there is an inconsistency between "test level" and "module". The test plans refer to the IEEE terminology "test level". I guess that "module" has come from the Perl programming.

I am not against leaving "test level" but we should consider if there is any problem of using "module" for the non-programming side of Zonemaster. In the code, module also refer to other parts that do not match any test levels. If we decide to do the change, we should also update other documents to have a complete switch.

When it comes to Pascal case, I cannot see that any of the names of the test levels (or modules) are to be written in Pascal case. I do not think we should use the term in the document. It does not add anything, and it makes the text harder to understand.

"The more specific terms Pascal case and upper camel case refer to a joined phrase where the first letter of each word is capitalized, including the initial letter of the first word." https://en.wikipedia.org/wiki/Camel_case

blacksponge:
All comments should have been addressed.

matsduf (approved):

blacksponge:
@matsduf please re-review, just fixed a typo.

pnax (approved):

PR #1223, opened 2023-11-22
Document dependency on MIME::Base32

marc-vanderwal:
## Purpose

This PR documents a new test-time dependency on MIME::Base32.

## Context

The change was made necessary by PR zonemaster/zonemaster-ldns#178.

## Changes

Add MIME::Base32 as dependency in various places.

## How to test this PR

Documentation update.

### DIFF
--- a/docs/internal/distrib-testing/Debian-build-environment.md
+++ b/docs/internal/distrib-testing/Debian-build-environment.md
@@ -17,7 +17,7 @@ This instruction is for creating it on Debian. See other files for other OSs.
 3. Install prerequisites
 
    ```sh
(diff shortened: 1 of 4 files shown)


tgreenx (approved):

mattias-p:
If it's a test-time dependency should it be included in the installation instruction? We're installing with --notest.

matsduf:
Since the dependency is added to the installation document, it is not needed in the documents under "distrib-testing" as far as I can see. But it does no harm, on the other side.

matsduf (approved):

mattias-p (approved):
> but we also advice running the installation without `--notest`

Fair enough!

PR #1250, opened 2024-04-03
Store release information in public

matsduf:
## Purpose

Add release information to the public document path. **This PR is not to be merged.** It is to show a proposal that would be executed at an upcoming release for that release.

File "public/RELEASE.md" is added. The file should be updated at release time with the release information of the latest release (and not including previous releases). The content is to be directly copied to the Github release page, e.g. https://github.com/zonemaster/zonemaster/releases/tag/v2023.2.1. No edits should be needed.

SUMMARY.md has also been updated with a link to the RELEASE.md file.

I built a "book" and the file works well into the book.

## How to test this PR

Review. Build a "book" following the instruction in https://github.com/zonemaster/zonemaster?tab=readme-ov-file#documentation.

### DIFF
--- a/docs/public/README.md
+++ b/docs/public/README.md
@@ -5,6 +5,7 @@ documentation.*
 
 The public documentation is divided into the following main categories:
 
+* [Release information](RELEASE.md) give the release information of latest release.
 * [Installation](installation/README.md) contains documents that explain how to
   install the different components of Zonemaster.
 * [Upgrading](upgrading/README.md) contains documents that explain how to upgrade
--- a/docs/public/README.md
+++ b/docs/public/README.md
@@ -5,7 +5,7 @@ documentation.*
 
 The public documentation is divided into the following main categories:
 
-* [Release information](RELEASE.md) give the release information of latest release.
+* [Release information](RELEASE.md) gives release information of latest release.
 * [Installation](installation/README.md) contains documents that explain how to
   install the different components of Zonemaster.
 * [Upgrading](upgrading/README.md) contains documents that explain how to upgrade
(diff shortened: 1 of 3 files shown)


mattias-p:
This looks really nice.

Perhaps rename RELEASE.md to RELEASES.md and keep a history of all releases in the same file, or of all releases up to a certain point (e.g. v2023.2.1 and newer). I'm not sure if there's huge value in this but it just feels a bit weird to me to remove the details of the last release every time we make a new one.

matsduf:
My idea is exactly to keep only the latest release notes in the file, and refer to Github release notes for previous releases. Then the content will be exactly the same as is entered in Github, and the focus will be on the latest release.

mattias-p:
Fair enough, I reckoned as much. I had a different idea and I wanted to raise it.

marc-vanderwal (approved):

mattias-p (approved):
I don't believe I've ever seen a software project that maintains a file with information about the current release only. The way people seem to go about things is to keep a proper change log where you can search the entire change history back to the inception of the file using basic tools.  Since we're deviating from this for no apparent reason I feel we're being a bit weird. But having a RELEASE file is better than not having one.

matsduf:
> I don't believe I've ever seen a software project that maintains a file with information about the current release only. The way people seem to go about things is to keep a proper change log where you can search the entire change history back to the inception of the file using basic tools. Since we're deviating from this for no apparent reason I feel we're being a bit weird. But having a RELEASE file is better than not having one.

In the release file there is a link to the historic Changes. The idea with this file is that it should be in the top of the documentation and then a complete file would be too long. Now we have the information for one release. We can discuss it later when we see how it looks like.

matsduf:
This PR has fullfilled is purpose and can be closed now.

Issue #1259, opened 2024-04-30
Separate handling of surrounding space from normalization in specification


mattias-p:
In item 3 of zonemaster/zonemaster-cli#364 a peculiar behavior of Zonemaster CLI is pointed out. I provided a fix in zonemaster/zonemaster-engine#1316 but @matsduf pointed out that the solution doesn't follow the specification in RequirementsAndNormalizationOfDomainNames.md and he wanted to update the specification before item 3 of zonemaster/zonemaster-cli#364 is fixed.

tgreenx:
Fixed by #1260 in release v2025.1.

PR #1260, opened 2024-04-30
Makes removal of leading and trailing space optional when domain name is normalized

matsduf:
## Purpose

In current version of "Requirements and normalization of domain names in input" leading and trailing space characters are to be removed from domain names in all contexts. Issue https://github.com/zonemaster/zonemaster-cli/issues/364 sees it as a problem that a domain name that has been entered within quote marks on a command line should have leading and trailing space removed.

The reporter, however, wants to keep the current behavior when domain names are entered in an HTML form.

This PR make removal of leading and trailing space optional and depending on context, which makes it less predictable what behavior is to be applied. If such space characters are not removed, the domain name will be classified as illegal and no testing will be performed.

It is not obvious that the change in this PR will give any benefits. Also, the specification will be less precise. It will, however, only affect some marginal cases or cases where white space characters have been included by mistake. 

## Context

* https://github.com/zonemaster/zonemaster-cli/issues/364
* #1259
* https://github.com/zonemaster/zonemaster-engine/pull/1316

## Changes

"Requirements and normalization of domain names in input" is updated. If this PR is merged then https://github.com/zonemaster/zonemaster-engine/pull/1316 can be merged.

## How to test this PR

Review.

### DIFF
--- a/docs/public/specifications/tests/RequirementsAndNormalizationOfDomainNames.md
+++ b/docs/public/specifications/tests/RequirementsAndNormalizationOfDomainNames.md
@@ -52,7 +52,11 @@ The process defined in this specification will normalize *Domain Name* and outpu
 a normalized form to be used by all Zonemaster test cases. The objectives of the
 normalization are
 
-1. Remove leading and trailing white space characters.
+1. Optionally remove leading and trailing white space characters. This is done as
+   a pre-process in the client if appropriate, e.g. in fields in an HTLM form in
+   a web form.
+   * If the proprocess is not applied, then any leading or trailing space will be
+     caught as an illegal ASCII character.
 2. Convert other dot characters to regular dot (or "FULL STOP").
 3. Create legal IDNA 2008 U-labels from convenient alternative forms.
 4. Create consistent representation of the same zone name.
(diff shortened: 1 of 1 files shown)


marc-vanderwal:
If there is any whitespace in the illegal domain name, we could also add a message to the user suggesting to run the test on the domain name with all whitespace stripped. Something like:

> The domain name ` example .com ` contains characters that cannot appear in domain names. Did you mean to test `example.com`?

The same could be done when the user input looks like an HTTP URL, because looking through zonemaster.net’s logs shows many attempts at testing domain names such as `https://example.com`. That could look like:

> The domain name `https://example.com` contains illegal characters that cannot appear in domain names. Did you mean to test `example.com`?

In all other cases, we don’t suggest anything.

How about going that route instead? It can be implemented consistently in CLI and GUI, becomes an opportunity to educate the user about the difference between a domain name and a URL, and also avoids “trying to be smarter” than the user by attempting to automatically fix their mistakes.

mattias-p:
After our discussion in the work group meeting I started thinking about this in a pure Zonemaster GUI context. I had a quick look and counted four types of text fields in our UI: domain name, IP address, keytag and digest. If we think the trimming of surrounding space is important enough that we need to specify how it is done, we should specify it for text fields of all types, not only domain names. We could specify it for each type individually like it is done here for domain names, but I don't feel this is the most reasonable approach. A better approach, IMHO, would be to have a separate specification of how surrounding space is trimmed and apply that to all text fields in Zonemaster GUI as a preprocessing step before any validation is done.

matsduf:
It is true that trimming of leading and trailing white space should be done all fields in GUI.

matsduf:
In GUI, if IP address, key ID or digest has white space an error is issued. I think that is done in Backend, isn't it? I think trimming would be more helpful. (Algorithm and digest type are drop down lists.)

mattias-p (dismissed):
This should be enough to go forward with zonemaster/zonemaster-engine#1316, right?

mattias-p:
> In GUI, if IP address, key ID or digest has white space an error is issued. I think that is done in Backend, isn't it? I think trimming would be more helpful. (Algorithm and digest type are drop down lists.)

If we specify trimming in the Backend API we'd have to add trimming rules for every data type in the Backend API that happens to use a text field for input in the GUI. If we switch the input widgets for a field in the GUI, we'd have to update the Backend API as well to adapt to the normalization characteristics of the new widget. IMHO it's better to separate concerns and deal with GUI stuff in GUI.

But for the purpose of zonemaster/zonemaster-engine#1316 this PR should be good enough, and it would be nice to be done with that one in time for the release.

docs/public/specifications/tests/RequirementsAndNormalizationOfDomainNames.md
```diff
@@ -144,41 +148,43 @@ Tables 1, 2, 3 and 4 are found in the [Detailed requirements] section below.
     3. Set of Unicode full stops (dot characters) in Table 4 below
        ("Unicode Full Stops").
 
-2.  If *Domain Name* starts with one or more of *White Space* then those are
-    removed from *Domain Name* before further processing.
+2.  This step only applies if removal of leading and traling spaces has been
+    found to be appropriate for the client and context.
+    1. If *Domain Name* starts with one or more of *White Space* then those are
+       removed from *Domain Name* before further processing.
```
marc-vanderwal, line 154:
```suggestion
2.  This step only applies if removal of leading and trailing spaces has been
    found to be appropriate for the client and context.
    1. If *Domain Name* starts with one or more of *White Space* then those are
       removed from *Domain Name* before further processing.
```

matsduf:
> If we specify trimming in the Backend API we'd have to add trimming rules for every data type in the Backend API that happens to use a text field for input in the GUI. If we switch the input widgets for a field in the GUI, we'd have to update the Backend API as well to adapt to the normalization characteristics of the new widget. IMHO it's better to separate concerns and deal with GUI stuff in GUI.

I am not sure what you mean, but we should both make sure that we do not get garbage into the system and that users do not get error messages on trivial deviations if field content, i.e. where it is obvious what the users want.

Domain names are special and complex, as the document shows, but we could other elements to it such as IP addresses. 

Another example is digest for DS. Bind often shows it two parts, but the space should be ignored. If you enter the digest in that way Zonemaster will complain. Maybe Zonemaster could be helpful, but should probably be up to GUI to do.

marc-vanderwal (approved):

MichaelTimbert (approved):

PR #1261, opened 2024-05-01
file mode 644 for tmpfiles.d/zonemaster.conf

bitmand:
## Purpose

The tmpfiles config file itself should just be `644` - no need for it to be executable.

## Changes

Very small change that updates two `install` commands in the documentation.


matsduf:
@bitmand, thanks for fixing that.

PR #1276, opened 2024-06-24
Remove unused dependency

mattias-p:
## Purpose

Clean up an unused dependency from the installation instruction.

## Context

N/A

## Changes

Text::Reflow is removed from the CLI installation on all OSes.

## How to test this PR

Follow the updated installation instruction.


matsduf (approved):
Should be removed from Travis in Zonemaster-CLI.

PR #1257, opened 2024-04-24
Updates the Basic01 test case specification

matsduf:
## Purpose

There are deficiencies in algorithm to find the parent zone in the current specification. This PR tries to improved the algorithm.

## Changes

The steps to traverse the delegation tree.

## How to test this PR

Review it.

PR #1287, opened 2024-07-05
Update MethodsV2 to match Basic01

matsduf:
## Purpose

Method `Get parent NS IP addresses` in MethodsV2 should use the same algorithm as Basic01, which was true before the change introduced by #1257 which modifies the way the tree is traversed to find the parent zone and its name servers. This PR adapts to that modification.

The change should better match the intention of what the parent zone is. And the change should match the test scenarios created in #1254.

## Context

* #1257
* #1254

## How to test this PR

Review the changes.

### DIFF
--- a/docs/public/specifications/tests/MethodsV2.md
+++ b/docs/public/specifications/tests/MethodsV2.md
@@ -111,8 +111,8 @@ the root zone or if the test is an undelegated test, the list is defined but
 empty. If the parent zone cannot be determined, then an undefined list is
 returned.
 
-This Method must, in general, use the similar algorithm as Test Case [BASIC01],
-but the test case extracts more information.
+This Method must, in general, use the same algorithm as Test Case [Basic01], but
+the test case extracts more information and outputs messages.
 
 ### Inputs
 
@@ -120,8 +120,7 @@ This Method uses the following input units defined in section [Methods Inputs]:
 
 * "Child Zone" - The name of the child zone to be tested.
 * "Root Name Servers"
-* "Test Type" - "undelegated test" or "normal test".
-
+* "Test Type" - "[undelegated test]" or "normal test".
 
 ### Procedures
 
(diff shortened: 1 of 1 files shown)


matsduf:
> Implementation is done in [zonemaster/zonemaster-engine#1373](https://github.com/zonemaster/zonemaster-engine/pull/1373). Current implementation details to note:
> 
>     * The method returns `Zonemaster::Engine::Nameserver` objects, not just IP addresses as the specification describes. It is necessary to have such objects in Engine to be able to perform queries. The method could then be renamed as simply `get_parent_ns`, or a disclaimer can be added (just like with the new test case specifications).

Currently the test case specifications only use the parent name server IP addresses, not the names, so the method only needs to find the addresses. There is no issue if the implementation must have a name as long as we do not start using the name as such. I see no need of update of the specification.

>     * The method will return `undef` if there are more than one parent found (I assume this should not be possible from this new algorithm but still, it is a precaution)

There are still situations when there is more than one parent. Let us say that cc.xa has two NS. One has a delegation to  bb.cc.xa and the other to aa.bb.cc.xa. And parent.grandparent.xa has two NS and both has a delegation aa.bb.cc.xa. The NS for aa.bb.cc.xa are the same in both delegations. Which is the parent of aa.bb.cc.xa? Except DNSSEC aa.bb.cc.xa will work. The two delegations could be different and it could still work if the zone data is the same, or at least relevant data is the same.

I just saw some errors in the specification, and then I again thought about the NXDOMAIN and NODATA situations. Since they do not give any delegation they could be ignored for the method (but not for Basic01). Now NXDOMAIN and NODATA are handled inconsistently by error. I will do an update where both are ignored (not considered to be a parent).

>     * A loop protection mechanism (currently set to 1000) is in place for the loop processing at step 5.11.

I am not sure how you count, but if there are many NS (both v4 and v6) and many levels (like IPv6 reverse) do you see a risk of reaching the limit? Will there be a message (DEBUG) if the loop protection breaks?

It would be interesting to see how many loops the code will do in the normal case. Have you checked?

matsduf:
I will adjust the expected parent data from some of the scenarios (remove server outputting NXDOMAIN or NODATA).

matsduf:
> Ok, I'll make the necessary changes in [the implementation](https://github.com/zonemaster/zonemaster-engine/pull/1373). But maybe the specification needs to be more explicit then? i.e., step 6.

Is it this step 6 you mean?
```
6. If the *Parent Found* set is non-empty then do:
   1. Extract the name server IP list.
   2. Return the following from the Method:
      1. The extracted list of name server IP addresses (parent zone name servers).
   3. Exit these procedures.
```
It just refers to parent name server IP addresses, not parent zone.

marc-vanderwal (approved):

PR #1290, opened 2024-08-28
Clarify MethodsV2?

matsduf:
## Purpose

This PR makes it explicit that CNAME should be followed. Does this match the implementation

## How to test this PR

Review the change and judge if the change is desired and if it matches the implementation.

### DIFF
--- a/docs/public/specifications/tests/MethodsV2.md
+++ b/docs/public/specifications/tests/MethodsV2.md
@@ -172,7 +172,7 @@ This Method uses the following input units defined in section [Methods Inputs]:
        in the additional section.
 
    9.  Do [DNS Lookup] of name server names (A and AAAA) not already listed in the
-       additional section of the response.
+       additional section of the response. Follow CNAME if provided.
        1. For each IP address add the IP address and *Zone Name* to the
           *Remaining Servers* set unless the IP address is already listed in
           *Handled Servers* together with *Zone Name*.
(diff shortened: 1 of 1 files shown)


marc-vanderwal:
In this context, we are resolving A and AAAA resource records for names that we found in NS records, right? Yet, [RFC 2181, § 10.3](https://datatracker.ietf.org/doc/html/rfc2181#section-10.3) disallows pointing to an alias from an NS record. If anything, Zonemaster should report an error in this situation, instead of following CNAMEs.

matsduf:
> Yet, [RFC 2181, § 10.3](https://datatracker.ietf.org/doc/html/rfc2181#section-10.3) disallows pointing to an alias from an NS record. If anything, Zonemaster should report an error in this situation, instead of following CNAMEs.

It is correct that the hostname in the RDATA of an NS record must not have a CNAME record, and Bind will not follow a CNAME. This is the method collecting as many name servers for the zone as possible to be able report as many issues as possible. The methods do not report any issues at all. That is up to the test cases to do.

[Delegation05] checks if the name server names has a CNAME record, and if so outputs `NS_IS_CNAME` on `ERROR` level

[Delegation05]: https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/tests/Delegation-TP/delegation05.md

marc-vanderwal:
Okay, then I understand the rationale. I’d just suggest adding a remark on why Zonemaster follows aliases in that specific case, even though NS records aren’t supposed to point to any. Then it should be fine.

matsduf:
A clarification of the resolution of  CNAME has been added. @marc-vanderwal @tgreenx, please re-review.

marc-vanderwal (approved):

matsduf:
@tgreenx, can I just merge? Is any update in the implementation needed?

PR #1304, opened 2024-10-30
Redefines experiamental RPCAPI batch_status

matsduf:
## Purpose

Current `get_batch_job_result` is not optimal for repeated checking the status of a large batch since always included the test ID of finished tests. This PR redefines experimental `batch_status` which is only an alias for `get_batch_job_result`. With the new definition the response is by default minimal and only including the numbers.

This is a document change only. When this has been reviewed it will also be implemented in Zonemaster-Backend.

## How to test this PR

Review.

### DIFF
--- a/docs/public/using/backend/rpcapi-reference.md
+++ b/docs/public/using/backend/rpcapi-reference.md
@@ -38,14 +38,14 @@
   * [API method: start_domain_test](#api-method-start_domain_test)
   * [API method: test_progress](#api-method-test_progress)
   * [API method: get_test_results](#api-method-get_test_results)
+  * [API method: get_test_params](#api-method-get_test_params)
   * [API method: get_test_history](#api-method-get_test_history)
     * [Undelegated and delegated](#undelegated-and-delegated)
   * [API method: add_api_user](#api-method-add_api_user)
   * [API method: add_batch_job](#api-method-add_batch_job)
   * [API method: get_batch_job_result](#api-method-get_batch_job_result)
-  * [API method: get_test_params](#api-method-get_test_params)
+  * [API method: batch_status](#api-method-batch_status)
 * [Experimental API methods](#experimental-api-methods)
-
 
 ## Purpose
 
@@ -1366,6 +1366,9 @@ Trying to add a batch when the method has been disabled.
 
 ### API method: `get_batch_job_result`
 
+*Deprecated. To be removed with release v2025.2* Replaced by
+[API method: batch_status](#api-method-batch_status).
+
 Return all [*test id*][Test id] objects of a *batch test*, with the number of finished *test*.
 
 Example request:
@@ -1375,7 +1378,7 @@ Example request:
 {
     "jsonrpc": "2.0",
     "id": 147559211994909,
-    "method": "batch_status",
+    "method": "get_batch_job_result",
     "params": {
         "batch_id": "8"
     }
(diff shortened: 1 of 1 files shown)


mattias-p:
Looks great. It would be nice if we could also use the terminology from https://github.com/zonemaster/zonemaster-backend/issues/960 to the extent that it's not too much trouble.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1436,6 +1439,206 @@ If the `batch_id` is undefined the following error is returned:
...
+### API method: `batch_status`
+
+This method replaces deprecated method
+[API method: get_batch_job_result](#api-method-get_batch_job_result).
+
+Returns the number of created, running and finished *tests*. Optionally it also
+returns [*test id*][Test id] objects (created, running or finished) of a
+*batch test*.
+
+Example valid request:
```
mattias-p, line 1452: I have two suggestions that I believe could improve this list of examples.
1. Condense it to one minimal example and one more comprehensive example. This method is quite simple.
2. Show example responses right after their corresponding example requests. That way we get the explanation for the circumstance of the response for free.

I expect it would take slightly more effort to construct the examples this way, but we'd bring the information density up, and I believe the reading experience would be more rewarding.

mattias-p (approved):
I had a closer look at this. Nothing is wrong but I have a few suggestions on how it could be improved. Ultimately I think this is update fills an important gap in the API.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1436,6 +1439,206 @@ If the `batch_id` is undefined the following error is returned:
...
+
+* `"created"`: a [*non-negative integer*][Non-negative integer]. The number of
+  *created* tests ([*progress*][Progress percentage] is equal to 0).
+* `"running"`: a [*non-negative integer*][Non-negative integer]. The number of
+  *running* tests ([*progress*][Progress percentage] is in the interval 1 to 99,
+  inclusive).
+* `"finished"`: a [*non-negative integer*][Non-negative integer]. The number of
+  *finished* tests ([*progress*][Progress percentage] is equal to 100).
+* `"created_test_ids"`: a list of [*test ids*][Test id] (only if requested).
+  The set of *created* tests in this *batch*.
```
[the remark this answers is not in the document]
matsduf: I agree that "waiting" is better, and I suggest that you drop "created" because all tests could be said to be created, even the completed ones.

docs/public/using/backend/rpcapi-reference.md
matsduf: I think the examples are important, but they could be moved to an appendix at the end.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1436,6 +1439,206 @@ If the `batch_id` is undefined the following error is returned:
...
 
+
+### API method: `batch_status`
+
+This method replaces deprecated method
+[API method: get_batch_job_result](#api-method-get_batch_job_result).
+
+Returns the number of created, running and finished *tests*. Optionally it also
+returns [*test id*][Test id] objects (created, running or finished) of a
+*batch test*.
```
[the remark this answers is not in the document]
matsduf: I actually use the same language as in `get_batch_job_result`.

docs/public/using/backend/rpcapi-reference.md
matsduf: Having matching query and response together is a good suggestion.

matsduf: Fixed.

docs/public/using/backend/rpcapi-reference.md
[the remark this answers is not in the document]
matsduf: Fixed.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1436,6 +1439,151 @@ If the `batch_id` is undefined the following error is returned:
...
+      "running": 2
+   }
+}
+```
+
+#### `"params"`
+
+An object with the property:
+
+* `"batch_id"`: A [*batch id*][Batch id], required.
```
MichaelTimbert, line 1540: What is the type of the "batch_id" parameter ? String or Integer ?

mattias-p: It's specified as a number in the "Batch id" data type specification (higher up in the document), but in the examples they're shown as strings. This discrepancy should probably be fixed.

matsduf: I have just followed how "batch_id" is shown in other examples.

marc-vanderwal: After some digging around, it turns out that Zonemaster::Backend::Validator allows batch_ids that are either plain non-negative numbers or strings containing non-negative numbers. However, it will correctly reject:
 - negative numbers or strings that parse to a negative number;
 - booleans, nulls and objects.

Be aware that JSON numbers map to double-precision floats: some integers cannot be appropriately represented as a JSON number type. It’s therefore safer, and I think perfectly appropriate, to use number-strings in this interface.

My suggestion: document `batch_id` better and state clearly that both numbers and number-strings are allowed.

marc-vanderwal (approved):
Looks good.

~~I suggest that we address that comment about the type of the `batch_id` parameter in a different PR. I’ll handle that.~~

Never mind, the documentation already states: “When a method expects an integer arguments (*sic*), numbers encoded in strings are also accepted and used transparently, and numbers with fractions are rounded to the nearest integer.”

mattias-p (dismissed):
This looks good to me.

It seems I forgot to mention a couple of minor nits in my last review, so I'm adding them here. Sorry about that. But I don't want to stall this PR so I'm approving anyways.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -38,15 +38,15 @@
...
   * [API method: test_progress](#api-method-test_progress)
   * [API method: get_test_results](#api-method-get_test_results)
+  * [API method: get_test_params](#api-method-get_test_params)
   * [API method: get_test_history](#api-method-get_test_history)
     * [Undelegated and delegated](#undelegated-and-delegated)
   * [API method: add_api_user](#api-method-add_api_user)
   * [API method: add_batch_job](#api-method-add_batch_job)
   * [API method: get_batch_job_result](#api-method-get_batch_job_result)
-  * [API method: get_test_params](#api-method-get_test_params)
+  * [API method: batch_status](#api-method-batch_status)
```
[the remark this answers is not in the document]
matsduf: I decided to reuse the already defined name batch_status. It is there, but experimental. I felt that we had already decided on those names.

At the end of the current version of the document you find
```
## Experimental API methods

There are also some experimental API methods documented only by name:

* system_versions
* conf_profiles
* conf_languages
* lookup_address_records
* lookup_delegation_data
* job_create
* job_status
* job_results
* job_params
* domain_history
* user_create
* batch_create
* batch_status
```

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1436,6 +1439,151 @@ If the `batch_id` is undefined the following error is returned:
...
+An object with the following properties:
+
+* `"waiting"`: a [*non-negative integer*][Non-negative integer]. The number of
+  *waiting* tests ([*progress*][Progress percentage] is equal to 0).
+* `"running"`: a [*non-negative integer*][Non-negative integer]. The number of
+  *running* tests ([*progress*][Progress percentage] is in the interval 1 to 99,
+  inclusive).
+* `"finished"`: a [*non-negative integer*][Non-negative integer]. The number of
+  *finished* tests ([*progress*][Progress percentage] is equal to 100).
+* `"waiting_test_ids"`: a list of [*test ids*][Test id] (only if requested).
```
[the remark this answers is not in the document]
matsduf: See update.

mattias-p (dismissed):

marc-vanderwal (dismissed):

matsduf:
* Clarifies JSON examples
* Corrects that empty lists of test IDs are never included in
  a response.
* Clarifies that empty lists of test IDs are never included in
  responses.
* No change of logic.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1436,6 +1439,149 @@ If the `batch_id` is undefined the following error is returned:
...
+{
+   "jsonrpc": "2.0",
+   "id": 147559211994909,
+   "result": {
+      "finished_count": 0,
+      "running_count": 5,
+      "waiting_count": 195
+   }
+}
+```
```
marc-vanderwal, line 1508: From what I understand, this is a response to a request where `list_finished_tests` is `true`, `list_waiting_tests` is `false` and `list_running_tests` is `false`.

I’d expect to always have a `finished_tests` key in the response if and only if `list_finished_tests` is `true`, even if the list is empty.

marc-vanderwal (approved):

matsduf:
@mattias-p, please re-review.

mattias-p (approved):

PR #1320, opened 2024-11-25
Updates translation instructions

matsduf:
## Purpose

The main update is to make sure that the argument check tool is used for all repositories since arguments are used in the msgids in not only Zonemaster-Engine.

## How to test this PR

Review.

### DIFF
--- a/docs/internal/maintenance/Instructions-for-translators.md
+++ b/docs/internal/maintenance/Instructions-for-translators.md
@@ -121,7 +121,7 @@ Zonemaster-GUI see [Zonemaster GUI translation guide].)
 Here it is assumed that the three repositories are cloned from the same
 directory, i.e. all three clones are sub-directories to the same directory.
 
-* If you use "old" clones, instead of creating them afreash, then you must do the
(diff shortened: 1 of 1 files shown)


docs/internal/maintenance/Instructions-for-translators.md
```diff
@@ -114,29 +114,53 @@ handled as a database instead of as a plain file.
...
-Zonemaster-Engine as an example. You do the exact same steps for the other
-two repositories.
+"[Cloning a repository]"). You need that for all repositories with messages to
+translate -- Zonemaster-Engine, Zonemaster-CLI and Zonemaster-Backend. (For
+Zonemaster-GUI see [Zonemaster GUI translation guide].)
+
+Here it is assumed that the three repositories are cloned from the same
+directory, i.e. all three clones are sub-directories to the same directory.
+
+* If you use "old" clones, instead of creating them afreash, then you must do the
```
marc-vanderwal, line 124:
```suggestion
* If you use "old" clones, instead of creating them afresh, then you must do the
```

matsduf:
@marc-vanderwal, please re-review.

PR #1352, opened 2025-02-20
DS03_ILLEGAL_ITERATION_VALUE should be WARNING and not ERROR (DNSSEC03)

jtalir:
## Purpose

Even though RFC 9276 says "Validating resolvers MAY also return a SERVFAIL response when processing NSEC3 records with iterations larger than 0", we are not aware about any resolver that would be so strict at the moment. For now and apparently for not short time in the future non zero iterations will not cause immediate operational problem. Reporting non zero iterations as ERROR in DNS zone administration is too strict in this situation.

## Context

https://datatracker.ietf.org/doc/html/rfc9276#name-recommendation-for-validati


matsduf (approved):

marc-vanderwal (approved):
I agree: if such a misconfiguration isn’t an immediate threat to the zone’s ability to be resolved, then ERROR is too harsh.

matsduf:
Implementation in zonemaster/zonemaster-engine/pull/1452

PR #1406, opened 2025-07-11
Updates the authorization model for batches and adds support for non-authorized, small batches

matsduf:
## Purpose

This PR, firstly, moves away from authorization of batch creation by using username/password created through RPC API to using a token configured in the configuration file. Secondly, is supports the creation of small batches without authorization. In the latter case using undelegated data etc will not be supported.

When small batches without authorization can be created, that function is to be added to GUI.

* Updates configuration/backend.md:
  * Updates table of contents
  * Deprecates "enable_add_batch_job"
  * Deprecates "enable_add_api_user"
  * Redefines "enable_batch_create" to replace "enable_add_batch_job"
  * Defines "enable_batch_create" and "enable_add_batch_job" to be conflicting.
  * Removes experimental "enable_user_create"
  * Defines "batch_api_key" for a token to authorize batch creation.
  * Defines "max_batch_size_non_auth" to set to number of domains in batches created without batch_api_key.
  * Editorial updates.
* Updates rpcapi-reference.md
  * Updates table of contents
  * Defines data type "Batch API key"
  * Defines method "conf_max_batch_size_non_auth"
  * Deprecates methods "add_api_user" and "add_batch_job"
  * Redefines method "batch_create" to replace "add_batch_job"
  * Removes experimental "user_create"
  * Editorial updates
  * Deprecates "batch id" and introduces "hash batch id" which is the same kind of "hashed ID" as the test ID.
* Removal of deprecated parts:
  * Parts deprecated by this PR are to be removed by release v2026.2.

When the changes in this PR are approved the following changes are to be created:

* Updated docs/public/using/backend/Using-Zonemaster-Backend-for-batch-testing.md
* Implement the updates in Backend.

## Context

* https://github.com/zonemaster/zonemaster-backend/issues/1223 (batch authorization model)
* https://github.com/zonemaster/zonemaster-backend/issues/1154 ("hash format" on batch ID)
* https://github.com/zonemaster/zonemaster-backend/issues/1153 (batch creation without token)

## How to test this PR

Review the changes.


docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1539,24 +1791,27 @@ for all three status values and none was nil:
...
+* `"batch_id"`: A [*batch id*][Batch id] or a *[batch id type 2]*, required.
 * `"list_waiting_tests"`: a boolean, optional (default false). If true include
   the `"waiting_tests"` property in the result object.
 * `"list_running_tests"`: a boolean, optional (default false). If true include
   the `"running_tests"` property in the result object.
 * `"list_finished_tests"`: a boolean, optional (default false). If true include
   the `"finished_tests"` property in the result object.
 
+The use of *[batch id]* is deprecated from v2026.1. Use *[batch id type 2]*
+instead.
```
marc-vanderwal: That wording seems a tad ambiguous to me. How about:
```suggestion
The use of *[batch id]* is deprecated and will be removed by release v2026.1.
Use *[batch id type 2]* instead.
```

marc-vanderwal (dismissed):

docs/public/configuration/backend.md
```diff
@@ -59,31 +60,54 @@ In addition to the configuration file, some settings can configured using
...
+`enable_batch_create`, `batch_create_token`, `batch_create_max_size_non_auth`.
 
 ### enable_add_batch_job
 
-Boolean value to enable the `add_batch_job` and `batch_create` methods of the API.
+*Deprecated. To be removed by Zonemaster version v2026.1. Use
+`enable_batch_create` instead.*
+
+Boolean value to enable the `add_batch_job` and `batch_create` methods of the
+API. May not co-exist with [RPCAPI.enable_batch_create].
```
mattias-p:
```suggestion
API. Must not co-exist with [RPCAPI.enable_batch_create].
```
There are more of these.

docs/public/configuration/backend.md
```diff
@@ -59,31 +60,54 @@ In addition to the configuration file, some settings can configured using
...
 ### enable_batch_create
 
-An experimental alias for [enable_add_batch_job][RPCAPI.enable_add_batch_job].
+*Replaces deprecated `enable_add_batch_job`.*
+
+Boolean value to enable the `add_batch_job` and `batch_create` methods of the
+API. May not co-exist with [RPCAPI.enable_add_batch_job].
+
+Accepted values: `yes` (or `true`) or `no` (or `false`),
+default to `yes` (enabled).
```
mattias-p: IMHO we should pick either `yes` or `true` and avoid aliasing them.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -253,7 +278,8 @@ Properties:
 
 Basic data type: number (integer)
 
-A non-negative integer is either zero or strictly positive.
+A non-negative integer is either zero or strictly positive. In this specification
+the integer is decimal, unless specified otherwise.
```
mattias-p: I don't understand what this is trying to express. In JSON, aren't all numbers decimal? If we want to deviate from the definition of the data type, shouldn't we define a new data type?

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -172,6 +191,12 @@ Represents the version of the client.
 Used for monitoring which client (GUI) uses the API.
 
 
+### Configuration file
+
+The configuration of the Zonemaster Backend *RPC API daemon* is done as
+specified in the [Configuration] document.
```
mattias-p: This looks rather out of place in the Data types section.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -410,6 +446,42 @@ An object with the following properties:
 > TODO: List all possible error codes and describe what they mean enough for clients to know how react to them.
 >
 
+### API method: `conf_batch_create_max_size_non_auth`
+
+Returns the maximum number of elements that the `"domains"` key may containt when
+method `batch_create` is requested without `token`.
+
+Key `"result"` in the response is set to a *[non-negative integer]*.
```
mattias-p: The `"result"` value should be an object to make it easier to change the API in the future.

matsduf:
Renames some methods and keys. No change of logic.

docs/public/configuration/backend.md
```diff
@@ -59,31 +60,54 @@ In addition to the configuration file, some settings can configured using
...
 
-Available keys: `enable_batch_create`, `enable_user_create`,
-`enable_add_batch_job`, `enable_add_api_user`.
+Available keys: `batch_api_key`, `enable_add_batch_job`, `enable_add_api_user`,
+`enable_batch_create`, `max_batch_size_non_auth`.
+
+### batch_api_key
+
+String to use in `batch_create` method call to authorize the creation of a
+batch.
```
mattias-p:
```suggestion
API key to use in requests to be granted `trusted` privilege level. E.g., to allow creation of larger batches using the `batch_create` method.
```

docs/public/configuration/backend.md
```diff
@@ -59,31 +60,54 @@ In addition to the configuration file, some settings can configured using
...
+
+A string of alphanumerics of at least 1 and at most 80 characters. I.e. a string
+matching `/^[a-zA-Z0-9]{1,80}$/`.
```
mattias-p: We should probably reject low-entropy API keys. Best practices guidance seems to recommend using at least 128 bits of entropy, and more than 256 bits seems excessive. If we want to hash our API keys in the future, we probably want to use a 256 bit hash algorithm, so having longer API than that doesn't give any additional security.

```suggestion
A string of alphanumerics of at least 1 and at most 80 characters. I.e. a string
matching `/^[a-zA-Z0-9]{22,43}$/`.
```

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -98,25 +105,22 @@ see the [architecture documentation].
...
 ## Privilege levels
 
 This API provides three classes of methods:
 
 * *Unrestricted* methods are available to anyone with access to the API.
-* *Authenticated* methods have parameters for [*username*][Username] and [*api key*][API key]
-  credentials.
+* *Authenticated* methods have either
+  * Parameters for [*username*][Username] and [*api key*][API key] credentials, or
+  * Parameter for [*batch api key*][Batch API Key] credential.
```
mattias-p, line 123: I believe we're describing privilege levels the wrong way. We should categorize RPC calls instead of RPC methods. Different calls to the same method could be granted different privilege levels.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -410,6 +450,42 @@ An object with the following properties:
...
+Key `"result"` is set to zero (0) in the response if any of the following is
+true:
+* [RPCAPI.enable_add_batch_job] is set to false,
+* [RPCAPI.enable_batch_create] is set to false,
+* [RPCAPI.max_batch_size_non_auth] is set to 0.
+
+Else, `"result"` is set to the value of
+[RPCAPI.max_batch_size_non_auth], if that is set.
+
+Else, `"result"` is set to 5.
```
mattias-p: Instead of restating the default value here, we should probably just rely on the hyperlink and let the configuration documentation be the source of truth.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -410,6 +450,42 @@ An object with the following properties:
...
+Returns the maximum number of elements that the `"domains"` key may contain when
+method `batch_create` is requested without `batch_api_key`.
+
+Key `"result"` in the response is set to a *[non-negative integer]*.
+
+Key `"result"` is set to zero (0) in the response if any of the following is
+true:
+* [RPCAPI.enable_add_batch_job] is set to false,
+* [RPCAPI.enable_batch_create] is set to false,
+* [RPCAPI.max_batch_size_non_auth] is set to 0.
```
mattias-p: The logic for zeroing the return value if batch creation is disabled had me scratching my head at first. But I guess it's to signal that anonymous batch creation is disabled if the entire batch creation feature is disabled? That makes sense to me, but maybe it would be helpful to spell that out in the documentation?

Or, I have another idea. I haven't thought a great deal about this, but how about if we instead had a more general method for querying? It could return a subset of the configuration that should be visible to clients? E.g., whether the batch feature is enabled at all, the max size for unauthenticated batch creation, available languages, etc, in a single JSON structure.

In any case, the "result" member in the response should not be a scalar, but rather an object. We previously changed all scalar "result" members into object in order to make it possible to update our APIs as "additions" rather than "breaking changes" per semver terminology. In this vain I suggest that we have `"result":{"value":5}` or `"result":{"max_batch_size_non_auth":5}` or something similar.

docs/public/configuration/backend.md
```diff
@@ -59,31 +60,54 @@ In addition to the configuration file, some settings can configured using
 
 ## RPCAPI section
 
-Available keys: `enable_batch_create`, `enable_user_create`,
-`enable_add_batch_job`, `enable_add_api_user`.
+Available keys: `batch_api_key`, `enable_add_batch_job`, `enable_add_api_user`,
+`enable_batch_create`, `max_batch_size_non_auth`.
+
+### batch_api_key
```
mattias-p, line 69: I'm thinking we should have a unified authorization model for the entire RPC-API where RPC calls are either "authenticated" or "unrestricted"* depending on whether a recognized API key was provided. From this perspective I think it would be better to not have "batch" in the name of the API key option.

On a related note, we should have a note somewhere that API keys should only be used over secure channels.

*) IMHO, "unrestricted" is an unfortunate name. We should probably rename it to "public" or "anonymous" or something.

docs/public/configuration/backend.md
```diff
@@ -59,31 +60,54 @@ In addition to the configuration file, some settings can configured using
...
+default to `yes` (enabled).
 
-### enable_user_create
+### max_batch_size_non_auth
 
-An experimental alias for [enable_add_api_user][RPCAPI.enable_add_api_user].
+Maximal number of domain names in a batch created without batch_api_key. If set
+to 0, batches can only be created with valid batch_api_key included.
 
+* Acceptable value: Non-negative decimal integer
```
mattias-p: What is the point of the word "decimal" here?

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -137,16 +141,45 @@ I.e. a string matching `/^[a-zA-Z0-9-_]{1,512}$/`.
 
 Represents the password of an authenticated account (see *[Privilege levels]*)
 
+Note that [*Batch API key*][Batch API Key] is another data type.
+
+
+### Batch Api Key
```
mattias-p: As I remarked on the `RPCAPI.batch_api_key` option, the use of API keys should probably not be limited to the batch feature. In order to avoid future API churn, it would be better to expand the existing `API key` data type to accept the new API key format in addition to the old password format. And we should include a clarification that the data type is used for both API keys and passwords, but that the usage and format for passwords is deprecated.

Also, "API key" should probably be spelled with uppercase API and lowercase key.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -137,16 +141,45 @@ I.e. a string matching `/^[a-zA-Z0-9-_]{1,512}$/`.
...
+
 Basic data type: number
 
-A strictly positive integer.
+A strictly positive integer that the client can treat as a string.
 
 The unique id of a *batch*.
 
 
+### Hash batch id
```
mattias-p, line 177: IMHO, the "hash" in "Hash batch id" is unfortunate. The batch id is just supposed to be a unique identifier, right? It's not supposed to be a hash of anything in particular. The fact that we're currently employing a hash function in the generation of test ids should not be reflected in the name of the batch id data type name.

It's better to extend the current `batch id` data type to accept the new hex string format. And the deprecation should apply to the integer format, not to the entire data type.

docs/public/configuration/backend.md
```diff
@@ -59,31 +60,54 @@ In addition to the configuration file, some settings can configured using
 
 ## RPCAPI section
 
-Available keys: `enable_batch_create`, `enable_user_create`,
-`enable_add_batch_job`, `enable_add_api_user`.
+Available keys: `batch_api_key`, `enable_add_batch_job`, `enable_add_api_user`,
+`enable_batch_create`, `max_batch_size_non_auth`.
```
mattias-p: Do we really need to have a paragraph listing these? There's already a table of contents and their names are also clearly highlighted in the sub-section headings below.

docs/public/configuration/backend.md
```diff
@@ -59,31 +60,54 @@ In addition to the configuration file, some settings can configured using
...
+*Replaces deprecated `enable_add_batch_job`.*
+
+Boolean value to enable the `add_batch_job` and `batch_create` methods of the
+API. May not co-exist with [RPCAPI.enable_add_batch_job].
+
+Accepted values: `yes` (or `true`) or `no` (or `false`),
+default to `yes` (enabled).
 
-### enable_user_create
+### max_batch_size_non_auth
```
mattias-p: Instead of having a negation in the name, I think it would be better to refer to the privilege level it does apply to. I.e., "unrestricted" according to the current naming, although I have a comment in this review to rename it.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1366,6 +1447,180 @@ Trying to add a batch when the method has been disabled.
 ```
 
 
+
+### API method: `batch_create`
+
+This method replaces deprecated method
+[API method: add_batch_job](#api-method-add_batch_job)
+
+Add a new *batch test* composed by a set of [*domain name*][Domain name] and a *params* object.
```
mattias-p, line 1639: We should make up our minds on whether the data model entity is called "batch" or "batch test". Personally I vote for "batch".

Is it really a correct description of the data model to state that a batch has a "params" object? Isn't it the individual tests that have "params" objects?

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1366,6 +1447,180 @@ Trying to add a batch when the method has been disabled.
...
+by default).
+
+A *batch_api_key* can be added to the [configuration file] as
+[RPCAPI.batch_api_key].
+
+*Tests* enqueued using this method are assigned a [*priority*][Priority] of 5.
+
+> In previous versions of Zonemaster-Backend a new batch could not be created by
+> the same [*username*][Username] if that *username* had created a batch that was
+> not yet finished. That restriction has been removed in version 2023.2.
```
mattias-p: Isn't this a good time to drop this historical comment from the API reference?

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1366,6 +1447,180 @@ Trying to add a batch when the method has been disabled.
...
+
+
+#### `"params"`
+
+An object with the following properties:
+
+* `"batch_api_key"`: A [*batch api key*][batch api key], see details on `"batch_api_key"` below.
+* `"domains"`: A list of [*domain names*][Domain name], required. The domains to be tested.
+* `"test_params"`: As described below, optional, (default: `{}`). The `"test_params"` key
+  must be absent or empty (`{}`) unless `"batch_api_key"` is included.
```
mattias-p: For consistency?
```suggestion
  must be absent or empty (`{}`) unless `"batch_api_key"` is present.
```

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1366,6 +1447,180 @@ Trying to add a batch when the method has been disabled.
...
+* `"ds_info"`: A list of [*DS info*][DS info] objects, optional. (default: `[]`)
+* `"ipv4"`: A boolean, optional. (default: [`net.ipv4`][net.ipv4] profile value).
+* `"ipv6"`: A boolean, optional. (default: [`net.ipv6`][net.ipv6] profile value).
+* `"priority"`: A [*priority*][Priority], optional. (default: `5`)
+* `"queue"`: A [*queue*][Queue], optional. (default: `0`).
+
+
+#### `"result"`
+
+*id* holds a [*hash batch id*][Hash batch id].
```
mattias-p: * We should maintain the format of the result sections.
* The `get_batch_job_result` method names its batch id parameter `batch_id`, so I think it's appropriate that we use the same name here. Also, it's nice to keep distinct names for test id and batch id fields.

```suggestion
An object with the following properties:

* `"batch_id"`: A [*hash batch id*][Hash batch id].
```

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1382,7 +1637,7 @@ Example request:
     "id": 147559211994909,
     "method": "get_batch_job_result",
     "params": {
-        "batch_id": "8"
+        "batch_id: "8"
```
mattias-p:
```suggestion
        "batch_id": "8"
```

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1539,24 +1794,27 @@ for all three status values and none was nil:
...
+* `"batch_id"`: A [*batch id*][Batch id] or a *[hash batch id]*, required.
 * `"list_waiting_tests"`: a boolean, optional (default false). If true include
   the `"waiting_tests"` property in the result object.
 * `"list_running_tests"`: a boolean, optional (default false). If true include
   the `"running_tests"` property in the result object.
 * `"list_finished_tests"`: a boolean, optional (default false). If true include
   the `"finished_tests"` property in the result object.
 
+The use of *[batch id]* is deprecated and will be removed by release v2026.1.
+Use *[hash batch]* instead.
```
mattias-p:
```suggestion
The use of *[batch id]* is deprecated and will be removed by release v2026.1.
Use *[hash batch id]* instead.
```

Also, we should probably provide a migration path for users who keep references to old batch ids? E.g., this could involve a script to export a translation table from the integer format to the hex string format. This script could be deprecated from its introduction and planned to be dropped at the same as the old batch id format.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -137,16 +141,45 @@ I.e. a string matching `/^[a-zA-Z0-9-_]{1,512}$/`.
 
 Represents the password of an authenticated account (see *[Privilege levels]*)
 
+Note that [*Batch API key*][Batch API Key] is another data type.
+
+
+### Batch Api Key
+
+Basic data type: string
```
mattias-p, line 155: We should probably accept a list of API keys here instead of just a single one. This would improve the UX of the use case to roll API keys. The RPCAPI could accept both the old and the new API key during a transition period.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -137,16 +141,45 @@ I.e. a string matching `/^[a-zA-Z0-9-_]{1,512}$/`.
...
 
 ### Batch id
 
+*Deprecated. To be removed with release v2026.1. To be replaced by
+[Hash batch id].*
+
 Basic data type: number
 
-A strictly positive integer.
+A strictly positive integer that the client can treat as a string.
```
mattias-p: I don't get this. The basic data type is number. Are you suggesting that we should also accept strings containing numbers? I don't see the point of that.

docs/public/configuration/backend.md
matsduf: Such a change is out of scope of this PR. I just keep it as is.

docs/public/configuration/backend.md
matsduf: I do not see why we should force long key. We could suggest.

docs/public/configuration/backend.md
matsduf: To clarify that is should not be octal or hexadecimal.

docs/public/using/backend/rpcapi-reference.md
matsduf: It sounds like a comment beyond this PR. Please clarify.

docs/public/using/backend/rpcapi-reference.md
matsduf: We do not have any other use today. Keeping the name is more confusing. If there is another use in the future, "API key" ("api_key") could be reintroduced.

docs/public/using/backend/rpcapi-reference.md
matsduf: Here it should not be a list. In the API method call you are not expected to use more than one. Your proposal is, if I understand it correctly, that the configuration file should accept a list keys of type batch_api_key

docs/public/using/backend/rpcapi-reference.md
matsduf: I will look at it again.

docs/public/using/backend/rpcapi-reference.md
matsduf: I will look at it.

docs/public/using/backend/rpcapi-reference.md
matsduf: I will look at it.

docs/public/using/backend/rpcapi-reference.md
matsduf: If all JSON numbers are decimal the clarification is not strictly needed, and should not open for non-decimal integers.

docs/public/using/backend/rpcapi-reference.md
matsduf: I will change that.

docs/public/using/backend/rpcapi-reference.md
matsduf: Combining several API methods into one is out of scope of this PR. The purpose of this API method is to make it possible for GUI to find out if it can create unauthenticated batches, and how large they can be. I am not sure if it should be exposed to the world of batches are allowed or not.

docs/public/configuration/backend.md
matsduf: I have used the same model as the current keys enable_add_batch_job and enable_add_api_user one and two steps up. I think this should have the same model as the other.

docs/public/using/backend/rpcapi-reference.md
matsduf: I will look at it.

docs/public/using/backend/rpcapi-reference.md
matsduf: There is one params object, or what do you mean?

docs/public/using/backend/rpcapi-reference.md
matsduf: Yes.

docs/public/using/backend/rpcapi-reference.md
matsduf: Good suggestion.

docs/public/using/backend/rpcapi-reference.md
mattias-p: Yes, you're right. I made the comment in the wrong place.

docs/public/configuration/backend.md
mattias-p: Yeah, I realize that. But this option is being stabilized, so this is the last chance to clean this up without breaking the API. And both of the other two boolean options are being deprecated, so we don't need to keep those aligned with the conventions of the rest of the format.

docs/public/using/backend/rpcapi-reference.md
mattias-p: Yes, they are strictly decimal: https://www.rfc-editor.org/rfc/rfc8259#section-6

docs/public/configuration/backend.md
mattias-p: I see this as future-proofing. Today we're only using the API keys for limiting batch sizes, which can be worked around by anonymous users by creating unlimited single jobs. But in the future we may reuse this authentication mechanism for other features that are more security sensitive. At that point we do want to restrict low-entropy API keys, but then it would be a breaking change. People who would be tempted to set a 6-bit API key should probably consider setting `max_batch_size_non_auth = 1000000` instead.

Allowing longer API keys than the 43 character that I suggested is probably a good idea though. I was only thinking about entropy, but it's also common practice to have API keys that include credentials that don't contribute towards entropy. So the upper limit should be at least 64 characters, but 80 should be just fine.

docs/public/using/backend/rpcapi-reference.md
mattias-p: You could definitely argue that this is out of scope for this PR. But I'm not sure what to clarify.

docs/public/using/backend/rpcapi-reference.md
mattias-p: For PRs that stabilize APIs I'd say scrutinizing/validating the API design is within scope. I haven't formed a strong opinion either way, but I do think this is the right time to find a good design to settle on.

docs/public/configuration/backend.md
mattias-p: All numbers in JSON are decimal. We do not have to re-specify that.

docs/public/using/backend/rpcapi-reference.md
mattias-p: It's the current situation that is confusing. The problem is that we're calling our passwords "API keys". Having passwords called API keys and API keys called batch API keys is no better. Having an API key data type that in some deprecated cases double as passwords is no worse. And it's a good thing if we can avoid breaking changes in the future.

docs/public/configuration/backend.md
mattias-p: Okay, I can make a PR to remove it.

docs/public/configuration/backend.md
matsduf: I have updated.

docs/public/configuration/backend.md
matsduf: Updated.

docs/public/configuration/backend.md
matsduf: Updated.

docs/public/configuration/backend.md
matsduf: Updated.

docs/public/using/backend/rpcapi-reference.md
matsduf: I do not understand you review comment.

docs/public/using/backend/rpcapi-reference.md
matsduf: Updated.

docs/public/using/backend/rpcapi-reference.md
matsduf: Renaming the data format in the future will not require deprecation. I think it is clearer to have two formats.

docs/public/using/backend/rpcapi-reference.md
matsduf: Moved.

docs/public/using/backend/rpcapi-reference.md
matsduf: Updated.

docs/public/using/backend/rpcapi-reference.md
matsduf: Updated.

docs/public/using/backend/rpcapi-reference.md
matsduf: Updated.

docs/public/using/backend/rpcapi-reference.md
matsduf: Updated.

docs/public/using/backend/rpcapi-reference.md
matsduf: Updated.

docs/public/configuration/backend.md
matsduf: Updated.

docs/public/using/backend/rpcapi-reference.md
mattias-p: It's still there. Can you check again?

docs/public/configuration/backend.md
mattias-p: I can't see the update. Can you check again?

matsduf: Renamed "non-auth" to "anonymous".

docs/public/using/backend/rpcapi-reference.md
mattias-p: I can't find any updates related to the migration path. Maybe we should create an issue to track that?

docs/public/using/backend/rpcapi-reference.md
matsduf: There were two instances of the same text. I removed one of them.

docs/public/configuration/backend.md
```diff
@@ -59,31 +60,57 @@ In addition to the configuration file, some settings can configured using
...
+
+API key to use in requests to be granted `trusted` privilege level. E.g., to
+allow creation of larger batches using the `batch_create` method.
+
+A string of alphanumerics of at least 10 and at most 64 characters. I.e. a string
+matching `/^[a-zA-Z0-9]{10,64}$/`. It is recommended to use an `batch_api_key` of
+at least 22 characters.
+
+Multiple `batch_api_key` is supported. All configured keys will be equal in
+authorization.
```
mattias-p: The syntax for multiple API keys should be specified.  For `LANGUAGE.locale` we're using a space separated list. That should work here too.

matsduf:
@mattias-p and @marc-vanderwal, can you please re-review?

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -141,20 +147,34 @@ I.e. a string matching `/^[a-zA-Z0-9-_]{1,512}$/`.
 
 Represents the password of an authenticated account (see *[Privilege levels]*)
 
+Note that [*Batch API key*][Batch API Key] is another data type.
+
+
+### Batch Api Key
```
marc-vanderwal:
```suggestion
### Batch API Key
```

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1499,15 +1495,15 @@ Example response:
 
 An object with the following properties:
 
-* `"token"`: A [*token*][token], see details on `"token"` below.
+* `"batch_api_key"`: A [*batch api key*][batch api key], see details on `"batch_api_key"` below.
```
marc-vanderwal:
```suggestion
* `"batch_api_key"`: A [*batch API key*][Batch API key], see details on `"batch_api_key"` below.
```

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1914,7 +1911,7 @@ There are also some experimental API methods documented only by name:
 [Test id]:                                    #test-id
 [Test result]:                                #test-result
 [Timestamp]:                                  #timestamp
-[Token]:                                      #token
+[Batch Api Key]:                              #batch-api-key
```
marc-vanderwal:
```suggestion
[Batch API key]:                              #batch-api-key
```

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1525,40 +1521,41 @@ The value of `"test_params"` is an object with the following properties:
...
+* `"batch_id"`: A [*hash batch id*][Hash batch id].
 
 
 #### `"error"`
 
 If the given `profile` is not among the [available profiles][Profile sections], a
 user error is returned, see the [profile name section][profile name].
 
-Trying to add a batch with wrong [*token*][token]:
+Trying to add a batch with wrong [*batch api key*][Batch API key]:
```
marc-vanderwal:
```suggestion
Trying to add a batch with wrong [*batch API key*][Batch API key]:
```

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1525,40 +1521,41 @@ The value of `"test_params"` is an object with the following properties:
...
   "id": 1,
   "error": {
-    "message": "Invalid token for batch_create",
+    "message": "Invalid batch_api_key for batch_create",
     "code": -32603,
     "data": {
-        "message": "Invalid token",
-        "path": "/token"
+        "message": "Invalid batch api key",
+        "path": "/batch api key"
```
marc-vanderwal:
```suggestion
        "message": "Invalid batch API key",
        "path": "/batch_api_key"
```

marc-vanderwal (commented):
I found some typos.

docs/public/configuration/backend.md
matsduf: Fixed.

docs/public/configuration/backend.md
```diff
@@ -59,36 +60,58 @@ In addition to the configuration file, some settings can configured using
...
 default to `no` (disabled).
 
 ### enable_batch_create
 
-An experimental alias for [enable_add_batch_job][RPCAPI.enable_add_batch_job].
+*Replaces deprecated `enable_add_batch_job`.*
 
-### enable_user_create
+Boolean value to enable the `add_batch_job` and `batch_create` methods of the
+API. May not co-exist with [RPCAPI.enable_add_batch_job].
```
mattias-p:
```suggestion
API. Mutually exclusive with [RPCAPI.enable_add_batch_job].
```

mattias-p:
```suggestion
API. Must not co-exist with [RPCAPI.enable_add_batch_job].
```

docs/public/configuration/backend.md
```diff
@@ -59,36 +60,58 @@ In addition to the configuration file, some settings can configured using
 
 ## RPCAPI section
 
-Available keys: `enable_batch_create`, `enable_user_create`,
-`enable_add_batch_job`, `enable_add_api_user`.
+
+### batch_api_key
```
mattias-p: We should remove "batch" from the name of this data type. That way we avoid API churn when find a new use for the authorization mechanism.
```suggestion
### api_key
```

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -61,8 +66,10 @@ This API is implemented using [JSON-RPC 2.0].
 JSON-RPC request objects are accepted in the body of HTTP POST requests to any path.
 The HTTP request must contain the header `Content-Type: application/json`.
 
-All JSON-RPC request and response objects have the keys `"jsonrpc"`, `"id"` and `"method"`.
-For details on these, refer to the [JSON-RPC 2.0] specification.
+All JSON-RPC request objects have the keys `"jsonrpc"`, `"id"` and `"method"`.
+The response object have the same keys `"jsonrpc"` and `"id"`. For details on
+these, refer to the [JSON-RPC 2.0] specification. The `"jsonrpc"` and `"id"` keys
+are assumed in the specification of the methods below.
```
mattias-p: Do you know what purpose this paragraph serves? AFAICT it only restates things from the specification. It's better to just refer to the specification. IMHO this entire paragraph could be removed.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -98,29 +105,32 @@ see the [architecture documentation].
...
-For more information on the validation error data format see
-[Validation error data].
-
-All error states that occur after the RPC method has been identified are reported as internal errors with code `-32603`.
+Error code | Description
+:----------|:-------------------------------------------------------------------
+-32700     | The request object is invalid JSON.
+-32601     | No method is specified or an invalid method is specified.
+-32602     | No `params` object is specified when required, or the `params` object is invalid. For more information on the validation error data format see [Validation error data].
+-32603     | All error states that occur after the RPC method has been identified are reported as internal errors.
```
mattias-p:
```suggestion
-32603     | An internal error occurred.
```

docs/public/configuration/backend.md
```diff
@@ -59,36 +60,58 @@ In addition to the configuration file, some settings can configured using
...
+API key to use in requests to be granted `trusted` privilege level. E.g., to
+allow creation of larger batches using the `batch_create` method.
+
+A string of alphanumerics of at least 10 and at most 64 characters. I.e. a string
+matching `/^[a-zA-Z0-9]{10,64}$/`. It is recommended to use an `batch_api_key` of
+at least 22 characters.
+
+Multiple `batch_api_key` is supported. All configured keys will be equal in
+authorization. For each key `batch_api_key` is repeated on a new line in the
+configuration file with the "Batch API Key" as value.
```
marc-vanderwal: So does that mean that something like this will work?

```ini
[RPCAPI]
batch_api_key = "NotSoSecret"
batch_api_key = "NotSoSecretEither"
batch_api_key = "LikewiseNotASecret"
```

docs/public/configuration/backend.md
```diff
@@ -108,7 +131,7 @@ SQLite            | `SQLite`
 
 ### polling_interval
 
-A strictly positive decimal number. Max 5 and 3 digits in the integer and fraction
+A strictly positive integer. Max 5 and 3 digits in the integer and fraction
 components respectively.
```
marc-vanderwal: Integers cannot have a fraction component.
```suggestion
A strictly positive integer. Max 5 digits.
```

docs/public/configuration/backend.md
matsduf: Yes.

docs/public/configuration/backend.md
matsduf: A good catch. I do not remember why I changed from "decimal number" to "integer". I will change it back. We should continue to support values smaller than one second.

docs/public/configuration/backend.md
matsduf: Do think the language should be clarified? Do you think it is a good or bad idea?

marc-vanderwal: I think it’s a good idea because it helps avoid having shared secrets between more than two components.

Yes, maybe you could add an example like the one I gave in my previous comment in the text to make the syntax clearer.

matsduf: @marc-vanderwal, I like examples because it helps to understand. In that file we do not have much examples, but we have several examples in the zonemaster-backend:share/config.ini file. Do you agree that it would be better to put your examples in the ini file? It will be updated in an upcoming PR.

marc-vanderwal: I think it’s best to have this example both in the documentation and in the example configuration file. The syntax is a bit unorthodox because configuration keys are usually unique within a file. It should be crystal clear for users that yes, you can have multiple `batch_api_key` variables and that it makes all API keys work, not just the first or last one.

matsduf: Updated as suggested.

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -136,16 +146,45 @@ I.e. a string matching `/^[a-zA-Z0-9-_]{1,512}$/`.
...
 
+Note that [*Batch API key*][Batch API Key] is another data type.
+
+
+### Batch API Key
+
+Basic data type: string
+
+A string of alphanumerics of at least 1 and at most 80 characters. I.e. a string
+matching `/^[a-zA-Z0-9]{1,80}$/`.
```
MichaelTimbert: Regex mismatch  `batch_api_key` defined in backend.md defined between 10 and 64 chars (/^[a-zA-Z0-9]{10,64}$/ (10–64 chars))

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1364,6 +1442,253 @@ Trying to add a batch when the method has been disabled.
...
+[RPCAPI.batch_api_key].
+
+*Tests* enqueued using this method are assigned a [*priority*][Priority] of 5.
+
+Example request:
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 147559211348450,
+  "method": "add_batch_job",
```
MichaelTimbert:
```suggestion
  "method": "batch_create",
```

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1364,6 +1442,253 @@ Trying to add a batch when the method has been disabled.
...
+  "params" : {
+    "batch_api_key": "123abcABC",
```
MichaelTimbert:
```suggestion
    "batch_api_key": "12341337abcdefgABCDEFG",
```
make it at least 22 characters to be compliant with backend.md

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1364,6 +1442,253 @@ Trying to add a batch when the method has been disabled.
...
+  }
+}
+```
+
+
+
+
+### API method: `get_batch_job_result`
+
+*Deprecated. To be removed with release v2025.2* Replaced by
```
MichaelTimbert:
```suggestion
*Deprecated. To be removed with release v2026.2* Replaced by
```

docs/public/using/backend/rpcapi-reference.md
```diff
@@ -1364,6 +1442,253 @@ Trying to add a batch when the method has been disabled.
...
+* `"priority"`: A [*priority*][Priority], optional. (default: `5`)
+* `"queue"`: A [*queue*][Queue], optional. (default: `0`).
+
+
+#### `"result"`
+
+An object with the following properties:
+
+* `"batch_id"`: A [*hash batch id*][Hash batch id].
+
```
MichaelTimbert, line 1523: It is confusing with the example. do we get 
```
"result": "B-c45a3f8256c4a155"
```
or 
```
"result": { "batch_id": "B-c45a3f8256c4a155" }
```

docs/public/using/backend/rpcapi-reference.md
matsduf: Good spot. I changed in one file because of a comment, but not in the other. I will update it to be 10-64 characters in both.

docs/public/using/backend/rpcapi-reference.md
matsduf: Updates the specification. It should use the same model as "start_domain_test" and "add_batch_job".

docs/public/using/backend/rpcapi-reference.md
matsduf: Already removed. I have to rebase the PR.

MichaelTimbert (dismissed):
Look good for me.

marc-vanderwal (dismissed):

PR #1412, opened 2025-07-18
Updates the test case specification for DNSSEC01

matsduf:
## Purpose

* Includse explicit msgid.
* Matches current template.
* New algorithms from IANA included.
* Removes check if the installation handles the algorithm. That is already done in DNSSEC02 and can be elaborated there.
* Always outputs some message for all tested zones.

Test scenarios are created in #1413.

When (if) #1418 is merged this PR is updated accordingly. All `ns_ip_list` arguments will be replaced by `ns_list` arguments. That change will not affect the logic.

This PR requires implementation.

## How to test this PR

Review.

### DIFF
--- a/docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
+++ b/docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
@@ -229,7 +229,7 @@ queries follow, unless otherwise specified below, what is specified for
        the key tag and output [DS01_DS_ALGO_2_MISSING] with key tag and the
        extracted list of IP addresses.
 
-10. If the *Responds Without Valid DS* is empty, the *Responds With DS* sets is
+10. If the *Responds Without Valid DS* is empty, the *Responds With DS* set is
     empty and the *Ignored Parent NS IP* set is non-empty, then output
     *[DS01_NO_RESPONSE]* with the name server IP from the *Ignored Parent NS IP*
     set.


docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -20,80 +21,94 @@
...
+| 1                | SHA-1                      | DS01_DS_ALGO_DEPRECATED   |
+| 2                | SHA-256                    | DS01_DS_ALGO_OK           |
+| 3                | GOST R 34.11-94            | DS01_DS_ALGO_DEPRECATED   |
+| 4                | SHA-384                    | DS01_DS_ALGO_OK           |
+| 5                | GOST R 34.11-2012          | DS01_DS_ALGO_OK           |
+| 6                | SM3                        | DS01_DS_ALGO_OK           |
+| 7-127            | Unassigned                 | DS01_DS_ALGO_UNASSIGNED   |
+| 128-252          | Reserved                   | DS01_DS_ALGO_RESERVED     |
+| 253-254          | Reserved for Private Use   | DS01_DS_ALGO_PRIVATE      |
+| 5-255            | Unassigned                 | DS01_DS_ALGO_UNASSIGNED   |
```
marc-vanderwal:
```suggestion
| 255              | Unassigned                 | DS01_DS_ALGO_UNASSIGNED   |
```

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -20,80 +21,94 @@
...
+| DS01_DS_ALGO_UNASSIGNED  | ERROR   | ip_list, keytag, ds_algo_num                | The DS record with keytag {keytag} uses a unassigned digest algorithm {ds_algo_num} on parent name servers (IP) "{ip_list}".                                 |
+| DS01_NO_RESPONSE         | WARNING | ip_list                                     | No response or error in response from all parent name servers on the DS query. Name servers are "{ip_list}".                                                 |
+| DS01_PARENT_SERVER_NO_DS | ERROR   | ip_list                                     | The following name servers do not provide DS record or have not been properly configured. Fetched from parent name servers (IP) "{ip_list}".                 |
+| DS01_PARENT_ZONE_NO_DS   | NOTICE  | ip_list                                     | The parent zone provides no DS records for the child zone. Fetched from parent name servers (IP) "{ip_list}".                                                |
```
marc-vanderwal: The previous version of the message tags used `ns_ip_list` arguments. Wouldn’t it be better to report the NS names along with their IPs?

matsduf: MethodV2 *Get-Parent-NS-IP* only provides the IP addresses of the parent name servers, not the names. Maybe the implementation picks the names up, but that is then beyond how it is specified.

We could update the method to also find the names of the parent name servers.

marc-vanderwal: I think it’s better if we had a *Get-Parent-NS-Names-and-IPs* method. If we introduce it, then *Get-Parent-NS-IP* can be defined in terms of *Get-Parent-NS-Names-and-IPs*.

(Besides, it could be a good opportunity to fix a small inconsistency in [MethodsV2.md](https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/tests/MethodsV2.md): on the one hand we have *Get-Parent-NS-IP*, ending in *-IP*, but on the other hand there is *Get-Del-NS-IPs* and *Get-Zone-NS-IPs*. Shouldn’t it be named *Get-Parent-NS-IPs* instead?)

marc-vanderwal: I’ve created #1418, which proposes a new method, *Get-Parent-NS-Names-and-IPs*.

marc-vanderwal: Well, actually, when digging in Zonemaster-Engine, I find that the implementation of method *Get-Parent-NS-IP* actually returns Nameserver objects, so it is in fact doing what *Get-Parent-NS-Names-and-IPs* should do. We can fix this discrepancy in -Engine by simply renaming the method and by updating the `t/methodsv2.t` file so as to test the parent name server name/IP pairs.

matsduf: In the consistent situation it is simple. The delegation of the parent from all grandparents have the same name server names as the names of the NS in the parent zone on all parent name servers. I think I wanted to avoid the inconsistent situation and therefore avoided the names.

matsduf: > The previous version of the message tags used `ns_ip_list` arguments. Wouldn’t it be better to report the NS names along with their IPs?

@marc-vanderwal, I made a mistake which mislead you. `ns_ip_list` is a list of name servers by IP address. `ip_list` is an undefined argument name, which I used by mistake. What you ask for is `ns_list`, which is a list of name servers by name and address pair.

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -20,80 +21,96 @@
...
+| DS01_DS_ALGO_DEPRECATED  | ERROR   | ns_ip_list, keytag, ds_algo_num, ds_algo_descr | The DS record with keytag {keytag} uses a deprecated digest algorithm {ds_algo_num} ({ds_algo_descr}). Fetched from parent name servers (IP) "{ns_ip_list}".               |
+| DS01_DS_ALGO_NOT_DS      | ERROR   | ns_ip_list, keytag, ds_algo_num, ds_algo_descr | The DS record with keytag {keytag} uses a digest algorithm {ds_algo_num} ({ds_algo_descr}) not meant for DS records. Fetched from parent name servers (IP) "{ns_ip_list}". |
+| DS01_DS_ALGO_OK          | INFO    | ns_ip_list, keytag, ds_algo_num, ds_algo_descr | The DS record with keytag {keytag} uses digest algorithm {ds_algo_num} ({ds_algo_descr}), which is OK. Fetch from parent name servers (IP) "{ns_ip_list}".                 |
+| DS01_DS_ALGO_PRIVATE     | ERROR   | ns_ip_list, keytag, ds_algo_num                | The DS record with keytag {keytag} uses a digest algorithm {ds_algo_num} for private use. parent name servers (IP) "{ns_ip_list}".                                         |
```
tgreenx:
```suggestion
| DS01_DS_ALGO_PRIVATE     | ERROR   | ns_ip_list, keytag, ds_algo_num                | The DS record with keytag {keytag} uses a digest algorithm {ds_algo_num} for private use. Fetched from parent name servers (IP) "{ns_ip_list}".                                         |
```

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -20,80 +21,96 @@
...
+| DS01_DS_ALGO_RESERVED    | ERROR   | ns_ip_list, keytag, ds_algo_num                | The DS record with keytag {keytag} uses a reserved digest algorithm {ds_algo_num} on name servers (IP) "{ns_ip_list}".                                                     |
+| DS01_DS_ALGO_UNASSIGNED  | ERROR   | ns_ip_list, keytag, ds_algo_num                | The DS record with keytag {keytag} uses a unassigned digest algorithm {ds_algo_num} on parent name servers (IP) "{ns_ip_list}".                                            |
```
tgreenx:
```suggestion
| DS01_DS_ALGO_UNASSIGNED  | ERROR   | ns_ip_list, keytag, ds_algo_num                | The DS record with keytag {keytag} uses an unassigned digest algorithm {ds_algo_num} on parent name servers (IP) "{ns_ip_list}".                                            |
```

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -108,81 +125,111 @@ follows the specification for DNS queries as specified in
...
+7.  If the *Non-Algo 2 DS* set is non-empty do:
+
+    1. For each pair of IP address and key tag in the *Algo 2 DS* set remove the
+       same pair from the *Non-Algo 2 DS* set.
+    2. For each key tag from the *Non-Algo 2 DS* set extract all IP addresses for
+       the key tag and output DS01_DS_ALGO_2_MISSING with key tag and the
+       extracted list of IP addresses.
+
+8.  If the *Responds Without Valid DS* and *Responds With DS* sets are empty
+    then output *[DS01_NO_RESPONSE]*.
```
tgreenx:
```suggestion
    then output *[DS01_NO_RESPONSE]* with the name server IP from
    the *Ignored Parent NS IP* set.
```

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -108,81 +125,111 @@ follows the specification for DNS queries as specified in
...
+
+9.  If the *Responds Without Valid DS* is non-empty then do:
+    1. If the *Responds With DS* set is empty then output
+       *[DS01_PARENT_ZONE_NO_DS]* with name server IP from the *Responds without
+       valid DS* set.
```
tgreenx:
```suggestion
       *[DS01_PARENT_ZONE_NO_DS]* with name server IP from the *Responds Without
       Valid DS* set.
```

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -108,81 +125,111 @@ follows the specification for DNS queries as specified in
...
+    * Sets:
+      * *[DS01_DS_ALGO_DEPRECATED]*
+      * *[DS01_DS_ALGO_RESERVED]*
+      * *[DS01_DS_ALGO_UNASSIGNED]*
+      * *[DS01_DS_ALGO_PRIVATE]*
+      * *[DS01_DS_ALGO_NOT_DS]*
+      * *[DS01_DS_ALGO_OK]*
+
+7.  If the *Non-Algo 2 DS* set is non-empty do:
+
```
tgreenx:
```suggestion
```

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -219,23 +266,32 @@ No special terminology for this test case.
...
 [DNSSEC Response]:                                    ../DNSQueryAndResponseDefaults.md#default-handling-of-a-dnssec-response
-[DS01_DIGEST_NOT_SUPPORTED_BY_ZM]:                    #summary
 [DS01_DS_ALGO_2_MISSING]:                             #summary
 [DS01_DS_ALGO_DEPRECATED]:                            #summary
 [DS01_DS_ALGO_NOT_DS]:                                #summary
+[DS01_DS_ALGO_OK]:                                    #summary
+[DS01_DS_ALGO_PRIVATE]:                               #summary
 [DS01_DS_ALGO_RESERVED]:                              #summary
+[DS01_DS_ALGO_UNASSIGNED]:                            #summary
+[DS01_PARENT_SERVER_NO_DS]:                           #summary
```
tgreenx, line 298:
```suggestion
[DS01_NO_RESPONSE]:                                   #summary
[DS01_PARENT_SERVER_NO_DS]:                           #summary
```

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -20,80 +21,96 @@
...
+| DS01_PARENT_SERVER_NO_DS | ERROR   | ns_ip_list                                     | The following name servers do not provide DS record or have not been properly configured. Fetched from parent name servers (IP) "{ns_ip_list}".                            |
+| DS01_PARENT_ZONE_NO_DS   | NOTICE  | ns_ip_list                                     | The parent zone provides no DS records for the child zone. Fetched from parent name servers (IP) "{ns_ip_list}".                                                           |
+| DS01_ROOT_N_NO_UNDEL_DS  | INFO    |                                                | Tested zone is the root zone, but no undelegated DS has been provided. DS is not tested.                                                                                   |
+| DS01_UNDEL_N_NO_UNDEL_DS | INFO    |                                                | Tested zone is undelegated, but no undelegated DS has been provided. DS is not tested.                                                                                     |
```
tgreenx: I suggest that every "(IP)" string are removed from message ids in this table. It's a change that doesn't add anything, imho.

matsduf: When #1418 has been merged we can change from `ns_ip_list` to `ns_list`.

matsduf: Removed "(IP)".

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -191,14 +193,26 @@ queries follow, unless otherwise specified below, what is specified for
...
 
-6.  For each of the sets matching each of the following message tags do if the set
+6.  To be considered in the steps below:
+    * If when outputting a message tag a list of name servers ("ns_list" in the
+      [message tag specification][Summary]) is included, the name or names of the
+      name server should be looked up from the IP address in the
+      *Parent Name and IP* set (if multiple names, it results from multiple name
+      server names resolving into the same IP address).
+    * For message tags with "ns_list", name and IP address pairs should be
+      included, possibly multiple pairs for the same IP address.
```
tgreenx: In previous (recent) specifications we took the habit of simply stating the sentence below in the `Summary` section, instead of over-complicating the test procedure with specificities such as these. E.g. see DNSSEC10 spec.

> The name server names are assumed to be available at the time when the msgid is created, if the argument name is "ns" or "ns_list" even when in the "[Test procedure]" below it is only referred to the IP address of the name servers.

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -108,81 +125,128 @@ follows the specification for DNS queries as specified in
...
+       same pair from the *Non-Algo 2 DS* set.
+    2. For each key tag from the *Non-Algo 2 DS* set extract all IP addresses for
+       the key tag and output DS01_DS_ALGO_2_MISSING with key tag and the
+       extracted list of IP addresses (complemented with the name server name or
+       names).
+
+9.  If the *Responds Without Valid DS* and *Responds With DS* sets are empty
+    then output *[DS01_NO_RESPONSE]* with the name server IP from
+    the *Ignored Parent NS IP* set (complemented with the name server name or
+    names).
```
tgreenx: Currently this tag will be outputted for any undelegated test (see comment above).

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -108,81 +125,128 @@ follows the specification for DNS queries as specified in
...
-6.  Retrieve all name server IP addresses for the parent zone of
-    *Child Zone* using [Method1] (store as "Parent NS IP").
-
-7.  For each parent name server in *Parent NS IP* do:
+1. Create a [DNSSEC Query] with query type DS and query name *Child Zone*
+   ("DS Query").
+
+2.  Retrieve all name server names and IP addresses for the parent zone of
+    *Child Zone* using method [Get-Parent-NS-Names-and-IPs]
+    ("Parent Name and IP").
```
tgreenx: This method returns an empty set for undelegated test. This is not handled in the following test procedure. See comment https://github.com/zonemaster/zonemaster/pull/1413#discussion_r2469465436.

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -108,81 +125,128 @@ follows the specification for DNS queries as specified in
...
+       Valid DS* set (complemented with the name server name or names).
+    2. Else, output *[DS01_PARENT_SERVER_NO_DS]* with name server IP from the
+       *Responds Without Valid DS* set (complemented with the name server name or
+       names).
+
+11. If *Child Zone* is "." (i.e. root zone) and *Undelegated DS* is empty then
+    output *[DS01_ROOT_N_NO_UNDEL_DS]*.
+
+12. If *Undelegated Test* is TRUE and *Undelegated DS* is empty then output
+    *[DS01_UNDEL_N_NO_UNDEL_DS]*.
```
tgreenx: Incompatible with Engine implementation of root data for root zone. See comment https://github.com/zonemaster/zonemaster/pull/1413#discussion_r2469465643.

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
matsduf: That is also a possible model. I feel that the solution in DNSSEC10 is more a fix. What is the problem with the way I wrote it now?

In DNSSEC the possibility that the same IP address can have more than one name is not considered. We should have that at least.

In this test case an IP address can be specified as "-" when there is no IP address. I think that should be explicitly handled.

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
matsduf: That cannot be true. If undelegated test is TRUE then *Parent Name and IP* will be empty, which means that that the procedure will not enter the loop at 5, which means that *Ignored Parent NS IP* will be empty.

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
matsduf: Yes, the set will be empty (by design). I do not see the problem with that. Where do you run into problem?

```
>   Note: The *Parent Name and IP* set will be empty if *Undelegated test* is
>   TRUE, if *Undelegated DS* is non-empty or if *Child Zone* is ".", i.e. root
>   zone.
```

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
matsduf: > Currently this also outputs DS01_UNDEL_N_NO_UNDEL_DS, because of how root data is implemented in Engine. The method that is used to check for undelegated data of a zone is the same method used for populating the root data of the root zone. Which means that the root zone is always considered to have undelegated data, and so "Undelegated Test" is true. This part of the new specification should be changed with that in mind.

Does that mean that it is not possible to make a distinction between testing root as pointed out by the hint and testing root by undelegated data? -- If so, should testing root always be considered to be normal test or always be considered to be undelegated test? I think that normal test is more reasonable.

(To be honest, I do not think that Zonemaster should test the root zone at all. That would simplify things.)

tgreenx:
Implementation in https://github.com/zonemaster/zonemaster-engine/pull/1474.

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
matsduf: I realize that there is a missing requirement that *Ignored Parent NS IP* is non-empty. I will fix.

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
matsduf: Root is always root. There is never any DS unless specified. I fix that.

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
matsduf: I have done an update.

docs/public/specifications/tests/DNSSEC-TP/dnssec01.md
```diff
@@ -221,32 +222,30 @@ queries follow, unless otherwise specified below, what is specified for
...
+       extracted list of IP addresses.
 
-10. If the *Responds Without Valid DS* is non-empty then do:
-    1. If the *Responds With DS* set is empty then output
-       *[DS01_PARENT_ZONE_NO_DS]* with name server IP from the *Responds Without
-       Valid DS* set (complemented with the name server name or names).
-    2. Else, output *[DS01_PARENT_SERVER_NO_DS]* with name server IP from the
-       *Responds Without Valid DS* set (complemented with the name server name or
-       names).
+10. If the *Responds Without Valid DS* is empty, the *Responds With DS* sets is
```
tgreenx:
```suggestion
10. If the *Responds Without Valid DS* is empty, the *Responds With DS* set is
```

PR #1409, opened 2025-07-15
Updates DNSSEC05 test case specification

matsduf:
## Purpose

This PR updates the test case specification for DNSSEC05

* Includse explicit msgid.
* Matches current template.
* New algorithms from IANA and new classifications.
* Compacter outputs with only one output for the same key ID.

## Context

Discussions at F2F in July 2025.

The changes in the specification requires an update of the implementation and update of the translations of the messages.

## How to test this PR

Review.

PR #1415, opened 2025-07-31
Zone11: different handling of root zone, TLD zones and zones in .arpa

### LINKED ISSUE #1256: SPF warning for TLD
With the MX check you explicitly check that a TLD should not have an MX.
But the SPF check, warns if a TLD does not have a SPF record.

The SPF standard explicitly tells clients not to check for SPF records on anything on the Public Suffix List. All TLDs are on the list.

marc-vanderwal:
## Purpose

This PR amends the specification for Zone11 so that different message tags of different severity levels are output if Zonemaster is run on the root zone, a TLD zone or a zone inside .arpa. It also cleans up some rough edges and migrates to MethodsV2.

## Context

See #1256.

## Changes

- Root zones, TLD zones and zones inside .arpa are tested differently: an INFO is output if no SPF or a null SPF is found, and a message tag of higher level is output if a non-null SPF is found.
- Use MethodsV2 instead of legacy Methods.
- Miscellaneous small rewordings

### Breaking message tag changes

This PR changes the arguments to the following existing message tags:

| Tag                              | Old arguments | New arguments |
|:---------------------------------|:--------------|:--------------|
| Z11_DIFFERENT_SPF_POLICIES_FOUND | ns_ip_list    | ns_list       |

This PR replaces the following message tags with new tags:

| Old tag                   | Old tag arguments | New tag                  | New tag arguments |
|:--------------------------|:------------------|:-------------------------|:------------------|
| Z11_SPF1_MULTIPLE_RECORDS | ns_ip_list        | Z11_SPF_MULTIPLE_RECORDS | ns_list           |
| Z11_SPF1_SYNTAX_ERROR     | ns_ip_list        | Z11_SPF_SYNTAX_ERROR     | ns_list           |
| Z11_SPF1_SYNTAX_OK        | ns_ip_list        | Z11_SPF_SYNTAX_OK        | ns_list           |

## How to test this PR

Review.

### DIFF
--- a/docs/public/specifications/tests/Zone-TP/zone11.md
+++ b/docs/public/specifications/tests/Zone-TP/zone11.md
@@ -63,18 +63,18 @@ server.
 
 ## Summary
 
-Message Tag                      | Level   | Arguments          | Message ID for message tag
-:--------------------------------|:--------|:-------------------|:--------------------------------------------
-Z11_DIFFERENT_SPF_POLICIES_FOUND | NOTICE  | ns_ip_list         | The following name servers returned the same SPF policy. Name servers: {ns_ip_list}.
-Z11_INCONSISTENT_SPF_POLICIES    | WARNING |                    | One or more name servers do not publish the same SPF policy as the others.
-Z11_NO_SPF_FOUND                 | NOTICE  | domain             | No SPF policy was found for {domain}.
-Z11_NO_SPF_NON_MAIL_DOMAIN       | INFO    | domain             | No SPF policy was found for {domain}, which is a type of domain (root, TLD or under .ARPA) not expected to be used for email.
-Z11_NON_NULL_SPF_NON_MAIL_DOMAIN | NOTICE  | domain             | A non-null SPF policy was found on {domain}, although this type of domain (root, TLD or under .ARPA) is not expected to be used for email.
-Z11_NULL_SPF_NON_MAIL_DOMAIN     | INFO    | domain             | A null SPF policy was found on {domain}, which is a type of domain (root, TLD or under .ARPA) not expected to be used for email.
-Z11_SPF_MULTIPLE_RECORDS         | ERROR   | ns_ip_list         | The following name servers returned more than one SPF policy. Name servers: {ns_ip_list}.
-Z11_SPF_SYNTAX_ERROR             | ERROR   | domain, ns_ip_list | The SPF policy of {domain} has a syntax error. Policy retrieved from the following nameservers: {ns_ip_list}.
-Z11_SPF_SYNTAX_OK                | INFO    | domain             | The SPF policy of {domain} has correct syntax.
-Z11_UNABLE_TO_CHECK_FOR_SPF      | ERROR   |                    | None of the zone’s name servers responded with an authoritative response to queries for SPF policies.
+Message Tag                      | Level   | Arguments       | Message ID for message tag
+:--------------------------------|:--------|:----------------|:--------------------------------------------
+Z11_DIFFERENT_SPF_POLICIES_FOUND | NOTICE  | ns_list         | The following name servers returned the same SPF policy. Name servers: {ns_list}.
+Z11_INCONSISTENT_SPF_POLICIES    | WARNING |                 | One or more name servers do not publish the same SPF policy as the others.
+Z11_NO_SPF_FOUND                 | NOTICE  | domain          | No SPF policy was found for {domain}.
+Z11_NO_SPF_NON_MAIL_DOMAIN       | INFO    | domain          | No SPF policy was found for {domain}, which is a type of domain (root, TLD or under .ARPA) not expected to be used for email.
+Z11_NON_NULL_SPF_NON_MAIL_DOMAIN | NOTICE  | domain          | A non-null SPF policy was found on {domain}, although this type of domain (root, TLD or under .ARPA) is not expected to be used for email.
+Z11_NULL_SPF_NON_MAIL_DOMAIN     | INFO    | domain          | A null SPF policy was found on {domain}, which is a type of domain (root, TLD or under .ARPA) not expected to be used for email.
+Z11_SPF_MULTIPLE_RECORDS         | ERROR   | ns_list         | The following name servers returned more than one SPF policy. Name servers: {ns_list}.
+Z11_SPF_SYNTAX_ERROR             | ERROR   | domain, ns_list | The SPF policy of {domain} has a syntax error. Policy retrieved from the following nameservers: {ns_list}.
+Z11_SPF_SYNTAX_OK                | INFO    | domain          | The SPF policy of {domain} has correct syntax.
+Z11_UNABLE_TO_CHECK_FOR_SPF      | ERROR   |                 | None of the zone’s name servers responded with an authoritative response to queries for SPF policies.
 
 The value in the Level column is the default severity level of the message. The
 severity level can be changed in the [Zonemaster-Engine profile]. Also see the
@@ -82,6 +82,10 @@ severity level can be changed in the [Zonemaster-Engine profile]. Also see the
 
 The argument names in the Arguments column lists the arguments used in the
 message. The argument names are defined in the [argument list].
+
+The name server names are assumed to be available at the time when the msgid
+is created, if the argument name is "ns_list" even when in the "[Test
+procedure]" below it is only referred to the IP address of the name servers.
 
 ## Test procedure
 
@@ -94,28 +98,30 @@ same specification.
 1. Create a [DNS Query] with query type TXT and query name *Child Zone* ("TXT
    query").
 
-2. Create an empty set of pairs of IP addresses and strings, "SPF-Policies".
+2. Create an empty set of pairs of (names and IP address) pairs and strings,
+   "SPF-Policies".
 
-3. Retrieve all name server IP addresses for *Child Zone* using methods
-   [Get-Del-NS-IPs] and [Get-Zone-NS-IPs] ("Name Server IP").
+3. Retrieve all name server names and IP addresses for *Child Zone* using
+   methods [Get-Del-NS-Names-and-IPs] and [Get-Zone-NS-Names-and-IPs] ("Name
+   Servers").
 
-4. For each name server in *Name Server IP* do:
+4. For each distinct name server IP address in *Name Servers* do:
 
    1. Send *TXT Query* to the name server and collect the [DNS Response].
 
-   2. Go to the next name server IP address if at least one of the following
-      criteria is met:
+   2. Go to the next name server if at least one of the following criteria is
+      met:
 
       1. There is no DNS response.
       2. [RCODE Name] of the response is not "NoError".
       3. The AA flag is not set in the response.
 
    3. If the name server responds with no TXT record, then add the pair
-      consisting of the *Name Server IP* and the empty string to the
+      consisting of the *Name Servers* and the empty string to the
       *SPF-Policies* set.
 
    4. If the name server responds with at least one TXT record and none is an
-      [SPF TXT record], then add the pair consisting of the *Name Server IP*
+      [SPF TXT record], then add the pair consisting of the *Name Servers*
       and the empty string to the *SPF-Policies* set.
 
    5. If the name server responds with at least one TXT record that is an [SPF
@@ -123,7 +129,7 @@ same specification.
 
        1. [Concatenate] all strings in the RDATA field.
        2. Lowercase the resulting string.
-       3. Add a pair consisting of the *Name Server IP* and the lowercase
+       3. Add a pair consisting of the *Name Servers* and the lowercase
           string thus derived from the RDATA field to the *SPF-Policies* set.
 
    6. Go to the next name server.
@@ -139,29 +145,33 @@ same specification.
 
    2. Else, output *[Z11_NO_SPF_FOUND]* and terminate the test.
 
-7. Compare the set of *SPF-Policies* retrieved from all name servers. If at
+7. For all messages outputted below, if an IP address in *Name Servers* is
+   connected to more than one name server name, then all names should be
+   included with the message tag.
+
+8. Compare the set of *SPF-Policies* retrieved from all name servers. If at
    least two different name servers have returned different sets of SPF
    policies, then:
 
    1. Output *[Z11_INCONSISTENT_SPF_POLICIES]*.
    2. Group *SPF-Policies* by equal sets of SPF policies, such that a set of
-      SPF policies is mapped to the list of *Name Server IPs* that returned it.
+      SPF policies is mapped to the list of *Name Servers* that returned it.
    3. For each such group of name servers, output
       *[Z11_DIFFERENT_SPF_POLICIES_FOUND]*.
    4. Terminate the test.
 
-8. If the *SPF-Policies* set contains at least two entries with the same IP
+9. If the *SPF-Policies* set contains at least two entries with the same IP
    address, then output *[Z11_SPF_MULTIPLE_RECORDS]* with the list of
    nameservers that returned more than one SPF policy and terminate the test.
 
-9. The following steps assume that all pairs in the *SPF-Policies* set have
-   the same string ("SPF policy").
+10. The following steps assume that all pairs in the *SPF-Policies* set have
+    the same string ("SPF policy").
 
-10. If the *SPF Policy* does not [pass the syntax check][passing the syntax
+11. If the *SPF Policy* does not [pass the syntax check][passing the syntax
     check] for SPF records, then output *[Z11_SPF_SYNTAX_ERROR]* and terminate
     the test.
 
-11. If the *Child Zone* is the root zone ("."), a [TLD] or a zone under
+12. If the *Child Zone* is the root zone ("."), a [TLD] or a zone under
     .ARPA, then:
 
    1. If the *SPF Policy* is a [Null SPF] policy, then output
@@ -170,7 +180,7 @@ same specification.
    2. If the *SPF Policy* is not a [Null SPF] policy, then output
       *[Z11_NON_NULL_SPF_NON_MAIL_DOMAIN]* and terminate the test.
 
-12. If no other message was outputted by this test case, then output
+13. If no other message was outputted by this test case, then output
     *[Z11_SPF_SYNTAX_OK]*.
 
 ## Outcome(s)
@@ -234,8 +244,8 @@ None.
 [Dotless Domains Considered Harmful]:   https://www.iab.org/documents/correspondence-reports-documents/2013-2/iab-statement-dotless-domains-considered-harmful/
 [Email Domain]:                         #terminology
 [ERROR]:                                ../SeverityLevelDefinitions.md#error
-[Get-Del-NS-IPs]:                       ../MethodsV2.md#method-get-delegation-ns-ip-addresses
-[Get-Zone-NS-IPs]:                      ../MethodsV2.md#method-get-zone-ns-ip-addresses
+[Get-Del-NS-Names-and-IPs]:             ../MethodsV2.md#method-get-delegation-ns-names-and-ip-addresses
+[Get-Zone-NS-Names-and-IPs]:            ../MethodsV2.md#method-get-zone-ns-names-and-ip-addresses
 [INFO]:                                 ../SeverityLevelDefinitions.md#info
 [Internet Architecture Board]:          https://www.iab.org/
 [Message Tag Specification]:            MessageTagSpecification.md
--- a/docs/public/specifications/tests/Zone-TP/zone11.md
+++ b/docs/public/specifications/tests/Zone-TP/zone11.md
@@ -83,9 +83,10 @@ severity level can be changed in the [Zonemaster-Engine profile]. Also see the
 The argument names in the Arguments column lists the arguments used in the
 message. The argument names are defined in the [argument list].
 
-The name server names are assumed to be available at the time when the msgid
-is created, if the argument name is "ns_list" even when in the "[Test
-procedure]" below it is only referred to the IP address of the name servers.
+Name servers may have multiple IP addresses bound to the same name, and the
+same IP address may be used by multiple name server names. Message tags whose
+argument lists include "ns_list" shall contain all such name and IP address
+pairs.
 
 ## Test procedure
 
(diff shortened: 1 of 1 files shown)


docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -33,6 +33,9 @@ It is assumed that *Child Zone* has been tested and reported by
 or name servers not giving a correct DNS response for an authoritative name
 server.
 
+The test is skipped if the zone being tested is the root zone, a [TLD] zone
+or a zone in the .ARPA tree.
```
matsduf: I disagree, but the tags should be adjusted to those kind of zones.

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -41,12 +44,13 @@ server.
...
 Z11_NO_SPF_FOUND                 | NOTICE  | domain             | No SPF policy was found for {domain}.
-Z11_SPF1_MULTIPLE_RECORDS        | ERROR   | ns_ip_list         | The following name servers returned more than one SPF policy. Name servers: {ns_ip_list}.
-Z11_SPF1_SYNTAX_ERROR            | ERROR   | domain, ns_ip_list | The SPF policy of {domain} has a syntax error. Policy retrieved from the following nameservers: {ns_ip_list}.
-Z11_SPF1_SYNTAX_OK               | INFO    | domain             | The SPF policy of {domain} has correct syntax.
+Z11_SPF_MULTIPLE_RECORDS         | ERROR   | ns_ip_list         | The following name servers returned more than one SPF policy. Name servers: {ns_ip_list}.
+Z11_SPF_SYNTAX_ERROR             | ERROR   | domain, ns_ip_list | The SPF policy of {domain} has a syntax error. Policy retrieved from the following nameservers: {ns_ip_list}.
+Z11_SPF_SYNTAX_OK                | INFO    | domain             | The SPF policy of {domain} has correct syntax.
+Z11_SKIPPING_NON_EMAIL_DOMAIN    | INFO    |                    | The test is skipped, because the domain is either the root name, a TLD or a zone in .ARPA.
```
matsduf: See above.

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -64,15 +68,18 @@ Response Defaults]. The handling of the DNS responses on the DNS queries follow,
 unless otherwise specified below, what is specified for [DNS Response] in the
 same specification.
 
-1. Create a [DNS Query] with query type TXT and query name *Child Zone* ("TXT
+1. If the *Child Zone* is the root zone ("."), a [TLD] or a zone in the .ARPA
+   [TLD], then output *[Z11_SKIPPING_NON_EMAIL_DOMAIN]* and terminate the test.
```
matsduf: See above.

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -41,12 +44,13 @@ server.
...
-Z11_INCONSISTENT_SPF_POLICIES    | WARNING |                    | One or more name servers do not publish the same SPF policy as the others.
 Z11_DIFFERENT_SPF_POLICIES_FOUND | NOTICE  | ns_ip_list         | The following name servers returned the same SPF policy, but other name servers returned a different policy. Name servers: {ns_ip_list}.
+Z11_INCONSISTENT_SPF_POLICIES    | WARNING |                    | One or more name servers do not publish the same SPF policy as the others.
 Z11_NO_SPF_FOUND                 | NOTICE  | domain             | No SPF policy was found for {domain}.
-Z11_SPF1_MULTIPLE_RECORDS        | ERROR   | ns_ip_list         | The following name servers returned more than one SPF policy. Name servers: {ns_ip_list}.
-Z11_SPF1_SYNTAX_ERROR            | ERROR   | domain, ns_ip_list | The SPF policy of {domain} has a syntax error. Policy retrieved from the following nameservers: {ns_ip_list}.
-Z11_SPF1_SYNTAX_OK               | INFO    | domain             | The SPF policy of {domain} has correct syntax.
+Z11_SPF_MULTIPLE_RECORDS         | ERROR   | ns_ip_list         | The following name servers returned more than one SPF policy. Name servers: {ns_ip_list}.
```
matsduf: Here an other places we should include "ns_list" instead of "ns_ip_list" because it gives more information to the user.

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -64,15 +68,18 @@ Response Defaults]. The handling of the DNS responses on the DNS queries follow,
...
+2. Create a [DNS Query] with query type TXT and query name *Child Zone* ("TXT
    query").
 
-2. Create an empty set of pairs of IP addresses and strings, "SPF-Policies".
+3. Create an empty set of pairs of IP addresses and strings, "SPF-Policies".
 
-3. Obtain the set of name server IP addresses using [Method4] and [Method5]
-   ("Name Server IP").
+4. Retrieve all name server IP addresses for *Child Zone* using methods
+   [Get-Del-NS-IPs] and [Get-Zone-NS-IPs] ("Name Server IP").
```
matsduf: We should get name and IP of the name servers to be able to serve the messages with both.

matsduf (changes requested):

marc-vanderwal:
I have changed the approach and addressed your review comments. Please re-review.

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -33,21 +33,29 @@ It is assumed that *Child Zone* has been tested and reported by
 or name servers not giving a correct DNS response for an authoritative name
 server.
 
+The root zone ("."), [TLD] zones and zones within .ARPA are treated
```
matsduf: For me "under" sounds better.
```suggestion
The root zone ("."), [TLD] zones and zones under .ARPA are treated
```

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -33,21 +33,29 @@ It is assumed that *Child Zone* has been tested and reported by
...
+differently because the corresponding domains are not expected to send e-mail.
+Additional discussion on the reason why these zones are unsuitable as e-mail
+domains is provided in the [Zone09 test specification].
```
matsduf: The Zone09 test case specification excludes all zones under .arpa, but in Objective it only refer to in-addr.arpa and ip6.arpa zones. It should be clarified there that it is about any .arpa sub-zone. Maybe this PR could clarify in https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/tests/Zone-TP/zone09.md?plain=1#L43-L47 that the exclusion is not only about reverse zones.

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -33,21 +33,29 @@ It is assumed that *Child Zone* has been tested and reported by
...
-Z11_UNABLE_TO_CHECK_FOR_SPF      | ERROR   |                    | None of the zone’s name servers responded with an authoritative response to queries for SPF policies.
+Message Tag                      | Level   | Arguments       | Message ID for message tag
+:--------------------------------|:--------|:----------------|:--------------------------------------------
+Z11_DIFFERENT_SPF_POLICIES_FOUND | NOTICE  | ns_list         | The following name servers returned the same SPF policy, but other name servers returned a different policy. Name servers: {ns_list}.
+Z11_INCONSISTENT_SPF_POLICIES    | WARNING |                 | One or more name servers do not publish the same SPF policy as the others.
+Z11_NO_SPF_FOUND                 | NOTICE  | domain          | No SPF policy was found for {domain}.
+Z11_NO_SPF_NON_MAIL_DOMAIN       | INFO    | domain          | No SPF policy was found for {domain}, which is not an e-mail domain.
+Z11_NON_NULL_SPF_NON_MAIL_DOMAIN | NOTICE  | domain          | A non-null SPF policy was found on {domain}, although it is not an e-mail domain.
```
matsduf: I think that "not an e-mail domain" is not precise enough. Any domain without MX could be said to be "not an e-mail domain".
```suggestion
Z11_NON_NULL_SPF_NON_MAIL_DOMAIN | NOTICE  | domain          | A non-null SPF policy was found on {domain}, although this type of domain (root, TLD or under .ARPA) is unexpected to be used for e-mail.
```

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -33,21 +33,29 @@ It is assumed that *Child Zone* has been tested and reported by
...
+Z11_NULL_SPF_NON_MAIL_DOMAIN     | INFO    | domain          | A null SPF policy was found on {domain}, which is not an e-mail domain.
```
matsduf:
```suggestion
Z11_NULL_SPF_NON_MAIL_DOMAIN     | INFO    | domain          | A null SPF policy was found on {domain}, which is a type of domain (root, TLD or under .ARPA) which is unexpected to be used for e-mail.
```

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -67,70 +75,86 @@ same specification.
...
+   "SPF-Policies".
 
-3. Obtain the set of name server IP addresses using [Method4] and [Method5]
-   ("Name Server IP").
+3. Retrieve all name server names and IP addresses for *Child Zone* using
+   methods [Get-Del-NS-Names-and-IPs] and [Get-Zone-NS-Names-and-IPs] ("Name
+   Servers").
 
-4. For each name server in *Name Server IP* do:
+4. For each name server in *Name Servers* do:
```
matsduf: This PR has chosen a different approach than #1409 for DNSSEC05. In the implementation the name server IP what counts. In the messages name/IP should be included. I think we should find a wording that we consistently use in all updates where name/IP are expected in the messages.

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -67,70 +75,86 @@ same specification.
...
    the same string ("SPF policy").
 
 10. If the *SPF Policy* does not [pass the syntax check][passing the syntax
-    check] for SPF version 1 records, then output *[Z11_SPF1_SYNTAX_ERROR]* and
-    terminate the test.
+    check] for SPF records, then output *[Z11_SPF_SYNTAX_ERROR]* and terminate
+    the test.
+
+11. If the *Child Zone* is the root zone ("."), a [TLD] or a zone within
+    .ARPA, then:
```
matsduf:
```suggestion
11. If the *Child Zone* is the root zone ("."), a [TLD] or a zone under
    .ARPA, then:
```

matsduf (dismissed):

docs/public/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: Sounds good indeed.

docs/public/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: I’d rather create a seperate PR for clarifying that paragraph in Zone09. Meanwhile, I’ll just rewrite this paragraph here so that the text does not refer to Zone09 anymore.

docs/public/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: Good suggestion.

docs/public/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: Good suggestion, but I’ll slightly reword in order to avoid repeating “which is” twice in the same sentence.

docs/public/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: Okay, I’ll reword as “For each name server IP address in *Name Server IP* do:”

matsduf: Well, there are some aspects of the approach in #1409. First there is the text in https://github.com/zonemaster/zonemaster/blob/e8b95a6cfd8d43d59577c92b94b179033c9c0aa5/docs/public/specifications/tests/DNSSEC-TP/dnssec05.md?plain=1#L115-L118 and then https://github.com/zonemaster/zonemaster/blob/e8b95a6cfd8d43d59577c92b94b179033c9c0aa5/docs/public/specifications/tests/DNSSEC-TP/dnssec05.md?plain=1#L175-L178

I am not sure if the approach in #1409 is the best one. In the implementation you probably want to go by IP address, and should that be reflected in the specification?

docs/public/specifications/tests/Zone-TP/zone11.md
matsduf: I think there is a point keeping the two specifications the same in this respect.

marc-vanderwal: It’s better to avoid cross-referencing specifications too much. Besides, the reason why we treat the root, TLD zones and .ARPA differently here is slightly different from Zone09: in Zone09, it’s because the resulting email addresses are unlikely to make sense as recipient addresses, but in Zone11, it’s because they are unlikely to make sense as sender addresses.

I took the liberty to make minor adjustments (other than what is strictly necessary) to the text instead of merely copy-pasting it from Zone09. What do you think? Maybe we can import my text back into Zone09 later?

docs/public/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: Oh, I didn’t notice those bits in the DNSSEC05 proposal.

Yes, it is a recurring pattern that we iterate on *name server IPs* under the hood but that we report *name server name and IP pairs* in user-visible messages. And we need to be careful with the edge case where a zone has more than one name server name pointing to the same IP address.

So I think the approach makes sense. Maybe we should keep it somewhere in a template?

matsduf: It is not only the edge case when a zone has serveral names to the same IP address, it is also the case when the name in the delegation does not match the name in the zone, but the IP is the same. The latter is probably more common.

We have a template in https://github.com/zonemaster/zonemaster/blob/master/docs/internal/templates/specifications/tests/Template01.md which probably needs some updates.

matsduf (dismissed):
Looks fine. Do you plan to create a scenario PR too?

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -1,7 +1,7 @@
-# ZONE11: SPF policy validation
+# Zone11: SPF policy validation
 
 ## Test case identifier
-**ZONE11**
+**Zone11**
```
tgreenx: This should probably not be lowercased now. I'm not sure if we even want that here, but if we do, it might be best to do it for all specifications at once so we do not get any discrepancies.

tgreenx (commented):
LGTM. Just one comment.

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -39,15 +63,18 @@ server.
...
-Z11_NO_SPF_FOUND                 | NOTICE  | domain             | No SPF policy was found for {domain}.
-Z11_SPF1_MULTIPLE_RECORDS        | ERROR   | ns_ip_list         | The following name servers returned more than one SPF policy. Name servers: {ns_ip_list}.
-Z11_SPF1_SYNTAX_ERROR            | ERROR   | domain, ns_ip_list | The SPF policy of {domain} has a syntax error. Policy retrieved from the following nameservers: {ns_ip_list}.
-Z11_SPF1_SYNTAX_OK               | INFO    | domain             | The SPF policy of {domain} has correct syntax.
-Z11_UNABLE_TO_CHECK_FOR_SPF      | ERROR   |                    | None of the zone’s name servers responded with an authoritative response to queries for SPF policies.
+Message Tag                      | Level   | Arguments       | Message ID for message tag
+:--------------------------------|:--------|:----------------|:--------------------------------------------
+Z11_DIFFERENT_SPF_POLICIES_FOUND | NOTICE  | ns_list         | The following name servers returned the same SPF policy, but other name servers returned a different policy. Name servers: {ns_list}.
```
matsduf: Z11_DIFFERENT_SPF_POLICIES_FOUND is always outputted together with Z11_INCONSISTENT_SPF_POLICIES. The message ID could the be shortened to `The following name servers returned the same SPF policy. Name servers: {ns_list}.`

docs/public/specifications/tests/Zone-TP/zone11.md
matsduf: As far as I can see it is always upper case, but we often refer to the test cases with mixed case (except DNSSEC). We should probably keep upper case as @tgreenx suggests.

marc-vanderwal: Okay, I’ll revert that change.

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -67,70 +98,90 @@ same specification.
...
           string thus derived from the RDATA field to the *SPF-Policies* set.
 
    6. Go to the next name server.
 
 5. If the *SPF-Policies* set is empty, then output
    *[Z11_UNABLE_TO_CHECK_FOR_SPF]* and terminate the test.
 
-6. If all the pairs in the *SPF-Policies* set contain empty strings, then
-   output *[Z11_NO_SPF_FOUND]* and terminate the test.
+6. If all the pairs in the *SPF-Policies* set contain empty strings, then:
```
matsduf: I think the following is easier to read.

```suggestion
6. If all the name server IPs in the *SPF-Policies* set contain empty strings (no "SPF policy"), then:
```

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -67,70 +98,90 @@ same specification.
...
+   2. Else, output *[Z11_NO_SPF_FOUND]* and terminate the test.
 
-7. Compare the set of *SPF-Policies* retrieved from all name servers. If at
+7. For all messages outputted below, if an IP address in *Name Servers* is
+   connected to more than one name server name, then all names should be
+   included with the message tag.
+
+8. Compare the set of *SPF-Policies* retrieved from all name servers. If at
    least two different name servers have returned different sets of SPF
    policies, then:
```
matsduf:
```suggestion
8. Compare the set of *SPF-Policies* retrieved from all name servers (in the
   the *SPF-Policies* set). If at least two different name servers have returned 
   different sets of SPF policies, then:
```

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -67,70 +98,90 @@ same specification.
...
 
-8. If the *SPF-Policies* set contains at least two entries with the same IP
-   address, then output *[Z11_SPF1_MULTIPLE_RECORDS]* with the list of
+9. If the *SPF-Policies* set contains at least two entries with the same IP
+   address, then output *[Z11_SPF_MULTIPLE_RECORDS]* with the list of
    nameservers that returned more than one SPF policy and terminate the test.
 
-9. The following steps assume that all pairs in the *SPF-Policies* set have
-   the same string ("SPF policy").
+10. The following steps assume that all pairs in the *SPF-Policies* set have
```
matsduf:
```suggestion
10. The following steps assume that all name server IPs in the *SPF-Policies* set have
```

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -67,70 +98,90 @@ same specification.
...
+12. If the *Child Zone* is the root zone ("."), a [TLD] or a zone under
+    .ARPA, then:
 
-11. If no other message was outputted by this test case, then output
-    *[Z11_SPF1_SYNTAX_OK]*.
+   1. If the *SPF Policy* is a [Null SPF] policy, then output
+      *[Z11_NULL_SPF_NON_MAIL_DOMAIN]* and terminate the test.
+
+   2. If the *SPF Policy* is not a [Null SPF] policy, then output
+      *[Z11_NON_NULL_SPF_NON_MAIL_DOMAIN]* and terminate the test.
```
matsduf: Not correctly indented in the Github rendering. I think a leading space is missing.

```suggestion
    1. If the *SPF Policy* is a [Null SPF] policy, then output
       *[Z11_NULL_SPF_NON_MAIL_DOMAIN]* and terminate the test.

    2. If the *SPF Policy* is not a [Null SPF] policy, then output
       *[Z11_NON_NULL_SPF_NON_MAIL_DOMAIN]* and terminate the test.
```

matsduf (changes requested):

docs/public/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: Indeed. This is fixed and I checked myself that it renders properly on GitHub.

docs/public/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: Fixed.

docs/public/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: Fixed.

docs/public/specifications/tests/Zone-TP/zone11.md
marc-vanderwal: Fixed.

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -67,70 +98,91 @@ same specification.
...
+8. Compare the set of *SPF-Policies* retrieved from all name servers (in the
+   *SPF-Policies* set). If at least two different name servers have returned
+   different sets of SPF policies, then:
 
    1. Output *[Z11_INCONSISTENT_SPF_POLICIES]*.
    2. Group *SPF-Policies* by equal sets of SPF policies, such that a set of
-      SPF policies is mapped to the list of *Name Server IPs* that returned it.
+      SPF policies is mapped to the list of *Name Servers* that returned it.
    3. For each such group of name servers, output
       *[Z11_DIFFERENT_SPF_POLICIES_FOUND]*.
```
matsduf: It should be stated that the list of name servers ("ns_list") is included with the tag. This applies to other places too.

docs/public/specifications/tests/Zone-TP/zone11.md
```diff
@@ -67,70 +98,91 @@ same specification.
...
-6. If all the pairs in the *SPF-Policies* set contain empty strings, then
-   output *[Z11_NO_SPF_FOUND]* and terminate the test.
+6. If all the name server IPs in the *SPF-Policies* set contain empty strings
+   (no "SPF policy"), then:
+
+   1. If the *Child Zone* is the root zone ("."), a [TLD] or a zone under
+      .ARPA, then output *[Z11_NO_SPF_NON_MAIL_DOMAIN]* and terminate the
+      test.
+
+   2. Else, output *[Z11_NO_SPF_FOUND]* and terminate the test.
```
matsduf: It should be stated that the domain is outputted with the tag.

matsduf (changes requested):
The test case has "ns_list" arguments for some tags, and then there is a chance that the same IP address is used by multiple name server names. It should be stated explicitly that "ns_list" includes all such name/IP pairs.

marc-vanderwal:
Step 7 in the test procedure already states:

> For all messages outputted below, if an IP address in *Name Servers* is connected to more than one name server name, then all names should be included with the message tag.

But I found a paragraph in the “Summary” section that was no longer true. I’ve updated it with a similar text. Does this make step 7 in the test procedure redundant?

matsduf (approved):

tgreenx (approved):
LGTM.

Implementation in https://github.com/zonemaster/zonemaster-engine/pull/1477.

tgreenx (approved):

matsduf:
@marc-vanderwal, please merge.

PR #1418, opened 2025-08-11
MethodsV2: add Get-Parent-NS-Names-and-IPs method

marc-vanderwal:
## Purpose

This PR introduces a new method in MethodsV2, named *Get-Parent-NS-Names-and-IPs* method, which obtains the set of name server names and IP addresses of a zone’s parent zone.

Some DNSSEC test cases need to look at information in the parent zone. There might be problems there, and we need to report the name servers that the information was obtained from. For good ergonomics, it’d be nice to report the name servers as name server/IP pairs, for example “ns1.parent.example/198.54.113.1”, like we do for name servers that appear in the parent’s delegation or in the zone apex’s NS record set.

However, no such method currently exists. A method called *Get-Parent-NS-IP* does exist, and I could reuse its outline for the *Get-Parent-NS-Names-and-IPs* method. Then, *Get-Parent-NS-IP* can be implemented as a method that uses the result of *Get-Parent-NS-Names-and-IPs*, exactly like *Get-Del-NS-IPs* and *Get-Zone-NS-IPs*.

I also took the opportunity to rename *Get-Parent-NS-IP* to *Get-Parent-NS-IPs* (in plural form) in order to make the method’s name consistent with other similarly-named methods in MethodsV2.

Test scenario specifications are also updated to test *Get-Parent-NS-Names-and-IPs* instead of *Get-Parent-NS-IP*.

## Context

[Review comment](https://github.com/zonemaster/zonemaster/pull/1412#discussion_r2265805193) on pull request #1412.

## Changes

- Add new MethodsV2 method *Get-Parent-NS-Names-and-IPs*;
- Redefine *Get-Parent-NS-IP* so that it uses the result of *Get-Parent-NS-Names-and-IPs*;
- Rename *Get-Parent-NS-IP* to *Get-Parent-NS-IPs* and update test case specifications that use this method;
- In MethodsV2 test scenario specification, test *Get-Parent-NS-Names-and-IPs* instead of *Get-Parent-NS-IPs* and update expected data for MethodsV2 test scenarios;
- As a bonus, fix some minor typos and errors I’ve encountered while preparing this PR.

## How to test this PR

Review. This PR is best reviewed commit by commit.


test-zone-data/start-coredns.sh
```diff
@@ -49,7 +49,7 @@ fi
 if ps --no-headers -fC coredns | grep coredns ; then
     echo "Cannot start CoreDNS since it is already running. Exit."
 else
-    coredns --conf ${maincfg}
+    sudo ~vanderwal/.local/bin/coredns --conf ${maincfg}
```
matsduf: No reason to change.

```suggestion
    coredns --conf ${maincfg}
```

docs/public/specifications/tests/MethodsV2.md
```diff
@@ -199,7 +205,8 @@ This Method uses the following input units defined in section [Methods Inputs]:
           *Intermediate Query Name* in the answer section, with the AA bit
           set and [RCODE Name] NoError then do:
           1. If *Intermediate Query Name* is equal to *Child Zone* then
-             1. Save *Server Address* to the *Parent NS IP* set.
+             1. Save the *Server Name* and *Server Address* pair to the
+                *Parent Name Servers* set.
```
matsduf, line 217: It should be clarified what name to store in case of CNAME, the orginal name in the NS record or the owner name of the address record.

docs/public/specifications/tests/MethodsV2.md
```diff
@@ -218,35 +225,32 @@ This Method uses the following input units defined in section [Methods Inputs]:
              5. Do [DNS Lookup] of name server names (A and AAAA) not already
                 listed in the additional section of the response. Follow CNAME if
                 provided.
-             6. For each IP address add the IP address and *Intermediate Query Name*
-                to the *Remaining Servers* set unless the IP address is already
-                listed in *Handled Servers* together with *Intermediate Query Name*.
+             6. For each name and IP address add the name, IP address and
+                *Intermediate Query Name* tuple to the *Remaining Servers* set
+                unless such a tuple is already listed in *Handled Servers*.
```
matsduf: It should be clarified what name to store in case of CNAME, the orginal name in the NS record or the owner name of the address record.

docs/public/specifications/tests/MethodsV2.md
```diff
@@ -137,24 +139,28 @@ This Method uses the following input units defined in section [Methods Inputs]:
...
 > In the loop below, the steps tries to capture the name of the parent zone of
 > **Child Zone** and the IP addresses of the name servers for that parent zone.
 > This is done using a modified version of the "QNAME minimization" technique
 > [RFC 9156]. SOA is the query type used for traversing the tree.
 
-5. While the *Remaining Servers* is non-empty pick next name server IP address
-   and zone name from the set ("Server Address" and "Zone Name") and do:
+5. While the *Remaining Servers* is non-empty pick next name server name, IP
+   address and zone name tuple from the set ("Server Name", "Server Address"
+   and "Zone Name") and do:
```
matsduf, line 158: With this design, if several names resolve to the same IP address, the same process will run for each name. Is there anyway to avoid that?

test-zone-data/start-coredns.sh
marc-vanderwal: My mistake. That was not meant to make it into the PR.

docs/public/specifications/tests/MethodsV2.md
marc-vanderwal: I think it should be the original name in the NS record. What do you think?

docs/public/specifications/tests/MethodsV2.md
marc-vanderwal: The “Handled Servers” set was originally a set of name server IP and query name pairs. Leaving it that way might help avoid duplicate queries.

**EDIT**: I might be wrong. Worded like that, the spec might fail to account for the edge case where multiple parent name servers point to the same IP.

**EDIT 2**: Found the solution, but it’s a little too cumbersome to explain in a review comment.

docs/public/specifications/tests/MethodsV2.md
marc-vanderwal: See above.

docs/public/specifications/tests/MethodsV2.md
matsduf: There are pros and cons with both solutions, but it is probably more helpful to do that (use the NS name). In a sane situation it is not an issue.

docs/public/specifications/tests/MethodsV2.md
matsduf: Or different parent servers have different NS names for the same server (IP address). I will read the specification.

docs/public/specifications/tests/MethodsV2.md
```diff
@@ -176,16 +181,17 @@ This Method uses the following input units defined in section [Methods Inputs]:
...
-       1. For each IP address add the IP address and *Zone Name* to the
-          *Remaining Servers* set unless the IP address is already listed in
-          *Handled Servers* together with *Zone Name*.
+   9.  Do [DNS Lookup] of name server names (A and AAAA) not already listed in
+       the additional section of the response. If a CNAME is encountered,
+       follow the chain of CNAME records but keep the original name server
+       name as obtained from the NS record.
+       1. For each IP address add the name server name, IP address and *Zone
+          Name* tuple to the *Remaining Servers* set unless the IP address is
+          already listed in *Handled Servers* together with *Zone Name*.
```
matsduf: This will result in additional names for the same IP will get lost, won't it?

marc-vanderwal: Yes, I saw the same problem as well.

The solution is to always add name, IP and zone name tuples to *Remaining Servers* without checking *Handled Servers*. Then, right after taking such a tuple out of the set at the beginning of the loop, check *Handled Servers* if the IP and zone name was already processed; if so, then if there is already an entry with the same IP in *Parent Name Servers*, we add it to the *Parent Name Servers*.

This lets us avoid sending duplicate requests for the same (IP, QNAME) tuple while capturing all the names bound to the IP address.

I’ve just pushed a new version of the spec that does this.

marc-vanderwal:
> We aren’t quite there yet, because I realize there are two problems in my latest work:
> 
>     * Adding items to _Remaining Servers_ unconditionally means that we’re adding the same name servers over and over again only to skip over them because the addresses were handled already. There’s got to be a better way.
> 
>     * I don’t think the expected output for the new test scenario PARENT-NS-SAME-IP-2 is correct. Because we merge the delegations from the grandparent and the NS records from the parent zone, we should expect that the parent name servers contains ns1, ns1a, ns1b (all with same IP) and ns2. Currently I only list ns1a, ns1b and ns2.

Fixed.

docs/public/specifications/tests/MethodsV2.md
```diff
@@ -137,55 +139,67 @@ This Method uses the following input units defined in section [Methods Inputs]:
...
+       tuple from *Remaining Servers*.
+   2.  If *Handled Servers* already contains a tuple with the same
+       *Server Address* and *Zone Name* (but not necessarily the same *Server
+       Name*), then:
+       1. If an item exists in *Parent Name Servers* whose IP address equals
+          *Server Address*, then add the *Server Name* and *Server Address*
+          pair to the *Parent Name Servers* set.
+       2. Go to next server in *Remaining Servers*.
+   3.  Insert the *Server Name*, *Server Address* and *Zone Name* tuple into
+       *Handled Servers*.
```
matsduf: Shouldn't (3) be before (2)?

docs/public/specifications/tests/MethodsV2.md
```diff
@@ -137,55 +139,67 @@ This Method uses the following input units defined in section [Methods Inputs]:
...
-       1. For each IP address add the IP address and *Zone Name* to the
-          *Remaining Servers* set unless the IP address is already listed in
-          *Handled Servers* together with *Zone Name*.
+   10. Do [DNS Lookup] of name server names (A and AAAA) not already listed in
+       the additional section of the response. If a CNAME is encountered,
+       follow the chain of CNAME records but keep the original name server
+       name as obtained from the NS record.
+       1. For each IP address add the name server name, IP address and *Zone
+          Name* tuple to the *Remaining Servers* set, unless such a tuple
+          already exists in *Handled Servers*.
```
matsduf, line 198: Could it be even clearer that "the name server name" in (1) should be the name from the NS record RDATA? Maybe "use the original name as obtained from the NS record RDATA when storing the data in next step".

docs/public/specifications/tests/MethodsV2.md
```diff
@@ -216,37 +231,35 @@ This Method uses the following input units defined in section [Methods Inputs]:
...
                 records in the additional section.
              5. Do [DNS Lookup] of name server names (A and AAAA) not already
-                listed in the additional section of the response. Follow CNAME if
-                provided.
-             6. For each IP address add the IP address and *Intermediate Query Name*
-                to the *Remaining Servers* set unless the IP address is already
-                listed in *Handled Servers* together with *Intermediate Query Name*.
+                listed in the additional section of the response. If a CNAME
+                is encountered, follow the chain of CNAME records but keep the
+                original name server name as obtained from the NS record.
```
matsduf: See comment above about CNAME.

docs/public/specifications/tests/MethodsV2.md
```diff
@@ -282,6 +296,83 @@ None.
...
+returned.
+
+Addresses for name servers (RDATA of NS records) are extracted even if the
+resolution goes through CNAME. It is, however, not permitted for a NS record
+to point at a name that has a CNAME, but that test is covered by Test Case
+[Delegation05]. This method should extract as much as possible to find all
+possible paths.
+
+This Method must, in general, use the same algorithm as Test Case [Basic01], but
+the test case extracts more information and outputs messages.
```
matsduf: This text repeats the objective of "Get-Parent-NS-Names-and-IPs". Couldn't it refer to that instead and just state the difference in objective of the two methods?

docs/public/specifications/test-zones/MethodsV2/methodsv2.md
```diff
@@ -644,18 +656,38 @@ Parent is delegated to two IB NS, where both NS names are aliases (CNAME)
...
 
 * child.parent.parent-ns-cname-2.methodsv2.xa
-  * Parent is delegated to `ns1-cname` and `ns2-cname` and both are aliases to 
+  * Parent is delegated to `ns1-cname` and `ns2-cname` and both are aliases to
     `parent-ns1.parent-ns-cname-2.methodsv2.xa` and
     `parent-ns2.parent-ns-cname-2.methodsv2.xa`
   * Both names can be resolved to A and AAAA via CNAME and give correct IP.
 
+### PARENT-NS-SAME-IP-1
+Parent is delegated to two IB NS. The delegation lists multiple name server
```
matsduf: Three IB NS? At least when it comes to names and NS records. "multiple" --> "two"?

docs/public/specifications/test-zones/MethodsV2/methodsv2.md
```diff
@@ -604,7 +616,7 @@ Zone is delegated to two IB NS, where both NS names are aliases (CNAME)
 to other names out of zone.
 
 * child.parent.child-ns-cname-2.methodsv2.xa
```
matsduf:
```suggestion
* Zone: child.parent.child-ns-cname-2.methodsv2.xa
```

Also for CHILD-NS-CNAME-1 above and CHILD-NS-CNAME-3, CHILD-NS-CNAME-4 and PARENT-NS-CNAME-1 below.

docs/public/specifications/test-zones/MethodsV2/methodsv2.md
```diff
@@ -644,18 +656,38 @@ Parent is delegated to two IB NS, where both NS names are aliases (CNAME)
 to other names out of zone.
 
 * child.parent.parent-ns-cname-2.methodsv2.xa
```
matsduf:
```suggestion
* Zone: child.parent.parent-ns-cname-2.methodsv2.xa
```

docs/public/specifications/test-zones/MethodsV2/methodsv2.md
```diff
@@ -644,18 +656,38 @@ Parent is delegated to two IB NS, where both NS names are aliases (CNAME)
...
+  * Parent is delegated to `ns1-cname` and `ns2-cname` and both are aliases to
     `parent-ns1.parent-ns-cname-2.methodsv2.xa` and
     `parent-ns2.parent-ns-cname-2.methodsv2.xa`
   * Both names can be resolved to A and AAAA via CNAME and give correct IP.
 
+### PARENT-NS-SAME-IP-1
+Parent is delegated to two IB NS. The delegation lists multiple name server
+names resolving to the same IP.
+
+* parent.parent-ns-same-ip-1.methodsv2.xa
```
matsduf:
```suggestion
* Zone: parent.parent-ns-same-ip-1.methodsv2.xa
```

docs/public/specifications/test-zones/MethodsV2/methodsv2.md
```diff
@@ -644,18 +656,38 @@ Parent is delegated to two IB NS, where both NS names are aliases (CNAME)
...
+  * Parent is delegated to `ns1a`, `ns1b` and `ns2`.
+  * Both `ns1a` and `ns1b` have A and AAAA records pointing to the same IPv4
+    and IPv6 address respectively.
+
+### PARENT-NS-SAME-IP-2
+Parent is delegated to two IB NS. Two of the in-zone NS records resolve to the
+same IP. The grandparent zone’s delegation lists name server names that are
+different from the in-zone NS record’s, but the sets of IP addresses are equal.
+
+* parent-ns-same-ip-2.methodsv2.xa
```
matsduf:
```suggestion
* Zone: parent-ns-same-ip-2.methodsv2.xa
```

docs/public/specifications/tests/MethodsV2.md
marc-vanderwal: Yes indeed. Otherwise *Handled Servers* is not updated if we take the “true” branch at (2). I’ll swap the steps, but (2) then needs a small change.

docs/public/specifications/tests/MethodsV2.md
marc-vanderwal: Yes. To be perfectly clear, it should say “in the next _substep_” instead of step.

test-zone-data/MethodsV2/parent.parent-ns-same-ip-1.methodsv2.xa.zone
```diff
@@ -0,0 +1,31 @@
...
+@       SOA     ns1 admin. (
+                2025081200      ; serial
+                6h              ; refresh
+                1h              ; retry
+                1w              ; expire
+                1d )            ; minimum
+
+		NS	ns1a
+		NS	ns1b
+                NS      ns2
```
tgreenx:
```suggestion
		NS	ns1b
        NS  ns2
```

test-zone-data/MethodsV2/parent.parent-ns-same-ip-2.methodsv2.xa.zone
```diff
@@ -0,0 +1,31 @@
...
+@       SOA     ns1 admin. (
+                2025081200      ; serial
+                6h              ; refresh
+                1h              ; retry
+                1w              ; expire
+                1d )            ; minimum
+
+		NS	ns1a
+		NS	ns1b
+                NS      ns2
```
tgreenx:
```suggestion
		NS	ns1b
        NS  ns2
```

test-zone-data/MethodsV2/parent.parent-ns-same-ip-1.methodsv2.xa.zone
```diff
@@ -0,0 +1,31 @@
...
+ns2 		A	127.40.1.42
+ns2 		AAAA	fda1:b2:c3:0:127:40:1:42
+
+$ORIGIN child            ; Must be relative
+@		NS	ns1
+@		NS	ns2
+ns1	A	127.40.1.51
+ns1	AAAA	fda1:b2:c3:0:127:40:1:51
+ns2	A	127.40.1.52
+ns2	AAAA	fda1:b2:c3:0:127:40:1:52
```
tgreenx:
```suggestion
@		NS	ns2
ns1	    A	127.40.1.51
ns1	    AAAA	fda1:b2:c3:0:127:40:1:51
ns2	    A	127.40.1.52
ns2	    AAAA	fda1:b2:c3:0:127:40:1:52
```

test-zone-data/MethodsV2/parent.parent-ns-same-ip-2.methodsv2.xa.zone
```diff
@@ -0,0 +1,31 @@
...
+ns2 		A	127.40.1.42
+ns2 		AAAA	fda1:b2:c3:0:127:40:1:42
+
+$ORIGIN child            ; Must be relative
+@		NS	ns1
+@		NS	ns2
+ns1	A	127.40.1.51
+ns1	AAAA	fda1:b2:c3:0:127:40:1:51
+ns2	A	127.40.1.52
+ns2	AAAA	fda1:b2:c3:0:127:40:1:52
```
tgreenx:
```suggestion
@		NS	ns2
ns1	    A	127.40.1.51
ns1	    AAAA	fda1:b2:c3:0:127:40:1:51
ns2	    A	127.40.1.52
ns2	    AAAA	fda1:b2:c3:0:127:40:1:52
```

tgreenx (dismissed):
LGTM. Small indentation nits

test-zone-data/MethodsV2/parent.parent-ns-same-ip-1.methodsv2.xa.zone
marc-vanderwal: Thanks. I reindented the file differently from your suggestions, but I made sure to convert it so that everything is indented with spaces only.

docs/public/specifications/test-zones/MethodsV2/methodsv2.md
```diff
@@ -643,19 +655,39 @@ to other names in parent zone.
...
+  * Parent is delegated to `ns1-cname` and `ns2-cname` and both are aliases to
     `parent-ns1.parent-ns-cname-2.methodsv2.xa` and
     `parent-ns2.parent-ns-cname-2.methodsv2.xa`
   * Both names can be resolved to A and AAAA via CNAME and give correct IP.
 
+### PARENT-NS-SAME-IP-1
+Parent is delegated to three IB NS. The delegation lists two name server names
+resolving to the same IP.
+
+* Zone: parent.parent-ns-same-ip-1.methodsv2.xa
```
matsduf: The zone that we test is always the child zone, i.e. the lowest zone in the chain. The name could be anything, but it would be more consistent if it is `child.parent-ns-same-ip-1.methodsv2.xa` and `parent-ns-same-ip-1.methodsv2.xa` is the parent zone. But here it sounds like `child.parent.parent-ns-same-ip-1.methodsv2.xa` should be the child zone and `parent.parent-ns-same-ip-1.methodsv2.xa` is the parent zone. And `parent-ns-same-ip-1.methodsv2.xa` the grandparent zone.

docs/public/specifications/test-zones/MethodsV2/methodsv2.md
```diff
@@ -643,19 +655,39 @@ to other names in parent zone.
...
+Parent is delegated to two IB NS. Two of the in-zone NS records resolve to the
+same IP. The grandparent zone’s delegation lists name server names that are
+different from the in-zone NS record’s, but the sets of IP addresses are equal.
+
+* Zone: parent-ns-same-ip-2.methodsv2.xa
+  * Parent is delegated to `ns1` and `ns2`.
+* Zone: parent.parent-ns-same-ip-2.methodsv2.xa
+  * In-zone NS records list `ns1a`, `ns1b` and `ns2`.
+  * Both `ns1a` and `ns1b` have A and AAAA records pointing to the same IPv4
+    and IPv6 address respectively.
```
matsduf: There should only be one (child) zone, i.e. the zone that we enter for testing.

docs/public/specifications/test-zones/MethodsV2/methodsv2.md
marc-vanderwal: The zone we’re testing in these scenarios is still `child.parent.parent-ns-same-ip-…` and that one stays in its default configuration, so there is no bullet point for the child zone. It is only in `parent.parent-ns-same-ip-…` that we introduce problems.

But your comment highlights a different problem: for each scenario, the zone under test is not explicitly specified. Without an explicit comment to that effect in the description of a scenario, the reader is misled into assuming that the zone under test is the first bullet point. That’s a problem with the template, I think.

docs/public/specifications/test-zones/MethodsV2/methodsv2.md
marc-vanderwal: We do enter only one child zone, but it is left in its default configuration. Like the previous scenario, the zone under test is still `child.parent.parent-ns-same-ip-…`.

docs/public/specifications/test-zones/MethodsV2/methodsv2.md
matsduf: The "zone:" statement should be the zone to be tested, "child.parent-ns-same-ip-1.methodsv2.xa" in this case. I see that in the defaults it is already stated for most (or maybe all scenarios) but for creating the unit test you need to have scenario name and correct zone to be tested.

docs/public/specifications/test-zones/MethodsV2/methodsv2.md
matsduf: The "zone:" statement should be the zone to be tested.

marc-vanderwal:
> > This has yet to be implemented in Engine.
> 
> Correct. Maybe we were too fast merge.

No, actually it’s fine. When that specification was stabilized, I went ahead and did the implementation as well in order to save time later. I have just rebased my work, did some finishing touches and created zonemaster/zonemaster-engine#1475.

PR #1444, opened 2025-11-27
Fix links

mattias-p:
## Purpose

Fix errors and warnings reported by `mdbook build`. I'm using:

 * mdbook v0.4.52
 * mdbook-linkcheck 0.7.7

## Context

The problems fixed here were introduced since last release, so I don't think they should be mentioned in the changelog.

## Changes

I made some best-effort fixes to pacify mdbook-linkcheck.

## How to review this PR

Verify that the fixes match the intention.


docs/public/specifications/test-zones/Zone-TP/zone11.md
```diff
@@ -90,7 +90,7 @@ irrelevant to the test scenario and must be ignored.
 * (1) All tags except for those specified as "Forbidden message tags" (no instances for these test scenarios)
 * (2) All tags except for those specified as "Mandatory message tags"
 
-## Test scenarios and setup of test zones
+## Zone setup for test scenarios
```
matsduf: The section actually defines the scenario features, so the existing header is probably better. Also used in many test scenario specifications.

mattias-p: Fixed.

PR #1477, opened 2026-04-07
Update CNAME test scenarios

tgreenx:
## Purpose

This PR updates test scenarios related to CNAME. See the `Changes` section below for more details.

## Context

N/A

## Changes

- Move content of scenario `TOO-LONG-CNAME-CHAIN` to a new scenario `TOO-MANY-CNAME`
- Update scenario `TOO-LONG-CNAME-CHAIN` to test for abnormally long chained CNAME lookups (tag `CNAME_CHAIN_TOO_LONG`)
- Add scenario `UNRESOLVABLE-CNAME`
- Add scenario `CNAME-CHAIN-TO-NODATA`

## How to test this PR

N/A. Review.

### DIFF
--- a/docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
+++ b/docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
@@ -47,8 +47,9 @@ TOO-LONG-CNAME-CHAIN         | Undefined and tags `CNAME_START`, `CNAME_CHAIN_TO
 TOO-MANY-CNAME               | Undefined and tags `CNAME_START`, `CNAME_RECORDS_TOO_MANY`
 TARGET-NO-MATCH-CNAME        | Undefined and tags `CNAME_START`, `CNAME_NO_MATCH`
 BROKEN-CNAME-CHAIN           | Undefined and tags `CNAME_START`, `CNAME_RECORDS_CHAIN_BROKEN`
-WRONG-CNAME-OWNER-NAME       | False and no tags
-EXTRA-CNAME-IN-ANSWER        | False and no tags
+WRONG-CNAME-OWNER-NAME       | True and no tags
+EXTRA-CNAME-IN-ANSWER        | True and no tags
+CNAME-CHAIN-TO-NODATA        | True and tags `CNAME_START`, `CNAME_TO_NODATA`
 UNRESOLVABLE-CNAME           | Undefined and tags `CNAME_START`, `CNAME_UNRESOLVABLE`
 
 ## Zone setup for test scenarios
@@ -279,12 +280,14 @@ above the limit.
 
 ### TARGET-NO-MATCH-CNAME
 The CNAME target name does not match the owner name of the `A` record.
+Note that this scenario also purposefully use varying case in names for
+additionnal testing purposes, but it is not strictly needed.
 
 * Query name: "target-no-match-cname.cname.recursor.engine.xa"
   * To be found in the answer section:
 ```
-   target-no-match-cname         CNAME  target-no-match-cname-two
-   target-no-match-cname-target  A      127.0.0.1
+   TARGET-no-match-cname         CNAME  target-NO-match-cname-two
+   target-no-MATCH-cname-target  A      127.0.0.1
 ```
 
 ### BROKEN-CNAME-CHAIN
@@ -333,10 +336,12 @@ cname-chain-to-nodata            CNAME target.cname-chain-to-nodata
     gives a NODATA response
 
 ### UNRESOLVABLE-CNAME
-The target in the CNAME record can not be resolved because
-of an issue with the zone.
-Currently this scenario considers a FORMERR response to an
-in-bailiwick CNAME target name.
+The target in the CNAME record can not be resolved due to a resolution failure
+with the name servers of the zone.
+This scenario involves a FORMERR response to a query for the target of a CNAME
+resource record when the target is a subdomain of the resource record’s owner name.
+It purposefully considers a FORMERR response (as opposed to e.g. SERVFAIL) as this is
+required to trigger the behavior in Zonemaster's recursive lookup implementation.
 
 * Query name: "unresolvable-cname.cname.recursor.engine.xa"
   * To be found in the answer section:
--- a/docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
+++ b/docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
@@ -43,11 +43,13 @@ LOOPED-CNAME-IN-ZONE-1       | Undefined and tags `CNAME_START`, `CNAME_LOOP_INN
 LOOPED-CNAME-IN-ZONE-2       | Undefined and tags `CNAME_START`, `CNAME_LOOP_INNER`
 LOOPED-CNAME-IN-ZONE-3       | Undefined and tags `CNAME_START`, `CNAME_LOOP_INNER`
 LOOPED-CNAME-OUT-OF-ZONE     | Undefined and tags `CNAME_START`, `CNAME_LOOP_OUTER`
-TOO-LONG-CNAME-CHAIN         | Undefined and tags `CNAME_START`, `CNAME_RECORDS_TOO_MANY`
+TOO-LONG-CNAME-CHAIN         | Undefined and tags `CNAME_START`, `CNAME_CHAIN_TOO_LONG`
+TOO-MANY-CNAME               | Undefined and tags `CNAME_START`, `CNAME_RECORDS_TOO_MANY`
 TARGET-NO-MATCH-CNAME        | Undefined and tags `CNAME_START`, `CNAME_NO_MATCH`
 BROKEN-CNAME-CHAIN           | Undefined and tags `CNAME_START`, `CNAME_RECORDS_CHAIN_BROKEN`
 WRONG-CNAME-OWNER-NAME       | False and no tags
 EXTRA-CNAME-IN-ANSWER        | False and no tags
+UNRESOLVABLE-CNAME           | Undefined and tags `CNAME_START`, `CNAME_UNRESOLVABLE`
 
 ## Zone setup for test scenarios
 
@@ -249,23 +251,29 @@ and the target name of the second CNAME record will point at the first.
 ```
 
 ### TOO-LONG-CNAME-CHAIN
+The query name will follow a CNAME chain and reach the maximum lookup limit
+in different, unique zones.
+
+* Query name: "too-long-cname-chain.sub2.cname.recursor.engine.xa"
+
+### TOO-MANY-CNAME
 The query name will resolve to one `A` record via ten CNAME records which is
 above the limit.
 
-* Query name: "too-long-cname-chain.cname.recursor.engine.xa"
+* Query name: "too-many-cname.cname.recursor.engine.xa"
   * To be found in the answer section:
 ```
-   too-long-cname-chain              CNAME too-long-cname-chain-two
-   too-long-cname-chain-two          CNAME too-long-cname-chain-three
-   too-long-cname-chain-three        CNAME too-long-cname-chain-four
-   too-long-cname-chain-four         CNAME too-long-cname-chain-five
-   too-long-cname-chain-five         CNAME too-long-cname-chain-six
-   too-long-cname-chain-six          CNAME too-long-cname-chain-seven
-   too-long-cname-chain-seven        CNAME too-long-cname-chain-eight
-   too-long-cname-chain-eight        CNAME too-long-cname-chain-nine
-   too-long-cname-chain-nine         CNAME too-long-cname-chain-ten
-   too-long-cname-chain-ten          CNAME too-long-cname-chain-target
-   too-long-cname-chain-target       A     127.0.0.1
+   too-many-cname              CNAME too-many-cname-two
+   too-many-cname-two          CNAME too-many-cname-three
+   too-many-cname-three        CNAME too-many-cname-four
+   too-many-cname-four         CNAME too-many-cname-five
+   too-many-cname-five         CNAME too-many-cname-six
+   too-many-cname-six          CNAME too-many-cname-seven
+   too-many-cname-seven        CNAME too-many-cname-eight
+   too-many-cname-eight        CNAME too-many-cname-nine
+   too-many-cname-nine         CNAME too-many-cname-ten
+   too-many-cname-ten          CNAME too-many-cname-target
+   too-many-cname-target       A     127.0.0.1
 ```
 
 ### TARGET-NO-MATCH-CNAME
@@ -312,6 +320,11 @@ besides the `A` record matching query name.
    extra-cname-in-answer-1       CNAME extra-cname-in-answer-2
 ```
 
+### UNRESOLVABLE-CNAME
+The target in the CNAME record can not be resolved because
+of an unidentified issue with the zone.
+
+* Query name: "unresolvable-cname.cname.recursor.engine.xa"
 
 
 [RCODE Name]:                                                     https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-6
--- a/docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
+++ b/docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
@@ -320,17 +320,30 @@ besides the `A` record matching query name.
    extra-cname-in-answer-1       CNAME extra-cname-in-answer-2
 ```
 
+### CNAME-CHAIN-TO-NODATA
+There is a NODATA response for an in-bailiwick CNAME target name.
+
+* Query name: "unresolvable-cname.cname.recursor.engine.xa"
+  * To be found in the answer section:
+```
+cname-chain-to-nodata            CNAME target.cname-chain-to-nodata
+```
+  * The CNAME target "target.cname-chain-to-nodata..cname.recursor.engine.xa"
+    gives a NODATA response
+
 ### UNRESOLVABLE-CNAME
 The target in the CNAME record can not be resolved because
 of an issue with the zone.
+Currently this scenario considers a FORMERR response to an
+in-bailiwick CNAME target name.
 
 * Query name: "unresolvable-cname.cname.recursor.engine.xa"
   * To be found in the answer section:
 ```
 unresolvable-cname            CNAME target.unresolvable-cname
-*.target.unresolvable-cname   TXT   "Breaks"
 ```
-
+  * The CNAME target "target.unresolvable-cname.cname.recursor.engine.xa"
+    gives a FORMERR response
 
 [RCODE Name]:                                                     https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-6
 [Test zone README file]:                                          ../../README.md
--- a/docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
+++ b/docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
@@ -63,6 +63,7 @@ specific scenario:
 * In the zone, the query name always has a `CNAME` record.
 * The zone is set up on one NS, ns1.
 * The [RCODE Name] in the response is NoError.
+* The maximum lookup limit is set to 10.
 
 ### GOOD-CNAME-1
 The query name will resolve to one `A` record via one CNAME.
--- a/docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
+++ b/docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
@@ -322,9 +322,14 @@ besides the `A` record matching query name.
 
 ### UNRESOLVABLE-CNAME
 The target in the CNAME record can not be resolved because
-of an unidentified issue with the zone.
+of an issue with the zone.
 
 * Query name: "unresolvable-cname.cname.recursor.engine.xa"
+  * To be found in the answer section:
+```
+unresolvable-cname            CNAME target.unresolvable-cname
+*.target.unresolvable-cname   TXT   "Breaks"
+```
 
 
 [RCODE Name]:                                                     https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-6
--- a/test-zone-data/Engine/Recursor-PM/recursor.cfg
+++ b/test-zone-data/Engine/Recursor-PM/recursor.cfg
@@ -45,26 +45,26 @@ cname.recursor.engine.xa:53 {
    }
 
    # For scenario LOOPED-CNAME-IN-ZONE-2
-   template IN ANY looped-cname-in-zone-2.cname.recursor.engine.xa {
+   template IN A looped-cname-in-zone-2.cname.recursor.engine.xa {
       answer     "looped-cname-in-zone-2.cname.recursor.engine.xa. 3600 IN CNAME looped-cname-in-zone-2-a.cname.recursor.engine.xa."
       answer     "looped-cname-in-zone-2-a.cname.recursor.engine.xa. 3600	IN CNAME looped-cname-in-zone-2-b.cname.recursor.engine.xa."
       answer     "looped-cname-in-zone-2-b.cname.recursor.engine.xa. 3600	IN CNAME looped-cname-in-zone-2-a.cname.recursor.engine.xa."
       authority  "cname.recursor.engine.xa. 3600	IN	NS	ns1.cname.recursor.engine.xa."
     }
 
-   # For scenario TOO-LONG-CNAME-CHAIN
-   template IN A too-long-cname-chain.cname.recursor.engine.xa {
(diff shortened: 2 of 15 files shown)


docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
```diff
@@ -312,6 +320,11 @@ besides the `A` record matching query name.
    extra-cname-in-answer-1       CNAME extra-cname-in-answer-2
 ```
 
+### UNRESOLVABLE-CNAME
+The target in the CNAME record can not be resolved because
+of an unidentified issue with the zone.
```
marc-vanderwal: You want a CNAME resource record pointing to a name that cannot resolve because all authoritative name servers respond with SERVFAIL or something like that?

test-zone-data/Engine/Recursor-PM/recursor.cfg
```diff
@@ -45,26 +45,26 @@ cname.recursor.engine.xa:53 {
    }
 
    # For scenario LOOPED-CNAME-IN-ZONE-2
-   template IN ANY looped-cname-in-zone-2.cname.recursor.engine.xa {
+   template IN A looped-cname-in-zone-2.cname.recursor.engine.xa {
```
marc-vanderwal: It should be `ANY`, or else we break RFC 1035. An authoritative server should always answer with CNAME for a name that is aliased, regardless of the query type.

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
tgreenx: No. This scenario should capture an edge case where we reach the end of the [`_resolve_cname()`](https://github.com/tgreenx/zonemaster-engine/blob/797964ef0f8979be1bcaf64038c7bd086f9df8f8/lib/Zonemaster/Engine/Recursor.pm#L162-L273) method, despite all the checks and safeguards already in place. It's a path in the code that should not be reachable normally, but it's there as another safety measure. So far I couldn't think of any such edge case.

marc-vanderwal: After looking carefully, I’ve found an edge case that triggers a `CNAME_UNRESOLVABLE` message.

If a name is an alias to a subdomain of itself and no resource record of the requested type exists at the CNAME’s target, you’ll get the message.

Here’s an example. Consider the following zone file for the `test` zone:

```zone
$ORIGIN test.
;; SOA records, NS records, etc.

break-me                IN  CNAME  sub.break-me
;; sub.break-me is an empty non-terminal
*.sub.break-me          IN  AAAA   ::
```
Then, attempting `Zonemaster::Engine::Recursor->recurse("break-me.test")` triggers `CNAME_UNRESOLVABLE`.

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
```diff
@@ -249,23 +251,29 @@ and the target name of the second CNAME record will point at the first.
 ```
 
 ### TOO-LONG-CNAME-CHAIN
+The query name will follow a CNAME chain and reach the maximum lookup limit
+in different, unique zones.
```
matsduf, line 256: This means that there is a limit set in Zonemaster. I suggest that there is a reference to that here,

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
```diff
@@ -312,6 +320,16 @@ besides the `A` record matching query name.
    extra-cname-in-answer-1       CNAME extra-cname-in-answer-2
 ```
 
+### UNRESOLVABLE-CNAME
+The target in the CNAME record can not be resolved because
+of an issue with the zone.
+
+* Query name: "unresolvable-cname.cname.recursor.engine.xa"
+  * To be found in the answer section:
```
matsduf, line 342: Is there any difference between the following cases assuming that the query is for "target.unresolvable-cname. A"

```
unresolvable-cname            CNAME target.unresolvable-cname
*.target.unresolvable-cname   TXT   "Breaks"
```

```
unresolvable-cname            CNAME target.unresolvable-cname
target.unresolvable-cname   TXT   "Breaks"
```

```
unresolvable-cname            CNAME target.unresolvable-cname
sub.target.unresolvable-cname   TXT   "Breaks"
```

```
unresolvable-cname            CNAME target.unresolvable-cname
```

```
unresolvable-cname            CNAME target.unresolvable-cname
*.target.unresolvable-cname   TXT   "Breaks"
```

```
unresolvable-cname            CNAME target.unresolvable-cname.
target.unresolvable-cname.local   TXT   "Breaks"
```

matsduf:
>     * Add scenario `UNRESOLVABLE-CNAME` but it is currently undefined
>       _-> I could not come up with a reasonable scenario that would lead to this outcome_

It is quite common to have a lookup involving CNAME but not having any final record of the requested type. Or what do you mean?

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
marc-vanderwal: Not really. The idea of the scenario is to have a CNAME chain end up at a final query that gives NODATA.

That’s also why I take issue with the scenario’s name. There is nothing wrong with the CNAME chain itself but we simply end up with a slightly different kind of negative response. Instead of `UNRESOLVABLE-CNAME`, this scenario should be called `CNAME-CHAIN-TO-NODATA` and the text be updated accordingly.

matsduf: Maybe two scenarios, `CNAME-CHAIN-TO-NODATA` and `CNAME-CHAIN-TO-NXDOMAIN`.

What is special here is that the answer section contains an unrelated record. Only records matching query name or the CNAME target (CNAME RDATA) are expected here. What is the point of the unrelated record? Do we expect Zonemaster to react?

tgreenx: `CNAME-CHAIN-TO-NODATA` scenario added. And `CNAME-CHAIN-TO-NXDOMAIN` is not needed, NXDOMAIN responses are already handled by the calling method (`Zonemaster::Engine::Recursor::_recurse()`).

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
tgreenx: Done. The limit is currently hardcoded in the `Constants.pm` module in Zonemaster-Engine (`CNAME_MAX_CHAIN_LENGTH`).

tgreenx:
@matsduf @marc-vanderwal comments have been addressed, [implementation PR](https://github.com/zonemaster/zonemaster-engine/pull/1518) is updated, please re-review.

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
```diff
@@ -320,17 +320,30 @@ besides the `A` record matching query name.
...
+```
+cname-chain-to-nodata            CNAME target.cname-chain-to-nodata
+```
+  * The CNAME target "target.cname-chain-to-nodata..cname.recursor.engine.xa"
+    gives a NODATA response
+
 ### UNRESOLVABLE-CNAME
 The target in the CNAME record can not be resolved because
 of an issue with the zone.
```
matsduf, line 336: Is it really correct to call that an issue? Say that you have the following in a zone:
```
www.namn.se.    CNAME  webb.namn.se.
webb.namn.se.   A            127.10.1.0
```
A query for "www.namn.se. A" returns an A record, but a query for "www.namn.se. AAAA" return NODATA. The zone is fine. There is no issue with the zone.

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
```diff
@@ -320,17 +320,30 @@ besides the `A` record matching query name.
...
+Currently this scenario considers a FORMERR response to an
+in-bailiwick CNAME target name.
 
 * Query name: "unresolvable-cname.cname.recursor.engine.xa"
   * To be found in the answer section:
```
matsduf, line 341: You said that "To be found in the answer section" above should be "To be found in the zone".

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
```diff
@@ -320,17 +320,30 @@ besides the `A` record matching query name.
...
 ```
 unresolvable-cname            CNAME target.unresolvable-cname
-*.target.unresolvable-cname   TXT   "Breaks"
 ```
-
+  * The CNAME target "target.unresolvable-cname.cname.recursor.engine.xa"
+    gives a FORMERR response
```
matsduf, line 346: Is that to test what happens when the RCODE is FORMERR? FORMERR is something that the server sends out when it cannot handle the format of the query. And then the DNS record sections are empty, except for the query section.

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
```diff
@@ -312,7 +321,30 @@ besides the `A` record matching query name.
...
+  * To be found in the answer section:
+```
+cname-chain-to-nodata            CNAME target.cname-chain-to-nodata
+```
+  * The CNAME target "target.cname-chain-to-nodata..cname.recursor.engine.xa"
+    gives a NODATA response
+
+### UNRESOLVABLE-CNAME
+The target in the CNAME record can not be resolved because
+of an issue with the zone.
```
marc-vanderwal, line 337: It isn’t an issue with the zone, but something that [RFC 9520](https://www.rfc-editor.org/rfc/rfc9520.html) calls a resolution failure. Instead, it’s an issue with the name server.

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
marc-vanderwal: More generally, it’s meant to test resilience against any kind of resolution failures, mainly those due to server failures. The RCODE could be SERVFAIL or FORMERR or something else, but I don’t think that matters a lot. SERVFAIL is more common, though.

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
```diff
@@ -312,7 +321,30 @@ besides the `A` record matching query name.
...
+Currently this scenario considers a FORMERR response to an
+in-bailiwick CNAME target name.
```
marc-vanderwal, line 339:
```suggestion
This scenario involves a FORMERR response to a query for the target of a CNAME resource record
when the target is a subdomain of the resource record’s owner name.
```

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
```diff
@@ -43,11 +43,13 @@ LOOPED-CNAME-IN-ZONE-1       | Undefined and tags `CNAME_START`, `CNAME_LOOP_INN
...
 LOOPED-CNAME-IN-ZONE-3       | Undefined and tags `CNAME_START`, `CNAME_LOOP_INNER`
 LOOPED-CNAME-OUT-OF-ZONE     | Undefined and tags `CNAME_START`, `CNAME_LOOP_OUTER`
-TOO-LONG-CNAME-CHAIN         | Undefined and tags `CNAME_START`, `CNAME_RECORDS_TOO_MANY`
+TOO-LONG-CNAME-CHAIN         | Undefined and tags `CNAME_START`, `CNAME_CHAIN_TOO_LONG`
+TOO-MANY-CNAME               | Undefined and tags `CNAME_START`, `CNAME_RECORDS_TOO_MANY`
 TARGET-NO-MATCH-CNAME        | Undefined and tags `CNAME_START`, `CNAME_NO_MATCH`
 BROKEN-CNAME-CHAIN           | Undefined and tags `CNAME_START`, `CNAME_RECORDS_CHAIN_BROKEN`
 WRONG-CNAME-OWNER-NAME       | False and no tags
 EXTRA-CNAME-IN-ANSWER        | False and no tags
+UNRESOLVABLE-CNAME           | Undefined and tags `CNAME_START`, `CNAME_UNRESOLVABLE`
```
marc-vanderwal, line 52: Scenario `CNAME-CHAIN-TO-NODATA` is newly introduced, but missing in this table.

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
matsduf: That is fine, but is that connected CNAME in some sense?

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
matsduf: In a weird case it could be that the CNAME points to a name in the same zone, but that part of the zone is served in such a way that you get a SERVFAIL. Else the target name could be in another zone on the same server, but that zone cannot be loaded.

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
matsduf: What does "a query for the target of a CNAME resource record"? The the query section you have a query name that could match an owner name with a CNAME record in the zone. Is that to query for the target? Or you can get a CNAME record from one query of an authoritative server, and the target points exists in another zone, and you send a new query to that zone's server. Is that to query for the target?

marc-vanderwal: Yes, it’s the name pointed to by the CNAME resource record. The domain name that appears in the CNAME resource record’s RDATA.

matsduf: Right, but it is not part of the query. It comes from the server. I think it is farfetched that a server sets FORMERR in the response (meaning that the query had a format error) because the server does not like the target name of the CNAME, which was not in the query but comes with the response (from the server). Here SERVFAIL would be more likely.

If a server could respond FORMERR in such a case, then it could respond FORMERR on a query for an A record because the A record that the server serves has a private IP address in RDATA.

I do not see the connection to CNAME in this scenario.

marc-vanderwal: So in a nutshell, you’d rather see a SERVFAIL response in this scenario?

Yes, the scenario is far-fetched, but so have been the problems we’ve seen in the wild that have caused issues in Zonemaster before.

matsduf: It would be good to make sure that Zonemaster handles strange responses in different situations. If you get a FORMERR or SERVFAIL as RCODE then you should ignore anything in the answer, authority or additional section besides OPT record in additional. In this scenario you do not query for CNAME so if  the server adds a CNAME or A record to the answer section is irrelevant. You have to handle the FORMERR or SERVFAIL RCODE in a reasonable way.

The scenario is fine, but not under a CNAME umbrella.

marc-vanderwal: > The scenario is fine, but not under a CNAME umbrella.

I disagree: a normal resolution failure (e.g. getting directly SERVFAIL in response to an initial query) is one thing; but getting a resolution failure in a subordinate query (e.g. query to name A gives a CNAME to B, then the query to B gives SERVFAIL) is likely to be a different code path. We need to test both. This scenario makes perfect sense here.

matsduf: I do not understand what you mean. Here is the text from the scenario specification:
````
### UNRESOLVABLE-CNAME
The target in the CNAME record can not be resolved because
of an issue with the zone.
Currently this scenario considers a FORMERR response to an
in-bailiwick CNAME target name.

* Query name: "unresolvable-cname.cname.recursor.engine.xa"
  * To be found in the answer section:
```
unresolvable-cname            CNAME target.unresolvable-cname
```
  * The CNAME target "target.unresolvable-cname.cname.recursor.engine.xa"
    gives a FORMERR response
````

Zonemaster sends a query QNAME=`unresolvable-cname.cname.recursor.engine.xa` and some QTYPE not specified but it must be something other than CNAME for CNAME to be special, say QTYPE=`TXT`.

The server responds with RCODE=`FORMERR` and with the following record in the answer section. 
```
unresolvable-cname.cname.recursor.engine.xa.  CNAME target.unresolvable-cname.cname.recursor.engine.xa.
```
In such a case the answer section should be ignored. Zonemaster should send any more queries for this case. What do you want to test? Do you want to test if Zonemaster erroneously tries to look the CNAME target up?

Possible variants of the scenario:
* Answer section is empty.
* The RRtype in the answer section is TXT or A.
* The owner name of the RR in the answer section is `something.xb`.
* RCODE is REFUSED or SERVFAIL.

Comment (marc-vanderwal) [reply docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md:339]: The point of this scenario is to test whether Zonemaster correctly returns a resolution failure and the appropriate message in case a query that tries to follow a CNAME gets a response that is neither a positive response, nor a negative response, but something else that indicates a resolution failure.

The query for unresolvable-cname.recursor.engine.xa should give RCODE NOERROR and the following resource record in the answer section:
```zone
unresolvable-cname.recursor.engine.xa.  IN  CNAME target.unresolvable-cname.recursor.engine.xa.
```

This, in turn, should trigger a subordinate query for target.unresolvable-cname.recursor.engine.xa. It’s that query that should yield a response with some RCODE that is neither NOERROR nor NXDOMAIN. The spec here says FORMERR, but SERVFAIL would work just as well. The answer, authority and additional sections should be empty.

matsduf: > The point of this scenario is to test whether Zonemaster correctly returns a resolution failure and the appropriate message in case a query that tries to follow a CNAME gets a response that is neither a positive response, nor a negative response, but something else that indicates a resolution failure.

That is OK.

> The query for unresolvable-cname.recursor.engine.xa should give RCODE NOERROR and the following resource record in the answer section:
> 
> ```
> unresolvable-cname.recursor.engine.xa.  IN  CNAME target.unresolvable-cname.recursor.engine.xa.
> ```

Query for unresolvable-cname.recursor.engine.xa and what QTYPE? That is important, CNAME or something else? Let's assume TXT.

If there is no record for the target name, there are two well-formed responses that require some records in the authority section.

* NODATA. AA-bit is set, and the SOA record is in the authority section.
* Referral: AA-bit is unset, and the authority section contains the NS records for the referral. And possibly even glue records.

If NODATA, there is no reason to requery. If referral, send query to the subzone NS. If neither, report error.

We could also change the target name to be something outside the zone, e.g. something.xb, and then there should be no records in the authority section.

> This, in turn, should trigger a subordinate query for target.unresolvable-cname.recursor.engine.xa. It’s that query that should yield a response with some RCODE that is neither NOERROR nor NXDOMAIN. The spec here says FORMERR, but SERVFAIL would work just as well. The answer, authority and additional sections should be empty.

This description makes more sense, but it must be clarified if it should be a NODATA or referral case. Or even the case with error, but then requerying is not possible. Maybe split into three scenarios?

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
tgreenx: Fixed.

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
tgreenx: It's a FORMERR response on purpose so that Zonemaster's implementation can reach that code path. I've added a comment.

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
```diff
@@ -249,33 +253,41 @@ and the target name of the second CNAME record will point at the first.
...
+   too-many-cname-eight        CNAME too-many-cname-nine
+   too-many-cname-nine         CNAME too-many-cname-ten
+   too-many-cname-ten          CNAME too-many-cname-target
+   too-many-cname-target       A     127.0.0.1
 ```
 
 ### TARGET-NO-MATCH-CNAME
 The CNAME target name does not match the owner name of the `A` record.
+Note that this scenario also purposefully use varying case in names for
+additionnal testing purposes, but it is not strictly needed.
```
matsduf, line 284: I suggest that those different testing purposes are given dedicated scenarios.

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
```diff
@@ -312,7 +324,32 @@ besides the `A` record matching query name.
    extra-cname-in-answer-1       CNAME extra-cname-in-answer-2
 ```
 
+### CNAME-CHAIN-TO-NODATA
+There is a NODATA response for an in-bailiwick CNAME target name.
+
+* Query name: "unresolvable-cname.cname.recursor.engine.xa"
```
matsduf, line 330: The query name here is the same the query name in the next scenario. Is that on purpose? Or could some other name be used, e.g. unresolvable.cname-chain-to-nodata.cname.recursor.engine.xa?

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
```diff
@@ -312,7 +324,32 @@ besides the `A` record matching query name.
...
+  * To be found in the answer section:
+```
+cname-chain-to-nodata            CNAME target.cname-chain-to-nodata
+```
+  * The CNAME target "target.cname-chain-to-nodata..cname.recursor.engine.xa"
```
matsduf, line 335:
```suggestion
  * The CNAME target "target.cname-chain-to-nodata.cname.recursor.engine.xa"
```

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
```diff
@@ -312,7 +324,32 @@ besides the `A` record matching query name.
...
+    gives a NODATA response
```
matsduf, line 336: The question is if that is a NODATA response. The owner name of CNAME in the answer section mismatches the query name.

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
```diff
@@ -312,7 +324,32 @@ besides the `A` record matching query name.
...
+It purposefully considers a FORMERR response (as opposed to e.g. SERVFAIL) as this is
+required to trigger the behavior in Zonemaster's recursive lookup implementation.
 
+* Query name: "unresolvable-cname.cname.recursor.engine.xa"
+  * To be found in the answer section:
+```
+unresolvable-cname            CNAME target.unresolvable-cname
+```
+  * The CNAME target "target.unresolvable-cname.cname.recursor.engine.xa"
+    gives a FORMERR response
```
matsduf, line 352: FORMERR response for any record type?

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
```diff
@@ -312,7 +324,32 @@ besides the `A` record matching query name.
...
+cname-chain-to-nodata            CNAME target.cname-chain-to-nodata
+```
+  * The CNAME target "target.cname-chain-to-nodata..cname.recursor.engine.xa"
+    gives a NODATA response
+
+### UNRESOLVABLE-CNAME
+The target in the CNAME record can not be resolved due to a resolution failure
+with the name servers of the zone.
+This scenario involves a FORMERR response to a query for the target of a CNAME
+resource record when the target is a subdomain of the resource record’s owner name.
```
matsduf, line 342: Subdomain could mean either the same zone or a subzone.

matsduf (commented):
Scenario NXDOMAIN-VIA-CNAME should retuern NXDOMAIN but it does not according to README.md.

Maybe scenarios LOOPED-CNAME-IN-ZONE-1 and LOOPED-CNAME-IN-ZONE-3 can be implemented using template in the cfg file?

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
tgreenx: No, typo. Fixed to `cname-chain-to-nodata.cname.recursor.engine.xa`.

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
tgreenx: Fixed

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
tgreenx: Query name was wrong (see above), fixed.

docs/public/specifications/test-zones/Engine/Recursor-PM/CNAME.md
tgreenx: Right now it's just for A, but it could be any. Updated the scenario description.

matsduf (approved):

PR #1488, opened 2026-06-01
Updates Basic01 and MethodsV2 to handle RFC 8020 non-compliant servers

### LINKED ISSUE #1484: Reverse zone cannot be tested (Basic01)
`zonemaster-cli` says that `0.d.b.9.1.b.9.0.1.0.0.2.ip6.arpa` does not exist as a zone:

```
Seconds Level    Testcase       Message
======= ======== ============== =======
   0.00 INFO     Unspecified    Using version v8.1.1 of the Zonemaster engine.
  25.74 INFO     Basic01        The parent zone is "1.b.9.0.1.0.0.2.ip6.arpa" as returned from name servers "ns-auth-01.bahnhof.net/158.174.209.10;ns-auth-01.bahnhof.net/2001:9b0:fff5:a::2;ns-auth-02.bahnhof.net/158.174.209.18;ns-auth-02.bahnhof.net/2001:9b0:fff5:b::2".
  25.74 ERROR    Basic01        "0.d.b.9.1.b.9.0.1.0.0.2.ip6.arpa" does not exist as a DNS zone. Try to test "d.b.9.1.b.9.0.1.0.0.2.ip6.arpa" instead.
  25.74 CRITICAL Unspecified    Not enough data about 0.d.b.9.1.b.9.0.1.0.0.2.ip6.arpa was found to be able to run tests.
```

But if the servers of the parent zone (as listed by `zonemaster-cli`) are queried then they all give the following referal which means that the response from `zonemaster-cli` does not look correct.

```
;; AUTHORITY SECTION:
0.d.b.9.1.b.9.0.1.0.0.2.ip6.arpa. 86400	IN NS	ns1.flashdance.cx.
0.d.b.9.1.b.9.0.1.0.0.2.ip6.arpa. 86400	IN NS	ns2.flashdance.cx.
0.d.b.9.1.b.9.0.1.0.0.2.ip6.arpa. 86400	IN NS	ns3.flashdance.cx.
```

The issue was reported to Zonemaster by @iocc-git. Issues are always welcome.

matsduf:
## Purpose

This PR will remove the limitation revealed in issue #1484 by going beyond NXDOMAIN and finding the delegation even if the name servers give an incorrect NXDOMAIN between apex and the delegation point.

Added test scenarios for both Basic01 and  MethodsV2 are added in #1490.

## Context

#1484

## Changes

* Basic01 specification
* Get-Parent-NS-Names-and-IPs specification in MethodsV2

## How to test this PR

Review.


docs/public/specifications/tests/Basic-TP/basic01.md
```diff
@@ -107,28 +107,30 @@ DNS queries follow, unless otherwise specified below, what is specified for
...
+   into the *Remaining Servers* set.
 
 > In the loop below, the steps tries to capture the name of the parent zone of
 > **Child Zone** and the IP addresses of the name servers for that parent zone.
 > This is done using a modified version of the "QNAME minimization" technique
 > [RFC 9156]. SOA is the query type used for traversing the tree.
 
-6. While the *Remaining Servers* is non-empty pick next name server IP address
-   and zone name from the set ("Server Address" and "Zone Name") and do:
+1. While the *Remaining Servers* is non-empty pick next name server IP address
```
marc-vanderwal: Shouldn’t this be step 6?

It seems like the paragraph above confuses GitHub (and possibly Emacs’s `markdown-mode`) because both think that the list is interrupted by the paragraph.

docs/public/specifications/tests/Basic-TP/basic01.md
marc-vanderwal, line 144: Maybe something worth updating as well: if RCODE is NXDOMAIN, the text says that B01_SERVER_ZONE_ERROR should be output, but I don’t think we want to do that. That’s a specification oversight that, unfortunately, happens in several places here.

docs/public/specifications/tests/Basic-TP/basic01.md
```diff
@@ -154,14 +156,19 @@ DNS queries follow, unless otherwise specified below, what is specified for
...
+          *Remaining Servers* set, unless the IP address is already listed in
+          *Handled Servers*, together with *Zone Name*.
+   9.  Do a [DNS Lookup] of name server names (A and AAAA) not already listed in
+       the additional section of the response, and extract IP addresses (A and
+       AAAA records) from responses.
+       1. For each IP address add the IP address and *Zone Name* to the
+          *Remaining Servers* set, unless the IP address is already listed in
+          *Handled Servers*, together with *Zone Name*.
+       2. Ignore any failing lookups or lookups, such as NODATA, NXDOMAIN, non-AA
+          response or no response at all.
```
marc-vanderwal: Typo, I think?
```suggestion
       2. Ignore any failing lookups or negative responses, such as NODATA, NXDOMAIN, non-AA
          response or no response at all.
```

docs/public/specifications/tests/Basic-TP/basic01.md
```diff
@@ -55,20 +55,20 @@ Input for this Test Case:
...
+| B01_PARENT_FOUND            | INFO    | domain, ns_list                      | The parent zone is "{domain}" as returned from name servers "{ns_list}".                                                                                |
+| B01_PARENT_NOT_FOUND        | WARNING |                                      | The parent zone cannot be found.                                                                                                                        |
+| B01_PARENT_UNDETERMINED     | WARNING | ns_list                              | The parent zone cannot be determined on name servers "{ns_list}".                                                                                       |
+| B01_ROOT_HAS_NO_PARENT      | INFO    |                                      | This is a test of the root zone which has no parent zone.                                                                                               |
+| B01_UNEXPECTED_NXDOMAIN     | ERROR   | query_name, ns_list                  | Unexpected NXDOMAIN on intermediate name "{query_name}" between apex and delegation point. Returned from name servers "{ns_list}".                      |
```
marc-vanderwal: This message is technically correct but I think it could do better in expressing how serious of a problem these false NXDOMAIN responses are. The ERROR level seems perfectly appropriate to me. Taking inspiration from @pawal’s work, I came up with the following idea for the message:

> Parent nameserver "{ns}" returns NXDOMAIN for "{domain_parent}" even though a subdomain, "{domain_child}", exists. This violates RFC 8020. The zone might not work properly on some resolvers.

What do you think? For `{domain_child}`, we could substitute the name of the zone cut below which the broken nameserver is no longer responsible. This could be of great help when identifying the root cause of the error (i.e. which parent zone has the problem).

docs/public/specifications/tests/Basic-TP/basic01.md
matsduf: Bad self-reviewing. Already updated the document, but I have implemented your proposal in an update.

docs/public/specifications/tests/Basic-TP/basic01.md
matsduf: Emacs got confused, and that is what happened. But neither Github och MDbook have a problem with it. Fixed.

docs/public/specifications/tests/Basic-TP/basic01.md
matsduf: No, here we query for something that should be start of zone, and then NXDOMAIN is not thinkable.

docs/public/specifications/tests/Basic-TP/basic01.md
matsduf: "{domain_parent}" and "{domain_child}" are not registered argument names.

"Parent nameserver" indicates that it is a name server for the parent of *Child Zone* but is could be further up in the delegation path. Should we have references to RFCs here without a link? I think that fits better in the explanations. I'll do some update and you can see what you think.

docs/public/specifications/tests/Basic-TP/basic01.md
```diff
@@ -55,20 +55,20 @@ Input for this Test Case:
...
+| B01_PARENT_NOT_FOUND        | WARNING |                                      | The parent zone cannot be found.                                                                                                                                                               |
+| B01_PARENT_UNDETERMINED     | WARNING | ns_list                              | The parent zone cannot be determined on name servers "{ns_list}".                                                                                                                              |
+| B01_ROOT_HAS_NO_PARENT      | INFO    |                                      | This is a test of the root zone which has no parent zone.                                                                                                                                      |
+| B01_UNEXPECTED_NXDOMAIN     | ERROR   | query_name, domain, ns_list          | Unexpected NXDOMAIN on intermediate name "{query_name}" between apex and delegation point. The {domain} might not be resolvable on some DNS resolvers. Returned from name servers "{ns_list}". |
```
marc-vanderwal: Looks OK to me, though I think it still needs a small tweak. How about:
```suggestion
| B01_UNEXPECTED_NXDOMAIN     | ERROR   | query_name, domain, ns_list          | Unexpected NXDOMAIN on intermediate name "{query_name}" between apex and delegation point. Because of that, the zone "{domain}" might not be resolvable on some DNS resolvers. Returned from name servers "{ns_list}". |
```

docs/public/specifications/tests/Basic-TP/basic01.md
```diff
@@ -194,20 +202,27 @@ DNS queries follow, unless otherwise specified below, what is specified for
...
-                records in the additional section.
+                   * Owner name of any of the NS records is not
+                     *Intermediate Query Name*.
+             4. Extract the name server names from the NS records in the answer
+                section and any matching address records in the additional
+                section.
              5. Do [DNS Lookup] of name server names (A and AAAA) not already
                 listed in the additional section of the response.
+                    1. Ignore any lookup that goes through a CNAME.
+                    2. Ignore any failing lookups or lookups, such as NODATA, NXDOMAIN, non-AA
```
marc-vanderwal: GitHub does not render this list as it should. Maybe it is indented one space too many? Also, the sentence seems to be missing some words:
```suggestion
                   1. Ignore any lookup that goes through a CNAME.
                   2. Ignore any failing lookups or negative responses, such as NODATA, NXDOMAIN, non-AA responses or no response at all.
```

docs/public/specifications/tests/Basic-TP/basic01.md
```diff
@@ -152,16 +154,22 @@ DNS queries follow, unless otherwise specified below, what is specified for
...
+       AAAA records) from responses.
        1. For each IP address add the IP address and *Zone Name* to the
-          *Remaining Servers* set unless the IP address is already listed in
-          *Handled Servers* together with *Zone Name*.
-       2. Ignore any failing lookups or lookups resulting in NODATA or NXDOMAIN.
+          *Remaining Servers* set, unless the IP address is already listed in
+          *Handled Servers*, together with *Zone Name*.
+       2. Ignore any lookup that goes through a CNAME.
+       3. Ignore any failing lookups or negative responses, such as NODATA,
+          NXDOMAIN, non-AA response or no response at all.
```
marc-vanderwal:
```suggestion
          NXDOMAIN, non-AA responses or no response at all.
```

docs/public/specifications/tests/MethodsV2.md
```diff
@@ -187,16 +187,20 @@ This Method uses the following input units defined in section [Methods Inputs]:
...
    10. Do [DNS Lookup] of name server names (A and AAAA) not already listed in
        the additional section of the response. If a CNAME is encountered,
        follow the chain of CNAME records but use the original name as obtained
        from the NS record RDATA when storing the data in the next substep.
        1. For each IP address add the name server name, IP address and *Zone
           Name* tuple to the *Remaining Servers* set, unless such a tuple
           already exists in *Handled Servers*.
-       2. Ignore any failing lookups or lookups resulting in NODATA or NXDOMAIN.
+       2. Ignore any failing lookups or lookups, such as NODATA, NXDOMAIN, non-AA
+          response or no response at all.
```
marc-vanderwal:
```suggestion
          responses or no response at all.
```

docs/public/specifications/tests/MethodsV2.md
```diff
@@ -255,7 +261,8 @@ This Method uses the following input units defined in section [Methods Inputs]:
                 Servers* set, unless such a tuple already exists in *Handled
                 Servers*.
           3. Go to next server in *Remaining Servers*.
-       7. Else, if the [RCODE Name] is NoError and the AA is set then do:
+       7. Else, if the [RCODE Name] is NoError or NXDomain, and the AA is set
+          then do:
```
marc-vanderwal: Just noticed a typo that was there in the previous text. Might as well fix it too:
```suggestion
       7. Else, if the [RCODE Name] is NoError or NXDomain, and the AA bit is set
          then do:
```

marc-vanderwal (commented):
I agree with the new logic and the new error message. I just saw some typos here and there.

Now we need new test scenarios to catch these situations.

We should have a scenario where the grandparent zone’s name servers break RFC 8020 but the parent zone’s are good, and a scenario where the grand parent zone’s nameservers work right but not the parent zone’s. Maybe more?

matsduf:
> We should have a scenario where the grandparent zone’s name servers break RFC 8020 but the parent zone’s are good, and a scenario where the grand parent zone’s nameservers work right but not the parent zone’s. Maybe more?

Yes, and maybe both. And maybe where there are several labels between apex and delegation where some gives NODATA and others NXDOMAIN. Also scenario where delegation is inconsistent and the new tag should not be outputted.

I will start that work today.

marc-vanderwal (dismissed):

marc-vanderwal (dismissed):

docs/public/specifications/tests/Basic-TP/basic01.md
```diff
@@ -264,24 +279,30 @@ DNS queries follow, unless otherwise specified below, what is specified for
...
+         * *AA CNAME Found*
+         * *CNAME with Referral Found*
+         * *AA DNAME Found*
+         * *AA NODATA Found*
+
+10. If *[B01_CHILD_FOUND]* has been outputted, but not
+    *[B01_INCONSISTENT_DELEGATION]*, and the *AA NXDomain Response* set is
+    non-empty then for each domain name in that set output
+    *B01_UNEXPECTED_NXDOMAIN* with the domain name from the set, the list of IP
+    addresses for that domain name, and the *Child Zone*.
```
marc-vanderwal, line 297: Why would you want to output *B01_UNEXPECTED_NXDOMAIN* unless *B01_INCONSISTENT_DELEGATION* is output? Can’t we just output *B01_UNEXPECTED_NXDOMAIN* if *B01_CHILD_FOUND* is output and *AA NXDomain Response* is non-empty?

Also, the *B01_UNEXPECTED_NXDOMAIN* tag should be turned into a link.

matsduf: The reasoning for this design is to keep the complexity of the change down. If you look at how B01_CHILD_FOUND is defined, then you can see that unexpected NXDOMAIN can be an affect of inconsistent delegation.

I will fix the missing link.

docs/public/specifications/tests/MethodsV2.md
```diff
@@ -15,8 +15,8 @@
 * [Method: Get zone NS names and IP addresses][Get-Zone-NS-Names-and-IPs]
 * [Method: Get zone NS IP addresses][Get-Zone-NS-IPs]
 * [Method: Get delegation (Internal)][Get-Delegation]
-* [Method: Get in-bailiwick address records in zone (Internal)][Get-IB-Addr-in-Zone]
-* [Method: Get out-of-bailiwick ip addresses (Internal)][Get-OOB-IPs]
+* [Method: Get In-Domain address records in zone (Internal)][Get-ID-Addr-in-Zone]
+* [Method: Get Not In-Domain ip addresses (Internal)][Get-NID-IPs]
```
marc-vanderwal: I think “IP” is better written in capitals:
```suggestion
* [Method: Get Not In-Domain IP addresses (Internal)][Get-NID-IPs]
```

docs/public/specifications/tests/MethodsV2.md
```diff
@@ -903,14 +910,14 @@ This Method depends on [Get-Zone-NS-Names] and [Get-Del-NS-IPs].
 [To top]
 
 
-## Method: Get out-of-bailiwick ip addresses (Internal)
+## Method: Get Not In-Domain ip addresses (Internal)
```
marc-vanderwal: “IP” is best written in capitals:
```suggestion
## Method: Get Not In-Domain IP addresses (Internal)
```

docs/public/specifications/tests/MethodsV2.md
```diff
@@ -1020,12 +1027,22 @@ None.
...
-  pages 24-25. In this document it is limited to the meaning "in domain" in the
-  RFC.
+* "Glue Record" - The term is used as defined in [RFC 9499][RFC 9499#secion7],
+  section 7.
+
+* "In-Domain" - The term is used as defined in [RFC 9499][RFC 9499#secion7],
+  section 7, in the subsection on "Glue Records", for name server names in the
+  referral of a zone. The name server name is on or below the zone cut of the
+  zone for which it is name server for. Previously the term "In-Bailiwick" was
+  used.
```
marc-vanderwal: Just a few typos and one suggestion.

In-domain name server names are always below the zone cut. Consider `example.com`: the zone cut is between `com` and `example.com`, so even a name server named `example.com` that is authoritative for `example.com` would be named below the zone cut.

```suggestion
* "Glue Record" - The term is used as defined in [RFC 9499][RFC 9499#section7],
  section 7.

* "In-Domain" - The term is used as defined in [RFC 9499][RFC 9499#section7],
  section 7, in the subsection on "Glue Records", for name server names in the
  referral of a zone. The name server name is below the zone cut of the zone
  for which it is authoritative. Previously the term "In-Bailiwick" was used.
```

docs/public/specifications/tests/MethodsV2.md
```diff
@@ -1020,12 +1027,22 @@ None.
...
 
-* "Out-Of-Bailiwick" - The terms means, in this document, what is not
-  "In-Bailiwick, in domain". [RFC 8499], section 7,  pages 24-25.
+* "Not In-Domain" - The term refer to a name server names that is not
+  "In-Domain". It is either "sibling domain" or "unrelated" as defined in
+  [RFC 9499][RFC 9499#secion7], section 7, in the subsection on "Glue Records",
+  for name server names in the referral of a zone. The name server name is
+  neither on or below the zone cut of the zone for which it is name server for.
+  It is above or at aside. Previously the term "Out-Of-Bailiwick" was
+  used.
```
marc-vanderwal:
```suggestion
* "Not In-Domain" - The term refers to a name server name that is not
  "In-Domain". It is either "sibling domain" or "unrelated" as defined in
  [RFC 9499][RFC 9499#section7], section 7, in the subsection on "Glue Records",
  for name server names in the referral of a zone. The name server name is
  not below the zone cut of the zone for which it is authoritative.
  It is above or in a different zone. Previously the term "Out-Of-Bailiwick" was
  used.
```

docs/public/specifications/tests/MethodsV2.md
```diff
@@ -1084,11 +1102,11 @@ None.
 [Method5]:                                           Methods.md#method-5-obtain-the-name-server-address-records-from-child
 [Methods]:                                           Methods.md
 [Methods Inputs]:                                    #methods-inputs
-[Out-Of-Bailiwick]:                                  #terminology
+[Not In-Domain]:                                     #terminology
 [Query type]:                                        https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-4
 [RCODE Name]:                                        https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-6
-[RFC 8499]:                                          https://www.rfc-editor.org/rfc/rfc8499.html#section-7
 [RFC 9156]:                                          https://www.rfc-editor.org/rfc/rfc9156.html
+[RFC 9499#secion7]:                                  https://datatracker.ietf.org/doc/html/rfc9499#section-7
```
marc-vanderwal:
```suggestion
[RFC 9499#section7]:                                 https://datatracker.ietf.org/doc/html/rfc9499#section-7
```

docs/public/specifications/tests/MethodsV2.md
matsduf: If "zone cut" really defined as between "com" and "example.com" then that would mean that DS records are below the zone cut. I would say that "example.com" is the zone cut, where some records belong to the parent zone and others to the child zone. I see no support for your definition in the RFC.

matsduf: I have worked in some of your suggestions.

docs/public/specifications/tests/MethodsV2.md
matsduf: I have worked in some of your suggestions.

docs/public/specifications/tests/MethodsV2.md
```diff
@@ -1020,12 +1027,22 @@ None.
...
+  zone for which it is name server for. Previously the term "In-Bailiwick" was
+  used.
 
-* "Out-Of-Bailiwick" - The terms means, in this document, what is not
-  "In-Bailiwick, in domain". [RFC 8499], section 7,  pages 24-25.
+* "Not In-Domain" - The term refers to a name server name that is not
+  "In-Domain". It is either "sibling domain" or "unrelated" as defined in
+  [RFC 9499][RFC 9499#section7], section 7, in the subsection on "Glue Records",
+  for name server names in the referral of a zone. The name server name is
+  neither on or below the zone cut of the zone for for which it is name server.
```
marc-vanderwal: There’s a double word:
```suggestion
  neither on or below the zone cut of the zone for which it is name server.
```

marc-vanderwal (dismissed):

marc-vanderwal (dismissed):

matsduf:
@MichaelTimbert, do you have time to review this too?

marc-vanderwal (approved):

PR #1492, opened 2026-06-08
Fixes a link error

matsduf:
## Purpose

This PR fixes a link error in the README file in the public/configuration directory.

## How to test this PR

Review and make sure that MD book compiles.


mattias-p (approved):
The change itself looks good.

But I have two questions regarding the surrounding text:

1. Should we really refer to the feature by the name "URL to TLD"? I thought its name was "TLD URL". "URL to TLD" sounds like a conversion from a URL to a TLD, and that's not very descriptive.

2. Should we really refer to "the upcoming release" in the public documentation? If we don't have a process to clean that up, it feels to me like there's a substantial risk that this will be outdated quite soon.

tgreenx (approved):

matsduf:
> But I have two questions regarding the surrounding text:

Good comments. Let us look at those before we release the feature in GUI.

PR #1516, opened 2026-07-08
Updates Zone09

### LINKED ISSUE #1515: MX records returned in random order causes warning ? (Zone09)
I tested my domain (burn.net) on zomemaster.net. Got this warning:

```
MX record present
Zone09
Warning
The MX RRset data is inconsistent between the name servers.
Info
Mail targets in the MX RRset "phosphor.burn.net.; mx1.burn.net." returned from name servers "2001:19f0:5c01:19a3:45:76:18:80; 45.76.18.80".
Info
Mail targets in the MX RRset "mx1.burn.net.; phosphor.burn.net." returned from name servers "208.79.241.3; 2607:fe70:0:4::b".
Info
Mail targets in the MX RRset "phosphor.burn.net.; mx1.burn.net." returned from name servers "208.79.240.3; 2607:fe70:0:3::b".
```

matsduf:
## Purpose

This PR resolves issue #1515

## Context

Resolves #1515

Scenarios are updated in #1517 to match the updates in this PR.

## Changes

* Ignores differences in TTL.
* Explicitly includes preference in messages and comparisons.
* Uses (mail) "exchange" instead of (mail) "target".
* Lowers the default level of some messages.

## How to test this PR

Review.


docs/public/specifications/tests/Zone-TP/zone09.md
```diff
@@ -98,21 +98,21 @@ correct DNS response for an authoritative name server.
...
+| Z09_MX_DATA               | INFO    | ns_list, mxrdata_list | The MX RDATA in the MX RRset, "{mxrdata_list}", as returned by name servers "{ns_list}".           |
+| Z09_MX_FOUND              | INFO    | ns_list               | MX RRset was returned by name servers "{ns_list}".                                                 |
+| Z09_NON_AUTH_MX_RESPONSE  | WARNING | ns_list               | Non-authoritative response on MX query from name servers "{ns_list}".                              |
+| Z09_NO_MX_FOUND           | INFO    | ns_list               | No MX RRset was returned by name servers "{ns_list}".                                              |
+| Z09_NO_RESPONSE_MX_QUERY  | WARNING | ns_list               | No response on MX query from name servers "{ns_list}".                                             |
+| Z09_NULL_MX_NON_ZERO_PREF | NOTICE  |                       | The zone has a Null MX with non-zero preference.                                                   |
+| Z09_NULL_MX_WITH_OTHER_MX | NOTICE | mxdata_list           | The zone has a Null MX mixed with other MX records "{mxrdata_list}".                               |
```
marc-vanderwal: That kind of misconfiguration can disrupt the domain’s ability to receive mail, like an inconsistent MX RRset across authoritative name servers would. So I think that a level of WARNING is warranted.

True, raising that message level to WARNING kind of bends the current definition of the WARNING level, but NOTICE downplays the seriousness of this condition.

Maybe we should broaden the definition of WARNING somewhat? Maybe we should include conditions that could affect the resolvability of the zone in degraded scenarios, _or the correct operation of essential services like email_?

docs/public/specifications/tests/ArgumentsForTestCaseMessages.md
```diff
@@ -48,39 +48,39 @@ The following is a message (*msgid* in this case) where this is in use:
...
+| label          | Domain name label                                                     | A single label, i.e. the string between the dots, from a domain name.                                                        |
+| locale         | Locale label                                                          | A locale label, user provided or taken from the environment                                                                  |
+| module         | A Zonemaster test module, or `all`                                    | The name of a Zonemaster test module.                                                                                        |
+| module_list    | List of Zonemaster test modules                                       | A list of Zonemaster test modules, separated by ":".                                                                         |
+| mxrdata        | Non-negative integer and domain name, space separated i a text string | Preference and (mail) exchange from MX RDATA ([RFC 1035][RFC1035#section-3.3.9], section 3.3.9).                             |
```
marc-vanderwal:
```suggestion
| mxrdata        | Non-negative integer and domain name, space separated in a text string | Preference and (mail) exchange from MX RDATA ([RFC 1035][RFC1035#section-3.3.9], section 3.3.9).                             |
```

docs/public/specifications/tests/Zone-TP/zone09.md
```diff
@@ -250,7 +263,7 @@ None.
 
 ## Terminology
 
-The term "Null MX" is used for an MX record where the mail target is "." as
+The term "Null MX" is used for an MX record where the (mail) exchange is "." as
```
marc-vanderwal: RFC 7505 says a null MX also has a preference of 0.

docs/public/specifications/tests/Zone-TP/zone09.md
matsduf: I think WARNING is fine in this case. I have updated.

Updating and clarifying the definitions of the levels is fine, but out of scope of this PR.

docs/public/specifications/tests/Zone-TP/zone09.md
matsduf: I have updated with an explicit statement.

marc-vanderwal (dismissed):
Looks good to me. I’ve double-checked the new message levels you proposed and I see no problem either.

matsduf:
> Looks good to me. I’ve double-checked the new message levels you proposed and I see no problem either.

I did a few more changes. Please check.

marc-vanderwal (dismissed):

matsduf:
Updates so that Z09_MX_DATA is always outputted if there is some MX record. New tag Z09_VALID_NULL_MX that is outputted if there is a valid Null MX.

matsduf:
Updates so that for all zone the test case will always output some message unless all name servers fail to respond on SOA query.

matsduf:
After this update there are three situations where no message is outputted on INFO level or higer:
* The zone does not exist, i.e. there are no NS to test against.
* There are no addresses for the NS in the delegation.
* No NS gives a correct (expected) response on the SOA query, e.g. SERVFAIL,non-AA or no response on all NS.

The above situation could happen if the test case is tested without testing the Basic test cases first, i.e. they will not happen in a normal test situation. The above situation is not unique for this test case. It is the same in most test cases.

Either we could leave it as is or we could do another update and add tests and message tags:
* If there are no NS from MethodV2, then output a CRITICAL message tag saying that there are no NS to test against.
* If all NS failed the SOA query, then output a CRITICAL message tag saying that all NS failed the SOA query.

docs/public/specifications/tests/Zone-TP/zone09.md
marc-vanderwal: I see a typo in a bit of text that wasn’t updated by this PR, but we might as well:
```suggestion
[Email Domains][Email Domain] ([RFC 5321][RFC 5321#section-2.3.5], section 2.3.5)
```

docs/public/specifications/tests/Zone-TP/zone09.md
```diff
@@ -136,19 +137,23 @@ queries follow, unless otherwise specified below, what is specified for
...
-3.  Obtain the set of name server IP addresses using [Method4] and [Method5]
-    ("Name Server IP").
+3.  Obtain the set of name server names and IP addresses using methods
+    [Get-Del-NS-Names-and-IPs] and [Get-Zone-NS-Names-and-IPs] ("Name Servers").
 
-4. Create the following empty sets
+4.  Extract the unique set of name server IP addresses from *Name Servers*
+    ("Name Server IPs").
+
+5. Create the following empty sets
```
marc-vanderwal:
```suggestion
5.  Create the following empty sets:
```

docs/public/specifications/tests/Zone-TP/zone09.md
```diff
@@ -136,19 +137,23 @@ queries follow, unless otherwise specified below, what is specified for
 2.  Create a [DNS Query] with query type  MX and query name *Child Zone*
```
marc-vanderwal:
```suggestion
2.  Create a [DNS Query] with query type MX and query name *Child Zone*
```

docs/public/specifications/tests/Zone-TP/zone09.md
marc-vanderwal: We can also do away with this item, because [the default handling of a DNS response](https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/tests/DNSQueryAndResponseDefaults.md#default-handling-of-a-dns-response) we specify already says that we should retry over TCP if the TC bit is set in the response.

docs/public/specifications/tests/Zone-TP/zone09.md
```diff
@@ -172,55 +177,72 @@ queries follow, unless otherwise specified below, what is specified for
...
+          2. For each MX record extract the RDATA as a text string of space
+             separated preference (integer) and exchange, i.e. in the same format
+             as MX RDATA is shown in presentation format.
+          2. Create a sorted list of the RDATA text strings where primary sort
+             key is the preference (ascending order) and the secondary sort key
+             is the exchange (ascending order).
+          3. Add the name server IP and the sorted list to the *MX RDATA Lists*
+             set.
+
+7.  If the set *No Response MX Query* is non-empty, then output
```
marc-vanderwal:
```suggestion
7.  If the *No Response MX Query* set is non-empty, then output
```

docs/public/specifications/tests/Zone-TP/zone09.md
```diff
@@ -172,55 +177,72 @@ queries follow, unless otherwise specified below, what is specified for
...
     *[Z09_NO_RESPONSE_MX_QUERY]* with the name server IP addresses from the set.
 
-7.  If the set *Unexpected RCODE MX Response* is non-empty, then for each RCODE
+8.  If the set *Unexpected RCODE MX Response* is non-empty, then for each RCODE
```
marc-vanderwal:
```suggestion
8.  If the *Unexpected RCODE MX Response* set is non-empty, then for each RCODE
```

docs/public/specifications/tests/Zone-TP/zone09.md
```diff
@@ -172,55 +177,72 @@ queries follow, unless otherwise specified below, what is specified for
...
     in the set, do:
     * Output *[Z09_UNEXPECTED_RCODE_MX]* with the RCODE value
       ([IANA RCODE List]) and the name server IP addresses from the set.
 
-8.  If the set *Non-authoritative MX* is non-empty, then output
+9.  If the set *Non-authoritative MX* is non-empty, then output
```
marc-vanderwal:
```suggestion
9.  If the *Non-authoritative MX* set is non-empty, then output
```

docs/public/specifications/tests/Zone-TP/zone09.md
```diff
@@ -172,55 +177,72 @@ queries follow, unless otherwise specified below, what is specified for
...
+          the list then output *[Z09_ROOT_EMAIL_DOMAIN]*.
+       5. If *Child Zone* is a zone in the ARPA tree, not .ARPA itself, with
+          [non-Null MX][Null MX] RDATA in the list then output
+          *[Z09_ARPA_EMAIL_DOMAIN]*.
+
+13. If the *No MX RRset* set is non-empty and the *MX RDATA Lists* set is empty
+    then:
+    1. If *Child Zone* is the root zone ("."), a [TLD] or a zone in the .ARPA
+       tree then output *[Z09_NO_MX_FOUND_OR_EXPECTED]*.
+    2. Else, Output *[Z09_MISSING_MAIL_EXCHANGE]* with the name server IP
```
marc-vanderwal:
```suggestion
    2. Else, output *[Z09_MISSING_MAIL_EXCHANGE]* with the name server IP
```

marc-vanderwal (commented):
Looks pretty good. I’ve found a few more little things that may be worth including in this update.

docs/public/specifications/tests/Zone-TP/zone09.md
matsduf: I do not see the typo.

matsduf: I'll update.

docs/public/specifications/tests/Zone-TP/zone09.md
matsduf: Why not "each"?

docs/public/specifications/tests/Zone-TP/zone09.md
matsduf: All items on the first level have that extra space if the number is below 10 to make it even with the ones above 9.

docs/public/specifications/tests/Zone-TP/zone09.md
marc-vanderwal: The link text says **RCF** 5321 instead of **RFC** 5321. (Somehow the suggestion didn’t work as I expected.)

docs/public/specifications/tests/Zone-TP/zone09.md
marc-vanderwal: My mistake. I only meant to suggest to move the “set” after “*Unexpected RCODE MX Response*”. I’ll update my suggestion.

matsduf: You do not have to. I have fixed that. And updated to use "RCODE Name" instead there (and in other places) with a link to IANA.

marc-vanderwal (dismissed):

matsduf:
@MichaelTimbert, can you review this? @marc-vanderwal has approved.

docs/public/specifications/tests/Zone-TP/zone09.md
```diff
@@ -133,94 +134,112 @@ queries follow, unless otherwise specified below, what is specified for
...
+6.  For each name server IP in *Name Server IPs* do:
 
     1. Send *SOA Query* over UDP to the name server.
     2. Go to next name server IP if at least one of the following criteria is
        met:
        1. There is no DNS response.
-       2. The RCODE of the response is not "NoError" ([IANA RCODE List]).
+       2. The [RCODE Name] of the response is not "NoError".
        3. The AA flag is not set in the response.
        4. There is no SOA record with owner name matching the query.
```
marc-vanderwal, line 164: Come to think of it, I don’t really understand the purpose of these two steps as well. Other test cases don’t send out SOA queries but just assume the zone is reachable, counting on Basic01 to fail early if the zone isn’t working at all. These steps also prevent the NON-AUTH-MX-RESPONSE scenario from working properly because the name server that serves the zone without the AA bit cannot be configured to respond with the AA bit set on SOA queries but with the AA bit unset on MX queries. Do you think we could remove this part?

matsduf: Without the SOA query we cannot check if the server does not respond on MX queries because they are MX queries or because the server does not respond at all.

There are other test cases where we test with SOA first.

matsduf: I think it is sound to have the SOA query first. Instead we should find another way to implement the scenarios where the AA bit is unset.

marc-vanderwal: Okay. I’ll just mark the NON-AUTH-MX-RESPONSE scenario as not testable in the unit tests for the time being.

marc-vanderwal:
> After this update there are three situations where no message is outputted on INFO level or higer:
> 
>     * The zone does not exist, i.e. there are no NS to test against.
> 
>     * There are no addresses for the NS in the delegation.
> 
>     * No NS gives a correct (expected) response on the SOA query, e.g. SERVFAIL,non-AA or no response on all NS.
> 
> 
> The above situation could happen if the test case is tested without testing the Basic test cases first, i.e. they will not happen in a normal test situation. The above situation is not unique for this test case. It is the same in most test cases.

I think it’s fine as it is. Otherwise we might end up with many cascading error messages when the zone is out of order. Best to avoid making too much noise in that case.

matsduf:
> > The above situation could happen if the test case is tested without testing the Basic test cases first, i.e. they will not happen in a normal test situation. The above situation is not unique for this test case. It is the same in most test cases.
> 
> I think it’s fine as it is. Otherwise we might end up with many cascading error messages when the zone is out of order. Best to avoid making too much noise in that case.

There are pros and cons. In the normal case there would not be any extra error messages. If a non-existing zone is tested with just this test case the output on info level will be empty. If testing all test cases except Basic then it could give many error messages if such messages are implemented i all test cases.

Always giving a message on INFO or higher can be helpful.

docs/public/specifications/tests/Zone-TP/zone09.md
```diff
@@ -133,94 +134,112 @@ queries follow, unless otherwise specified below, what is specified for
...
+          1. Extract the MX records from the response.
+          2. For each MX record down case the (mail) exchange (domain name).
+          2. For each MX record extract the RDATA as a text string of space
+             separated preference (integer) and exchange, i.e. in the same format
+             as MX RDATA is shown in presentation format.
+          2. Create a sorted list of the RDATA text strings where primary sort
+             key is the preference (ascending order) and the secondary sort key
+             is the exchange (ascending order).
+          3. Add the name server IP and the sorted list to the *MX RDATA Lists*
+             set.
```
MichaelTimbert, line 187:
```suggestion
          1. Extract the MX records from the response.
          2. For each MX record down case the (mail) exchange (domain name).
          3. For each MX record extract the RDATA as a text string of space
             separated preference (integer) and exchange, i.e. in the same format
             as MX RDATA is shown in presentation format.
          4. Create a sorted list of the RDATA text strings where primary sort
             key is the preference (ascending order) and the secondary sort key
             is the exchange (ascending order).
          5. Add the name server IP and the sorted list to the *MX RDATA Lists*
             set.
```

docs/public/specifications/tests/Zone-TP/zone09.md
```diff
@@ -133,94 +134,112 @@ queries follow, unless otherwise specified below, what is specified for
...
-          targets from the RDATA and the associated name server IP addresses in
-          the set.
     2. Else do:
-       1. If the mailtarget of any of the MX records in the RRset in *MX RRset*
-          is a [Null MX] then do:
-          1. If there are more than one record in the MX RRset, then output
-             *[Z09_NULL_MX_WITH_OTHER_MX]* with the mail targets from the RDATA
-             of MX records.
-          2. If the preference of the [Null MX] is non-zero then output
-             *[Z09_NULL_MX_NON_ZERO_PREF]*.
```
MichaelTimbert, line 211: In terminology, "Null MX" is defined by a preference value set to 0. I therefore do not understand how this state can be reached.

matsduf: Good catch! Will fix.

marc-vanderwal (approved):

docs/public/specifications/tests/Zone-TP/zone09.md
```diff
@@ -133,94 +134,110 @@ queries follow, unless otherwise specified below, what is specified for
...
-      2. *Child Zone* is a [TLD], or
-      3. *Child Zone* is a zone in the .ARPA tree.
+       1. Extract the unique list of RDATA from *MX RDATA Lists*.
+       2. If any of the MX in the list is a [Null MX] or a [Null-Type MX], then
+          do:
+          1. If there are more than one item in the list, then output
+             *[Z09_NULL_MX_WITH_OTHER_MX]*.
+          2. Else, if the preference in the MX RDATA in the list is non-zero
+             ([Null-Type MX]) then output *[Z09_NULL_MX_NON_ZERO_PREF]*.
+          3. Else, Output *[Z09_VALID_NULL_MX]*.
```
tgreenx:
```suggestion
          3. Else, output *[Z09_VALID_NULL_MX]*.
```

tgreenx (dismissed):

tgreenx (dismissed):

marc-vanderwal (approved):

tgreenx (approved):

marc-vanderwal (approved):

PR #1514, opened 2026-07-01
Update scenarios for Consistency05 match updated specification 

matsduf:
## Purpose

A bug was reported in zonemaster/zonemaster-engine#1537. Two new scenarios were created, ADDRESSES-MATCH-8 and ADDRESSES-MATCH-9, that could show that the is a bug in the implementation of Consistency05.

After that it was decided to first rewrite the specification of the test case (see #1523) and then rewrite the implementation. One important part is to use [MethodsV2] instead of [MethodsV1].

This is an update of the scenarios to match the updated specification in #1523. Most scenarios are unchanged (e.g. DDRESSES-MATCH-8 and ADDRESSES-MATCH-9) but the expected messages tags for in the output for each scenario is however updated. A number of scenarios have been added. 

This PR is ready for review. When there is an updated implementation, `Consistency-TP/consistency05/test-zones-output.md` will be updated.

Dependency on the centralized root zone in the CoreDNS environment has been removed. The test zones can be loaded with the following command:

`sudo coredns --conf Consistency-TP/consistency05/consistency05.cfg`

## Context

zonemaster/zonemaster-engine#1537, #1523 

## How to test this PR

Review and test the scenarios.

[MethodsV2]: https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/tests/MethodsV2.md
[MethodsV1]: https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/tests/Methods.md

PR #1523, opened 2026-07-22
Updates Consistency05 specification

matsduf:
## Purpose

This PR is a response of the bug reported in https://github.com/zonemaster/zonemaster-engine/issues/1537 which is reproduced in two scenarios in #1514. The conclusion is that it would be hard to fix the current implementation. Furthermore, the current specification also needs an update.

This is a complete rewrite of the specification and there is no meaning of direct comparing old and new.

* New title
* New messages
* New procedures
* New format

## Context

* https://github.com/zonemaster/zonemaster-engine/issues/1537 
* #1514
* https://github.com/zonemaster/zonemaster-engine/pull/1546 (fixes related to this update)
* https://github.com/zonemaster/zonemaster-engine/pull/1547

The scenarios also needs to be rewritten or at least thoroughly reviewed and adjusted.

## How to test this PR

Review.


pawal:
Do you intend to silently ignore non-responsive servers  and broken servers now?

(No more NO_RESPONSE and CHILD_NS_FAILED tags)

matsduf:
> Do you intend to silently ignore non-responsive servers and broken servers now?
> 
> (No more NO_RESPONSE and CHILD_NS_FAILED tags)

@pawal, I am glad that you have noticed that fact. It is stated in the updated specification:

```
## Scope
It is assumed that *Child Zone* is tested and reported by [Connectivity01]. This
test case will just ignore non-responsive name servers or name servers not giving
a correct DNS response for an authoritative name server. However, if there is no
working name sever for *Child Zone* then this test case will report that.
```

The two messages, or tags, that you refer to are by default outputted on DEBUG level, which means that they are not visible in GUI (or via Backend) and in practice not visible in CLI except in rare cases.

If you run just this test case you will miss the information, but if you also run Connectivity01 you will get it. It would be simple to also report non-responsive or broken name servers in this test case -- in all test cases -- but then there is a risk that the users will be drowned by those messages.

What is your opinion about the balance? What risk do you see by not outputting that information in this test case -- in all test cases?

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,327 @@
...
-         in the previous DNS lookups, output 
-         *[OUT_OF_BAILIWICK_ADDR_MISMATCH]*.
-
-8. If none of *[IN_BAILIWICK_ADDR_MISMATCH]*, *[EXTRA_ADDRESS_CHILD]* 
-   or *[OUT_OF_BAILIWICK_ADDR_MISMATCH]* has been outputted, output 
-   *[ADDRESSES_MATCH]*.
+* "Child Zone" - The name of the zone to be tested. It must be a [valid domain name].
+* "Undelegated Data" - Optional data. If included it must consist of a set of
+  at least one [valid name server name] and for each name server name an optional
+  set of at least one [valid IP address].
```
marc-vanderwal: If it’s an optional set of at least one, we might as well just say that it’s a set of zero or more:

```suggestion
  at least one [valid name server name] and for each name server name a set of zero or
  more [valid IP addresses][valid IP address].
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -165,41 +330,106 @@ None
...
+
+* "Send" - The terms are used when a DNS query is sent to a specific name server
+  (name server IP address).
+
+* "Valid Domain Name" -- The term stands for a non-empty domain name string that
+  has successfully passed the tests and normalizations in the
+  [Requirements and normalization] specification.
+
+* "Valid IP Address" -- The term stands for either an [IPv4] address or an [IPv6]
+  address in any address range.
```
marc-vanderwal: This term implies that we deem some IP addresses invalid, but in practice we don’t. As far as I can tell, we’d allow 0.0.0.0 and :: as “valid” IP addresses. How about just “IP address” instead of “valid IP address”?

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,327 @@
...
+| CS05_DELEGATION                | INFO     | ns_deleg_list, ns_list                    | Delegation of the child zone as provided by the parent name servers listed: "{ns_deleg_list}". Parent name servers: "{ns_list}".                       |
+| CS05_EXTRA_ADDR_CHILD          | NOTICE   | ns_list                                   | There is one or more extra address records found in the child zone that are not present as glue in the delegation: "{ns_list}".                        |
+| CS05_NO_MISMATCH_GLUE_ZONE     | INFO     |                                           | There is no mismatch between delegation from parent and authoritative data in the child zone.                                                          |
+| CS05_ID_ADDR_MISMATCH          | ERROR    | nsname, ns_ip_list_glue, ns_ip_list_zone  | For name server {nsname} the glue record in the delegation "{ns_ip_list_glue}" mismatches the address record in the child zone "{ns_ip_list_zone}".    |
```
marc-vanderwal:
```suggestion
| CS05_ID_ADDR_MISMATCH          | ERROR    | nsname, ns_ip_list_glue, ns_ip_list_zone  | For name server {nsname} the glue record in the delegation "{ns_ip_list_glue}" is different from the address record in the child zone "{ns_ip_list_zone}".    |
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,327 @@
...
+| CS05_MISSING_GLUE_FOR_NS       | WARNING  | nsname, ns_list                           | Expected glue record for {nsname} is missing in the delegation. Found in the parent name servers "{ns_list}".                                          |
+| CS05_MISSING_GLUE_FOR_NS_UNDEL | WARNING  | nsname                                    | IP address (glue record) is expected but missing for {nsname} in the undelegated data.                                                                 |
+| CS05_NO_NS_ADDR_CHILD          | CRITICAL |                                           | Child zone cannot be tested since there are no name server IP addresses for that zone.                                                                 |
+| CS05_OOD_ADDR_MISMATCH         | WARNING  | nsname, ns_ip_list_ref, ns_ip_list_lookup | For name server {nsname} the glue record in the delegation "{ns_ip_list_glue}" mismatches the address record in authoritativ zone "{ns_ip_list_zone}". |
```
marc-vanderwal:
```suggestion
| CS05_OOD_ADDR_MISMATCH         | WARNING  | nsname, ns_ip_list_ref, ns_ip_list_lookup | For name server {nsname} the glue record in the delegation "{ns_ip_list_glue}" is different from the address record in the child zone "{ns_ip_list_zone}". |
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,327 @@
...
+
+5.  If the *Parent NS IPs* is non-empty, then for each name server IP in the set
+    do:
+
+    1. [Send] *SOA Query* over UDP to the name server IP and fetch the response
+       (if any).
+    3. If the response (if any) contains a [Referral] of *Child Zone* then do:
+       1. Extract the name server names from delegation NS records in authority
+          section.
+       2. Downcase any uppcase letters in the names.
```
marc-vanderwal:
```suggestion
       2. Downcase any uppercase letters in the names.
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,327 @@
...
+       3. Create an sorted list of unique name server names (sorted ascendingly
+          on the name).
```
marc-vanderwal:
```suggestion
       3. Create a sorted list of unique name server names (sorted in ascending order
          on the name).
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,327 @@
...
+[DNS Query and Response Defaults]. The handling of the DNS responses on the DNS
+queries follow, unless otherwise specified below, what is specified for
+[DNS Response] in the same specification.
+
+1.  Create the following sets from fetched or derived data:
+
+    1. Fetch the parent name servers with IP addresses using method
+       [Get-Parent-NS-Names-and-IPs] ("Parent NS").
+    2. Fetch the parent name server IPs using method [Get-Parent-NS-IPs]
+       ("Parent NS IPs").
```
marc-vanderwal, line 112: It’s a bit awkward that we say that we should do this work, only to clear those sets if undelegated data was provided as input. How about:


```suggestion
    1. "Parent NS": if the *Undelegated Data* set (from input) is empty, then populate
       this set by fetching the parent name servers with IP addresses using method
       [Get-Parent-NS-Names-and-IPs]; otherwise, this set is empty.
    2. "Parent NS IPs": if the *Undelegated Data* set (from input) is empty, then populate
       this set by fetching the parent name server IPs using method [Get-Parent-NS-IPs];
       otherwise, this set is empty.
```

And then, step 2 can be removed.

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,327 @@
...
+          2. If the name already has an IP address attached to it, then copy the
+             name and create a new name/IP pair.
+          3. Only unique name/IP pairs are created.
+          4. Sort the name/IP pair with IPv4 first and then sorted ascendingly on
+             the IP address.
+       5. Add the name server IP and the created name server list to the
+          *Delegation* set.
+
+6.  If the *Undelegated Data* set is non-empty then the *Delegation* set is
+    empty.
```
marc-vanderwal: If the *Undelegated Data* set is non-empty, then we know that *Parent NS IPs* is empty, and therefore we skip step 5 where we populate the *Delegation* set, so this step is unnecessary.

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,327 @@
...
+    3. If the response (if any) contains a [Referral] of *Child Zone* then do:
+       1. Extract the name server names from delegation NS records in authority
+          section.
+       2. Downcase any uppcase letters in the names.
+       3. Create an sorted list of unique name server names (sorted ascendingly
+          on the name).
+       4. For each [glue record] in the additional section (if any) do:
+          1. Extract IP address from the [glue record] and attach it to the
+             correct name server name in the name server list to form a
+             name/IP pair.
```
marc-vanderwal: It’s unclear how to “attach [the IP address] to the correct name server name in the name server list”.

Can’t we just form a name/IP pair by extracting the resource record’s owner name and the IP address in the resource record’s RDATA?

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,327 @@
...
+          2. If the name already has an IP address attached to it, then copy the
+             name and create a new name/IP pair.
```
marc-vanderwal: Assuming the pair is constructed from the RR’s owner name and RDATA, what’s the point of this step?

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,327 @@
...
+          3. Only unique name/IP pairs are created.
+          4. Sort the name/IP pair with IPv4 first and then sorted ascendingly on
+             the IP address.
```
marc-vanderwal:
```suggestion
          4. Sort the name/IP pair by IP address, in ascending order, with IPv4
             addresses sorted before IPv6 addresses.
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,327 @@
...
+       5. Add the name server IP and the created name server list to the
+          *Delegation* set.
```
marc-vanderwal: For a while I didn’t understand this step, but after thinking a bit, I understand that we’re adding the following pair to the *Delegation* set:

 * the name server IP of the server to which we sent the NS query;
 * and the list we built in step 5.3.4.

Therefore, how about:

```suggestion
       5. Add the pair consisting of the parent name server IP (to which the NS query
          was sent) and the name server list created in the previous step to the
          *Delegation* set.
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,327 @@
...
+       1. If the name in the element is [In-Domain] then do:
+          1. If element is a name and not a name/IP pair then add the name server
+             IP address and the name to the *Missing Glue* set.
+          2. Add the element (name or name/IP pair) to the *Delegation ID NS*
+             set. 
+             1. Do not create duplicates in the set.
+             2. Elements just consiting of a name is not added if there already
+                is a name/IP pair with the same name.
+             3. A name/IP pair will overwrite an element consisting just of of
+                the same name.
```
marc-vanderwal:
```suggestion
             3. A name/IP pair will overwrite an element consisting just of
                the same name.
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,327 @@
...
+       2. If name in the element (name or name/IP pair) is [Out-Of-Domain] then
+          add the element (name or name/IP pair) to the *Delegation OOD NS* set. 
+          1. Do not create duplicates in the set.
+          2. Elements just consiting of a name is not added if there already is a
+             name/IP pair with the same name.
+          3. A name/IP pair will overwrite an element consisting just of of the
+             same name.
```
marc-vanderwal:
```suggestion
          3. A name/IP pair will overwrite an element consisting just of the
             same name.
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,327 @@
...
+   1. Compare the name server lists (of names or name/IP pairs) between all
+      parent name server IPs.
+   2. If not all name server lists are equal then do:
+      1. For each found name server list output *[CS05_DELEGATION]* with the name
+      server list and the list of parent name server IPs from which the list
+      came.
+      2. Output *[CS05_INCONSISTENT_DELEGATION]*.
+
+9.  If the *Undelegated Data* set is non-empty then do for each name server name
+    in the set do:
```
marc-vanderwal:
```suggestion
9.  If the *Undelegated Data* set is non-empty then for each name server name
    in the set do:
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: The way I have written it is the way I written it in several places. I think it is better to keep as it is.

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: I do not understand. We can have one NS record and several address records.

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: I will look at it again.

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: Yes, the name server IP to which we sent the SOA query. I will look at it again.

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: It is just informational when you follow the procedure "by hand".

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: I copied it from MethodsV2 specification. When you read what it says in the terminology section it clearly says that any IP address is accepted so I do not see how you read it.

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: Step 2 is just information.

matsduf: > It’s a bit awkward that we say that we should do this work, only to clear those sets if undelegated data was provided as input.

That is maybe a misunderstanding. If undelegated data has been provided MethodsV2 will return empty sets on `Get-Parent-NS-Names-and-IPs` and `Get-Parent-NS-IPs` so in step 2 nothing is changed. That is just informational.

marc-vanderwal: Oh I see. Let me come up with a different suggestion, then.

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,345 @@
...
+
+    1. Fetch the parent name servers with IP addresses using method
+       [Get-Parent-NS-Names-and-IPs] ("Parent NS").
+    2. Fetch the parent name server IPs using method [Get-Parent-NS-IPs]
+       ("Parent NS IPs").
+    3. Fetch the child name server IPs using methods [Get-Del-NS-IPs] and
+       [Get-Zone-NS-IPs] and create a unique set ("Child NS IPs").
+
+2.  If the *Undelegated Data* set (from input) is non-empty then the
+    *Parent NS* and *Parent NS IPs* sets are empty.
```
marc-vanderwal: Okay, so this “step” is informational; that’s fine. But it does bother me that there’s a numbered item that, in practice, doesn’t specify anything by itself. How about turning it into a note, like so?

```suggestion
    **Note**: If the *Undelegated Data* set (from input) is non-empty then the
    aforementioned methods return an empty set, and therefore the *Parent NS*
    and *Parent NS IPs* sets are empty.
```

If indented exactly this way, it’ll render exactly as intended, both on both GitHub itself and in the mdbook version: _i.e._ it will not break the list and the next step will be numbered 2, as it should. And in Emacs’s `markdown-mode`, `markdown-cleanup-list-numbers` (`C-c C-c n`) will renumber the list items appropriately.

docs/public/specifications/tests/Consistency-TP/consistency05.md
marc-vanderwal: Yes but each of the address records have an owner name, right? That’s how you “attach” the address record to the name server name in the authority section.

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: I have rephrased it.

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: I have rephrased it.

docs/public/specifications/tests/Consistency-TP/consistency05.md
marc-vanderwal: In that case, how about turning it into a note, like so?
```suggestion
    *Note*: If the *Undelegated Data* set is non-empty then the *Delegation*
    set is empty.
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
marc-vanderwal: Come to think of it, A and AAAA records’ RDATA could indeed be malformed… so I could also read “valid IP addresses” as only keeping address records with well-formed RDATA.

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -399,8 +402,8 @@ None
   has successfully passed the tests and normalizations in the
   [Requirements and normalization] specification.
 
-* "Valid IP Address" -- The term stands for either an [IPv4] address or an [IPv6]
-  address in any address range.
+* "IP Address" -- In this document the term stands for either an [IPv4] address or
+  an [IPv6] address in any address range.
```
marc-vanderwal: mdbook complains about `[IPv4]` and `[IPv6]` being broken links.

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,345 @@
...
+
+This test case will test the following:
+
+* The the [referral] (delegation) is identical on all parent name servers.
+* That the [referral] (delegation) contains at least one [glue record] for each
+  [in-domain] name server name.
+* That each [glue record] matches an authoritative [address record] with the
+  same IP address:
+  * In the *Child Zone* if the [glue record] is [in-domain].
+  * In another zone if the [glue record] is [out-of-zone].
```
marc-vanderwal:
```suggestion
  * In another zone if the [glue record] is [out-of-domain].
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -165,41 +348,111 @@ None
...
+  [Requirements and normalization] specification.
+
+* "IP Address" -- In this document the term stands for either an [IPv4] address or
+  an [IPv6] address in any address range.
+
+* "Valid Name Server Name" -- The term stands for a [Valid Domain Name] that
+  functions as the name of a name server.
+
+
+[Address Record]:                          #terminology
```
marc-vanderwal, line 411: Adding a missing link:
```suggestion
[Address Record]:                          #terminology
[Address Records Lookup]:                  #terminology
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -165,41 +348,111 @@ None
...
+[CS05_ID_ADDR_MISMATCH]:                   #summary
+[CS05_ID_ADDR_MISSING]:                    #summary
+[CS05_INCONSISTENT_DELEGATION]:            #summary
+[CS05_MISSING_GLUE_FOR_NS]:                #summary
+[CS05_MISSING_GLUE_FOR_NS_UNDEL]:          #summary
+[CS05_NO_MISMATCH_GLUE_ZONE]:              #summary
+[CS05_NO_NS_ADDR_CHILD]:                   #summary
+[CS05_OOD_ADDR_MISMATCH]:                  #summary
+[Connectivity01]:                          ../Connectivity-TP/connectivity01.md
+[DNS Query and Response Defaults]:         ../DNSQueryAndResponseDefaults.md
```
marc-vanderwal, line 429: Adding another missing link:
```suggestion
[Connectivity01]:                          ../Connectivity-TP/connectivity01.md
[DNS Lookup]:                              #terminology
[DNS Query and Response Defaults]:         ../DNSQueryAndResponseDefaults.md
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -165,41 +348,111 @@ None
...
-[NO_RESPONSE]:                    #outcomes
-[OUT_OF_BAILIWICK_ADDR_MISMATCH]: #outcomes
-[RFC 7719]:                       https://datatracker.ietf.org/doc/html/rfc7719
-[UNDEL_OOB_ADDR_MISMATCH]:        #outcomes
-[name server requirement]:        https://www.iana.org/help/nameserver-requirements
-[terminology]:                    #terminology
-[undelegated test]:               ../../test-types/undelegated-test.md
-
+* "Address Record" - The term is used for a DNS record of type A or AAAA. The
+  term is used as defined in [RFC 9499][[RFC 9499#section5]], section 5.
```
marc-vanderwal:
```suggestion
  term is used as defined in [RFC 9499][RFC 9499#section5], section 5.
```

marc-vanderwal (commented):
When doing a test build of the documentation with mdbook, it complained about a few broken links.

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: Fixed in a recent commit.

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: Fixed in a recent commit.

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: Fixed in a recent commit.

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: I have followed your proposal, which is a good idea.

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,344 @@
...
-[name server requirement].
+The delegation of the *Child Zone* may contain so called
+[glue records][glue record]. On the wire, any [glue records][glue record] are
+found in the [referral] sent from the name servers of the parent zone.
+
+If the name server name in the NS record is [in-domain] there must be at least
+one [glue record] in the delegation and in the [referral]. For each such glue
+record the equivalent [address record] must exist as an athoritative record in
+the child zone, or below, and the the two (glue record and authoritative record)
+must have the same [IP address].
```
marc-vanderwal:
```suggestion
If the name server name in the NS record is [in-domain] there must be at least
one [glue record] in the delegation and in the [referral]. For each such glue
record the equivalent [address record] must exist as an authoritative record in
the child zone, or below, and both records must have the same [IP address].
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,344 @@
...
+
+* The the [referral] (delegation) is identical on all parent name servers.
+* That the [referral] (delegation) contains at least one [glue record] for each
+  [in-domain] name server name.
+* That each [glue record] matches an authoritative [address record] with the
+  same IP address:
+  * In the *Child Zone* if the [glue record] is [in-domain].
+  * In another zone if the [glue record] is [out-of-domain].
+* If the *Child Zone* contains additional [address records][address record]
+  with the same name as the glue record, but different IP addresses.
```
marc-vanderwal:
```suggestion
* The [referral] (delegation) is identical on all parent name servers.
* That the [referral] (delegation) contains at least one [glue record] for each
  [in-domain] name server name.
* That each [glue record] matches an authoritative [address record] with the
  same IP address:
  * In the *Child Zone* if the [glue record] is [in-domain].
  * In another zone if the [glue record] is [out-of-domain].
* If the *Child Zone* contains additional [address records][address record]
  with the same name as the glue record, but different IP addresses.
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,344 @@
...
-in the delegation are consistent with authoritative data.
 
 ## Scope
 
-It is assumed that *Child Zone* is also tested by [Connectivity01]. This test
-case will set DEBUG level on messages for non-responsive name servers.
+It is assumed that *Child Zone* is tested and reported by [Connectivity01]. This
+test case will just ignore non-responsive name servers or name servers not giving
+a correct DNS response for an authoritative name server. However, if there is no
+working name sever for *Child Zone* then this test case will report that.
```
marc-vanderwal:
```suggestion
working name server for *Child Zone* then this test case will report that.
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -165,41 +347,115 @@ None
...
+  changes to the DNS tree introduced by an [undelegated test] must be respected.
+
+* "DNS Response" - The term is used when the DNS response is to be handled as
+  defined in [DNS Query and Response Defaults][DNS Response].
+
+* "DNS Query" - The term is used for a DNS query that is to follow the
+  specification for DNS queries in [DNS Query and Response Defaults][DNS Query].
+
+* "Glue Record" - [Address records][address record] in the [Referral]
+  whose owner name is equal to the [RDATA] in one of the NS record in the same
```
marc-vanderwal:
```suggestion
  whose owner name is equal to the [RDATA] in one of the NS records in the same
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,344 @@
...
+       (if any).
+    2. If the response (if any) contains the following add the name server
+       IP and the name server names extracted from the NS RRset to the
+       *Child Zone NS* set:
+       * An NS RRset of *Child Zone* in the answer section.
+       * An [RCODE Name] of "NoError".
+       * The AA flag is set.
+    3. Else, go to the next name server IP.
+    4. Create a unique set of [In-Domain] name server names:
+       1. Extract the [In-Domain] NS name server namns from the NS RRset
```
marc-vanderwal:
```suggestion
       1. Extract the [In-Domain] NS name server names from the NS RRset
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,344 @@
...
+10. If *Child NS IP* is empty then output *[CS05_NO_NS_ADDR_CHILD]* and exit
+    these test procedures.
+
+11. For each name server IP in *Child NS IPs* do:
+
+    1. [Send] *NS Query* over UDP to the name server IP and fetch the response
+       (if any).
+    2. If the response (if any) contains the following add the name server
+       IP and the name server names extracted from the NS RRset to the
+       *Child Zone NS* set:
```
marc-vanderwal:
```suggestion
    2. If the response (if any) meets all of the following criteria, add the name server
       IP and the name server names extracted from the NS RRset to the
       *Child Zone NS* set:
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,344 @@
...
+    5. For each [In-Domain] name server name do:
+       1.  Create a [DNS Query] with query type A and query name the NS name
+           server name ("A Query").
+       2.  [Send] *A Query* over UDP to the name server IP and fetch the
+           response (if any).
+       3.  If the response (if any) contains a [Referral] covering the NS name
+           server name then repeat *A Query* (recursively, if needed) to the
+           name servers in the referral until an A RRset is returned or the
+           querying is stopped by e.g. NXDOMAIN or no response.
+           1. If an A RRset is return, then use it in next step as if was a
```
marc-vanderwal:
```suggestion
           1. If an A RRset is returned, then use it in the next step as if was a
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,344 @@
...
+
+    *Note:* If the *Undelegated Data* set is non-empty then the
+    *Delegation* set is empty.
+
+5.  If the *Delegation* set is non-empty, then for each parent name server IP in
+    the set do:
+    1. For each element (name or name/IP pair) in the set do:
+       1. If the name in the element is [In-Domain] then do:
+          1. If element is a name and not a name/IP pair then add the name server
+             IP address and the name to the *Missing Glue* set.
```
marc-vanderwal: Do I understand this right?
```suggestion
          1. If element is a name and not a name/IP pair then add the parent name
             server IP address and the name to the *Missing Glue* set.
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,344 @@
...
+    empty, then do:
+
+    1. Fetch the root hint information using method [Get-Del-NS-Names-and-IPs]
+       ("Hint NS").
+    2. The information in *Hint NS* is assumed to be as single name server names
+       ("name") or as name/IP pairs.
+    3. The hint information is always [In-Domain] by definition.
+    4. For each element (name or name/IP pair) in the *Hint NS* set do:
+       1. If element is a name and not a name/IP pair then add the name server
+          IP address and the name to the *Missing Glue* set.
```
marc-vanderwal: I don’t understand. The element is a plain name, so whose IP address are we using to add an (IP, name) pair to the *Missing Glue* set?

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,344 @@
...
+6. If the *Delegation* set is non-empty, do:
+   1. Compare the name server lists (of names or name/IP pairs) between all
+      parent name server IPs.
+   2. If not all name server lists are equal then do:
+      1. For each found name server list output *[CS05_DELEGATION]* with the name
+      server list and the list of parent name server IPs from which the list
+      came.
+      2. Output *[CS05_INCONSISTENT_DELEGATION]*.
+
+7.  If the *Undelegated Data* set is non-empty then for each name server name
```
marc-vanderwal: Steps 7 and 8 are tricky when the zone being tested is the root zone. Right now, Zonemaster-Engine calls `Zonemaster::Engine::Recursor->add_fake_addresses( '.', … )` when loading the root hints, and when processing the undelegated data. Which means that there is no way to differentiate a test of the root zone that uses a regular hint file, and a test of the root zone with some undelegated data overriding the root name servers.

I understand that there is a need to have a special case for the root zone, but we should maybe assume that in this case, the *Undelegated Data* set is always non-empty. After all, the addresses of the root servers has to come from some non-DNS data, whatever happens.

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,344 @@
...
+       2. Else, add the name server name with IP addresses as one or several
+          name/IP pairs to *Delegation ID NS* set.
+    2. Else (the name server name is [Out-Of-Domain]) then do:
+       1. If there is no IP address for that name, add the name to the
+          *Delegation OOD NS* set.
+       2. Else, add the name and IP address or addresses as name/IP pairs to the
+          *Delegation OOD NS* set.
+
+8. If *Child Zone* is the root zone (".") and the *Undelegated Data* set is
+    empty, then do:
```
marc-vanderwal: Considering my previous comment, this step will never be executed because the root zone always has some undelegated data.

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,344 @@
...
+           name server name ("AAAA Query").
+       7.  [Send] *A Query* over UDP to the name server IP and fetch the
+           response (if any).
+       8.  If the response (if any) contains a [Referral] covering the NS name
+           server name then repeat *AAAA Query* as was done with the *A Query*
+           above.
+       9.  If the response (if any) contains the following then for each
+           unique AAAA record extract owner name and IP address in [RDATA],
+           create a name/IP pair and add that to the *Auth Addr Records In Child*
+           set.
```
marc-vanderwal:
```suggestion
       9.  If the response (if any) meets all of the the following criteria, then for each
           unique AAAA record extract owner name and IP address in [RDATA],
           create a name/IP pair and add that to the *Auth Addr Records In Child*
           set.
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,344 @@
...
+           server name then repeat *A Query* (recursively, if needed) to the
+           name servers in the referral until an A RRset is returned or the
+           querying is stopped by e.g. NXDOMAIN or no response.
+           1. If an A RRset is return, then use it in next step as if was a
+              response on the first query.
+       4.  If any query was not responded to or returned an [RCODE Name] not
+           being "NoError" then go to next NS name server name.
+       5.  If the response (if any) contains the following then for each
+           unique A record extract owner name and IP address in [RDATA], create
+           a name/IP pair and add that to the *Auth Addr Records In Child* set.
```
marc-vanderwal:
```suggestion
       5.  If the response (if any) meets all of the the following criteria, then for each
           unique A record extract owner name and IP address in [RDATA], create
           a name/IP pair and add that to the *Auth Addr Records In Child* set.
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: The root zone is tricky. Back in time I suggested that Zonemaster should refuse to test the root zone at all. That would have saved the methods and many test cases from special cases that are of questionable value.

But in this case I think it is fine that undelegated data overwrites the hint information. I think that is the purpose.

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: If you run this by hand it will make a difference. I will update so that step 7 excludes the root and step 8 is for root only, with or without undelegated data.

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: Copy-and-paste error. I will fix.

pawal:
> > Do you intend to silently ignore non-responsive servers and broken servers now?
> > (No more NO_RESPONSE and CHILD_NS_FAILED tags)
> 
> @pawal, I am glad that you have noticed that fact. It is stated in the updated specification:
> 
> ```
> ## Scope
> It is assumed that *Child Zone* is tested and reported by [Connectivity01]. This
> test case will just ignore non-responsive name servers or name servers not giving
> a correct DNS response for an authoritative name server. However, if there is no
> working name sever for *Child Zone* then this test case will report that.
> ```
> 
> The two messages, or tags, that you refer to are by default outputted on DEBUG level, which means that they are not visible in GUI (or via Backend) and in practice not visible in CLI except in rare cases.
> 
> If you run just this test case you will miss the information, but if you also run Connectivity01 you will get it. It would be simple to also report non-responsive or broken name servers in this test case -- in all test cases -- but then there is a risk that the users will be drowned by those messages.
> 
> What is your opinion about the balance? What risk do you see by not outputting that information in this test case -- in all test cases?

Well, in reality my major concern is that this is a huge change of the specification (and tags) for a minor fix. I did the same fix without changing the whole test case.

If you really think this little fix is enough cause to change the whole test case, ok with me. Although the the per-parent NS-set comparison could be nice to have.

matsduf:
> Well, in reality my major concern is that this is a huge change of the specification (and tags) for a minor fix. I did the same fix without changing the whole test case.
> 
> If you really think this little fix is enough cause to change the whole test case, ok with me. Although the the per-parent NS-set comparison could be nice to have.

We did not want to keep the old implementation. We wanted to use MethodsV2 instead of Methods(V1), and then the result is a rewrite of the specification.

pawal:
> We did not want to keep the old implementation. We wanted to use MethodsV2 instead of Methods(V1), and then the result is a rewrite of the specification.

But you also change the name of a lot of the tags. For me, this is not backwards compatible for somebody that makes use of the tags. Is this intended for a majors version upgrade?

matsduf:
> But you also change the name of a lot of the tags. For me, this is not backwards compatible for somebody that makes use of the tags. Is this intended for a majors version upgrade?

It is a good question. No, changes tag names or tag definitions are not considered to be major updates. Those happen in most releases.

pawal:
> > But you also change the name of a lot of the tags. For me, this is not backwards compatible for somebody that makes use of the tags. Is this intended for a majors version upgrade?
> 
> It is a good question. No, changes tag names or tag definitions are not considered to be major updates. Those happen in most releases.

I haven't seen any mention of a tag-rename in any release document. But ok, then I know what your view is.

matsduf:
> I haven't seen any mention of a tag-rename in any release document. But ok, then I know what your view is.

That is a good point. We probably should.

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,344 @@
...
+           response (if any).
+       3.  If the response (if any) contains a [Referral] covering the NS name
+           server name then repeat *A Query* (recursively, if needed) to the
+           name servers in the referral until an A RRset is returned or the
+           querying is stopped by e.g. NXDOMAIN or no response.
+           1. If an A RRset is returned, then use it in the next step as if was a
+              response on the first query.
+       4.  If any query was not responded to or returned an [RCODE Name] not
+           being "NoError" then go to next NS name server name.
+       5.  If the response (if any) meets all of the the following criteria, then for each
```
marc-vanderwal:
```suggestion
       5.  If the response (if any) meets all of the following criteria, then for each
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,344 @@
...
+           * An [RCODE Name] of "NoError".
+           * The AA flag is set.
+       6.  Create a [DNS Query] with query type AAAA and query name the NS
+           name server name ("AAAA Query").
+       7.  [Send] *A Query* over UDP to the name server IP and fetch the
+           response (if any).
+       8.  If the response (if any) contains a [Referral] covering the NS name
+           server name then repeat *AAAA Query* as was done with the *A Query*
+           above.
+       9.  If the response (if any) meets all of the the following criteria, then for each
```
marc-vanderwal:
```suggestion
       9.  If the response (if any) meets all of the following criteria, then for each
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,344 @@
...
+       the IP addresses extracted from *Parent Glue* for *NS Name* and the IP
+       addresses extracted from *Child Auth* for *NS Name*.
+    6. Else, if the *Parent Glue* set is not equal to the *Child Auth* set then
+       add name/IP pairs from the *Child Auth* set not present in the
+       *Parent Glue* set to the *Extra Address Child* set.
+
+14. If the *Extra Address Child* set is non-empty, then output
+    *[CS05_EXTRA_ADDR_CHILD]* with the list of name/IP pairs from the set.
+
+15. If the the *Delegation OOD NS* set is non-empty then for each name server
```
marc-vanderwal:
```suggestion
15. If the *Delegation OOD NS* set is non-empty then for each name server
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,344 @@
...
+    2. Go to next name server name if there are no name/IP pairs for the name.
+    3. Do [Address Records Lookup] of the name server name. Ignore any failing
+       lookups (such as response with SERVFAIL or no response) or lookups giving
+       response with NODATA or NXDOMAIN.
+    4. If any IP address is returned from the lookup, then do:
+       1. Compare the IP address(es) with the address(es) in the extracted
+          name/IP pairs.
+       2. If the comparison did not give an exact match, then output
+          *[CS05_OOD_ADDR_MISMATCH]* with the name server name, the address(es)
+          from the extracted name/IP pairs and the the addresses from the
```
marc-vanderwal:
```suggestion
          from the extracted name/IP pairs and the addresses from the
```

marc-vanderwal (dismissed):

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,355 @@
...
-For an [undelegated test] it is possible to intentionally insert data
-for [out-of-bailiwick][terminology] name servers that do not match what is found in
-public DNS. This Test Case will then report this as an ERROR which
-may not match the users expectation.
+[RFC 9471][RFC 9471#section3.1], section 3.1, requires that a parent name server
+include all available [in-domain] glue records in a [referral] of a child zone or
+set the [TC bit][RFC 1035#section4.1.1]. Some name servers fail to set the TC bit
+correctly in this case, and to make sure all glue records are provided for this
+test case, the parent name servers are queried over TCP with a fallback to UDP,
+as specified in the test procedures. Detecting a failing to set the TC bit is not
```
marc-vanderwal:
```suggestion
as specified in the test procedures. Detecting a failure to set the TC bit is not
```

marc-vanderwal (commented):
Looks good to me otherwise.

marc-vanderwal (approved):

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,355 @@
...
+* [Intercase dependencies](#intercase-dependencies)
+* [Terminology](#terminology)
+
+
 ## Objective
 
-For name servers that have IP addresses listed as glue, the IP addresses must
-match the authoritative A and AAAA records for that host. This is an IANA 
-[name server requirement].
+The delegation of the *Child Zone* may contain so called
```
tgreenx:
```suggestion
The delegation of the *Child Zone* may contain so-called
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,355 @@
...
+one [glue record] in the delegation and in the [referral]. For each such glue
+record the equivalent [address record] must exist as an authoritative record in
+the child zone, or below, and both records must have the same [IP address].
+
+The other alternative is that the name server name is [Out-Of-Domain], and in
+that case there can be a [glue record], and it must also match an address
+record with the same IP address in the authoritative zone. In this case that
+authoritative zone is not the *Child Zone* since the name is [Out-Of-Domain].
+
+It is an IANA [name server requirement] that [glue records][glue record] matches
```
tgreenx:
```suggestion
It is an IANA [name server requirement] that [glue records][glue record] match
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,355 @@
...
+The delegation of the *Child Zone* may contain so called
+[glue records][glue record]. On the wire, any [glue records][glue record] are
+found in the [referral] sent from the name servers of the parent zone.
+
+If the name server name in the NS record is [in-domain] there must be at least
+one [glue record] in the delegation and in the [referral]. For each such glue
+record the equivalent [address record] must exist as an authoritative record in
+the child zone, or below, and both records must have the same [IP address].
+
+The other alternative is that the name server name is [Out-Of-Domain], and in
```
tgreenx:
```suggestion
Alternatively the name server name is [Out-Of-Domain], and in
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,355 @@
...
+| CS05_NO_MISMATCH_GLUE_ZONE     | INFO     |                                           | There is no mismatch between glue in delegation and authoritative data in the child zone.                                                                  |
+| CS05_NO_NS_ADDR_CHILD          | CRITICAL |                                           | Child zone cannot be tested since there are no name server IP addresses for that zone.                                                                     |
+| CS05_OOD_ADDR_MISMATCH         | WARNING  | nsname, ns_ip_list_ref, ns_ip_list_lookup | For name server {nsname} the glue record in the delegation "{ns_ip_list_glue}" is different from the address record in the child zone "{ns_ip_list_zone}". |
+
+The value in the Level column is the default severity level of the message. The
+severity level can be changed in the [Zonemaster-Engine profile]. Also see the
+[Severity Level Definitions] document.
+
+The argument names in the Arguments column lists the arguments used in the
```
tgreenx:
```suggestion
The argument names in the Arguments column list the arguments used in the
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,355 @@
...
+       server IP and fetch the response.
+    3. If a response was received and it contains a [Referral] to *Child Zone*
+       then do:
+       1. Extract the name server names from delegation NS records in authority
+          section.
+       2. Downcase any uppercase letters in the names.
+       3. Create a sorted list of unique name server names (sorted in ascending order
+          on the name).
+       4. For each [glue record] in the additional section (if any) do:
+          1. Form an name/IP pair from the owner named and the IP address in
```
tgreenx:
```suggestion
          1. Form a name/IP pair from the owner name and the IP address in
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,355 @@
...
+5.  If the *Delegation* set is non-empty, then for each parent name server IP in
+    the set do:
+    1. For each element (name or name/IP pair) in the set do:
+       1. If the name in the element is [In-Domain] then do:
+          1. If element is a name and not a name/IP pair then add the parent name
+             server IP address and the name to the *Missing Glue* set.
+          2. Add the element (name or name/IP pair) to the *Delegation ID NS*
+             set.
+             1. Do not create duplicates in the set.
+             2. Elements just consiting of a name is not added if there already
```
tgreenx:
```suggestion
             2. Elements just consisting of a name is not added if there already
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,355 @@
...
+                is a name/IP pair with the same name.
+             3. A name/IP pair will overwrite an element consisting just of
+                the same name.
+       2. If name in the element (name or name/IP pair) is [Out-Of-Domain] then
+          add the element (name or name/IP pair) to the *Delegation OOD NS* set.
+          1. Do not create duplicates in the set.
+          2. Elements just consiting of a name is not added if there already is a
```
tgreenx:
```suggestion
          2. Elements just consisting of a name is not added if there already is a
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,355 @@
...
+    5. For each [In-Domain] name server name do:
+       1.  Create a [DNS Query] with query type A and query name the NS name
+           server name ("A Query").
+       2.  [Send] *A Query* to the name server IP and fetch the response (if
+           any).
+       3.  If the response (if any) contains a [Referral] covering the NS name
+           server name then repeat *A Query* (recursively, if needed) to the
+           name servers in the referral until an A RRset is returned or the
+           querying is stopped by e.g. NXDOMAIN or no response.
+           1. If an A RRset is returned, then use it in the next step as if was a
```
tgreenx:
```suggestion
           1. If an A RRset is returned, then use it in the next step as if it were a
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,355 @@
...
+       1. If element is a name and not a name/IP pair then output
+          *[CS05_MISSING_GLUE_FOR_ROOT_NS]* with the name server name.
+       2. Else, add the element (name/IP pair) to the *Delegation ID NS* set.
+          1. Do not create duplicates in the set.
+
+9. If the *Missing Glue* set is non-empty, then for each name server name output
+    *[CS05_MISSING_GLUE_FOR_NS]* with the name server name and the list of parent
+    IP addresses.
+
+10. If *Child NS IP* is empty then output *[CS05_NO_NS_ADDR_CHILD]* and exit
```
tgreenx:
```suggestion
10. If *Child NS IPs* is empty then output *[CS05_NO_NS_ADDR_CHILD]* and exit
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,355 @@
...
+with the severity level *[ERROR]* or *[CRITICAL]*.
 
 The outcome of this Test Case is "warning" if there is at least one message
-with the severity level *WARNING*, but no message with severity level
+with the severity level *[WARNING]*, but no message with severity level
 *ERROR* or *CRITICAL*.
 
-The outcome of this Test case is "pass" in all other cases.
+In other cases, no message or only messages with severity level
+*[INFO]* or *[NOTICE]*, the outcome of this Test Case is "pass".
```
tgreenx:
```suggestion
In other cases, where there is no message or only messages with severity
level *[INFO]* or *[NOTICE]*, the outcome of this Test Case is "pass".
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -165,41 +358,118 @@ None
...
+  used.
+
+* "IP Address" -- In this document the term stands for either an [IPv4] address or
+  an [IPv6] address in any address range.
+
+* "Out-Of-Domain" - The term refers to a name server name that is not
+  "In-Domain". It is either "sibling domain" or "unrelated" as defined in
+  [RFC 9499][RFC 9499#section7], section 7, in the subsection on "Glue Records",
+  for name server names in the referral of a zone. The name server name is
+  neither on or below the zone cut of the zone for which it is name server.
```
tgreenx:
```suggestion
  neither on nor below the zone cut of the zone for which it is name server.
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -165,41 +358,118 @@ None
...
+  The name server name belongs to another zone. It is above or at aside the
```
tgreenx: "Outside" rather? Or "alongside"?
```suggestion
  The name server name belongs to another zone. It is above or outside the
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,355 @@
...
+       1. Extract the name server names from delegation NS records in authority
+          section.
+       2. Downcase any uppercase letters in the names.
+       3. Create a sorted list of unique name server names (sorted in ascending order
+          on the name).
+       4. For each [glue record] in the additional section (if any) do:
+          1. Form an name/IP pair from the owner named and the IP address in
+             [RDATA].
+          2. Only unique name/IP pairs are created.
+          3. Add the pair to sorted list of name server names.
```
tgreenx:
```suggestion
          3. Add the pair to the sorted list of name server names.
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,355 @@
...
+| CS05_MISSING_GLUE_FOR_ROOT_NS  | WARNING  | nsname                                    | IP address (glue record) is expected but missing for {nsname} in the undelegated data or hint data for root.                                               |
+| CS05_NO_MISMATCH_GLUE_ZONE     | INFO     |                                           | There is no mismatch between glue in delegation and authoritative data in the child zone.                                                                  |
+| CS05_NO_NS_ADDR_CHILD          | CRITICAL |                                           | Child zone cannot be tested since there are no name server IP addresses for that zone.                                                                     |
+| CS05_OOD_ADDR_MISMATCH         | WARNING  | nsname, ns_ip_list_ref, ns_ip_list_lookup | For name server {nsname} the glue record in the delegation "{ns_ip_list_glue}" is different from the address record in the child zone "{ns_ip_list_zone}". |
```
tgreenx, line 84:
```suggestion
| CS05_OOD_ADDR_MISMATCH         | WARNING  | nsname, ns_ip_list_glue, ns_ip_list_zone | For name server {nsname} the glue record in the delegation "{ns_ip_list_glue}" is different from the address record in the child zone "{ns_ip_list_zone}". |
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,355 @@
...
+| CS05_CHILD_ZONE_LAME           | CRITICAL | ns_list                                   | There is no working name server for the child zone. Tested name servers are "{ns_list}".                                                                   |
+| CS05_DELEGATION                | INFO     | ns_deleg_list, ns_list                    | Delegation of the child zone as provided by the parent name servers listed: "{ns_deleg_list}". Parent name servers: "{ns_list}".                           |
+| CS05_EXTRA_ADDR_CHILD          | NOTICE   | ns_list                                   | There is one or more extra address records found in the child zone that are not present as glue in the delegation: "{ns_list}".                            |
+| CS05_ID_ADDR_MISMATCH          | ERROR    | nsname, ns_ip_list_glue, ns_ip_list_zone  | For name server {nsname} the glue record in the delegation "{ns_ip_list_glue}" is different from the address record in the child zone "{ns_ip_list_zone}". |
```
tgreenx, line 76: Arguments `ns_ip_list_glue` and `ns_ip_list_zone` are not defined in `ArgumentForTestCaseMessages.md`

docs/public/specifications/tests/Consistency-TP/consistency05.md
```diff
@@ -1,162 +1,355 @@
...
+| CS05_ID_ADDR_MISSING           | NOTICE   | nsname                                    | Address record for {nsname}, used as glue record in delegation, is missing in the child zone.                                                              |
+| CS05_INCONSISTENT_DELEGATION   | WARNING  |                                           | The delegation is inconsistent between the parent nameservers.                                                                                             |
```
tgreenx:
```suggestion
| CS05_INCONSISTENT_DELEGATION   | WARNING  |                                           | The delegation is inconsistent between the parent name servers.                                                                                            |
```

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: "glue" could work (instead of "ref" = "referal") but "zone" could be misleading. "Zone" usually refer to child zone, and this IP never comes from a lookup in the child zone. I felt "lookup" to be the best choice since the process of getting that IP is referred to as "lookup".

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: Yes, indirectly. Those are two separate instances of the same argument name `ns_ip_list` and it is specified in the document that this is the way to handle it (https://github.com/matsduf/zonemaster/blob/master/docs/public/specifications/tests/ArgumentsForTestCaseMessages.md#example-of-multiple-instances)

docs/public/specifications/tests/Consistency-TP/consistency05.md
matsduf: Now I see the mismatch between arguments listed and the msgid. And also the text in the msgid. I must have copied and then forgot to update. I will fix.

tgreenx (approved):

